PR #18, opened 2026-06-21
Phase 6: Saved views

naeemba:
## Summary

Implements the **Saved views** Phase 6 item: pin a filtered page URL under a name and recall it from a "Saved views" panel on the Dashboard. Third Phase 6 feature after CSV exports, base-currency selector, and the daily price fetcher.

URL-bookmark model — each view stores the full canonicalized path + query as one string. Six allowlisted filterable routes; server-side `targetPath` validation rejects anything outside the allowlist.

## What's included

- **Data layer** — `db/schema/savedView.ts` (`(userId, name)` unique index), `lib/savedViews/` Repository + Service + Zod schema, mirroring the existing `template` convention.
- **Server actions** — one file each under `features/savedViews/actions/`: `saveSavedView`, `renameSavedView`, `deleteSavedView`.
- **Creation surface** — `SaveViewButton` mounted on the six filter toolbars: transactions, balance (`DateFilter`), payees (`DateFilter`), monthly register (new `RegisterHeader`), account register (new `AccountHeader`). Save-or-overwrite-by-name flow.
- **Management surface** — `SavedViewsCard` on the Dashboard (above Recent transactions) with click-to-navigate, inline rename, and delete via row-actions dropdown. Coarse route label via `routeLabel` helper.

## Non-goals (v1)

Per-filter human summaries, sidebar group / command-palette action, pin ordering, stale-target detection, saving static (non-filterable) pages.

## Tests

497 passing (87 files). Saved-views coverage at every layer: schema, repository, service, all three actions, `routeLabel`, and an end-to-end save → list → rename → delete integration test. `type-check` and `lint` both clean.

Spec: `docs/superpowers/specs/2026-06-11-saved-views-design.md`
Plan: `docs/superpowers/plans/2026-06-11-saved-views.md`


lib/savedViews/schema.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export const savedViewNameSchema = z
+  .string()
+  .trim()
+  .min(1, 'Name is required')
+  .max(NAME_MAX, 'Name is too long')
+  .refine((v) => !/[\x00-\x1F]/.test(v), 'Name contains control characters');
+
+const ISO_DATE = /^\d{4}-\d{2}-\d{2}$/;
+const ACCOUNT_SEGMENT = /^[A-Za-z0-9:_\- ]+$/;
```
naeemba: **Account names with characters outside `[A-Za-z0-9:_\- ]` cannot be saved as views, even though those accounts are fully routable.**

`AccountHeader`/`RegisterHeader` build `targetPath` with `encodeURIComponent(account)`, and the pages gate on `isValidAccount` (`utils/validateAccount.ts`), which accepts *any* account except those starting with `-`, longer than 256, or containing NUL/newline. So `Expenses:Food & Dining`, `Assets:Café`, `Liabilities:Loan (2024)` all render fine and are linked across the app (Dashboard, Reconcile, Portfolio, AccountButtons). But `ACCOUNT_SEGMENT = /^[A-Za-z0-9:_\- ]+$/` here rejects every one of them after `decodeURIComponent`, so `canonicalizeTargetPath` throws and the Save view button silently fails with `invalid` for legitimate accounts.

Worse, an account containing `/` (e.g. `Assets:A/B`) encodes to `%2F`, which `hasTraversalSegment` flags as a traversal segment — also a false rejection.

The allowlist should accept the same account character set that `isValidAccount` accepts (or call `isValidAccount(decodeURIComponent(seg))` directly) rather than re-spelling a narrower regex. Reusing `isValidAccount` would also de-duplicate the validation rule.

lib/savedViews/schema.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const matchesAllowlist = (pathname: string): boolean => {
+  if (pathname === '/transactions') return true;
+  if (pathname === '/balance') return true;
+  const balanceRange = pathname.match(
+    /^\/balance\/(\d{4}-\d{2}-\d{2})\/(\d{4}-\d{2}-\d{2})$/
+  );
+  if (balanceRange)
```
naeemba: The date allowlist accepts syntactically-shaped but semantically-invalid dates: `/balance/2026-13-45/2026-99-99` passes `matchesAllowlist` (the regex only checks `\d{4}-\d{2}-\d{2}` shape, and the redundant `ISO_DATE.test(...)` re-check on lines 22-23 / 27-28 tests the exact substring the capture already matched, so it can never fail). A saved view can therefore store a target that ledger will reject at navigation time. Low impact since the dynamic route itself is shape-only too, but worth either using `parseISODate`/`Date` validation (consistent with how `app/balance/[from]/[to]/page.tsx` parses) or dropping the dead `ISO_DATE.test` re-checks. Flagging the dead re-check as the concrete cleanup.

app/balance/[from]/[to]/page.tsx
```diff
@@ -18,6 +21,11 @@ const PeriodBalance = async ({
   const { from: fromParam, to: toParam } = await params;
   const from = parseISODate(fromParam);
   const to = parseISODate(toParam);
+  const user = await requireUser();
```
Comment (naeemba) [inline app/balance/[from]/[to]/page.tsx:24]: The exact snippet `const existingViewNames = (await savedViewService.list(user.id)).map((v) => v.name);` is copy-pasted into five render paths: this file, `app/accounts/[account]/page.tsx`, `app/registers/monthly/[account]/page.tsx`, `features/payees/Payees.tsx`, and `features/transactions/Transactions.tsx`. Each also independently calls `requireUser()` just to get `user.id`. Consider a small helper on the service (e.g. `savedViewService.listNames(userId)`) so the projection lives in the logic layer instead of being re-spelled at every transport site — it also keeps the `name`-only contract in one place.

features/savedViews/routeLabel.ts
```diff
@@ -0,0 +1,20 @@
+export const routeLabel = (targetPath: string): string => {
+  const pathname = targetPath.split('?')[0] ?? targetPath;
+  if (pathname === '/transactions') return 'Transactions';
+  if (pathname === '/balance' || /^\/balance\/[^/]+\/[^/]+$/.test(pathname))
```
naeemba: nit: `routeLabel` re-spells the six-route structure that `matchesAllowlist` in `lib/savedViews/schema.ts` already encodes (transactions / balance / balance-range / payees-range / registers-monthly / accounts), with slightly different regexes (`[^/]+` here vs `\d{4}-\d{2}-\d{2}` there). If a route is ever added or its shape changes, both must be edited in lockstep or the label silently falls through to the raw pathname. Consider deriving both from a single shared route table/constant so the allowlist and the label stay in sync.

lib/savedViews/schema.ts
naeemba: Fixed in 871d481. The allowlist now validates account segments via `isValidAccount(decodeURIComponent(seg))` (through the shared route table in `lib/savedViews/routes.ts`) instead of the narrower regex, so accounts like `Expenses:Food & Dining`, `Assets:Café`, and `Liabilities:Loan (2024)` save correctly. The traversal check now decodes each segment before looking for `..`, so an encoded slash inside a single account segment (`Assets:A/B` -> `%2F`) is accepted while `..%2F` is still rejected.

lib/savedViews/schema.ts
naeemba: Fixed in 871d481. The dead `ISO_DATE.test(...)` re-checks are gone — the date routes now live in the shared route table where the `\d{4}-\d{2}-\d{2}` capture is the only shape gate. (Left full `parseISODate`/`Date` semantic validation out since the dynamic route is shape-only too, as noted; the concrete cleanup you flagged is addressed.)

app/balance/[from]/[to]/page.tsx
Comment (naeemba) [reply app/balance/[from]/[to]/page.tsx:24]: Fixed in b68c41b. Added `SavedViewService.listNames(userId)` and switched all five name-only render paths to it, so the `.map(v => v.name)` projection lives in the logic layer. Left the per-page `requireUser()` calls in place since those pages use `user.id` for other data loads too.

features/savedViews/routeLabel.ts
naeemba: Fixed in 871d481. `routeLabel` now derives from the same `SAVED_VIEW_ROUTES` table in `lib/savedViews/routes.ts` that `matchesAllowlist` uses, so the allowlist and the label can no longer drift — adding or reshaping a route is a single edit.

features/savedViews/actions/deleteSavedView.ts
```diff
@@ -0,0 +1,11 @@
+'use server';
+
+import { requireUser } from '@/lib/auth/require-user';
+import { savedViewService } from '@/lib/savedViews';
+import { revalidatePath } from 'next/cache';
+
+export const deleteSavedViewAction = async (id: string): Promise<void> => {
```
naeemba: The canonical sibling `deleteTemplateAction` (`features/templates/actions/deleteTemplate.ts`) returns a `DeleteTemplateResult` (`{ ok: true } | { ok: false; message }`) and wraps the repo call in try/catch so the caller can surface failures. This action diverges: it returns `Promise<void>` with no error handling, and the only caller (`SavedViewRowActions.onDelete`) unconditionally fires `toast.success('View deleted')` after awaiting it.

Result: if the delete throws (DB error, or `requireUser` rejecting), the user still sees a "View deleted" success toast while the row remains on the page — a false-success report and an unhandled rejection inside `startTransition`. Match the template pattern: return an `{ ok }` result and have `onDelete` branch on it (error toast on failure) like the rename path already does.

naeemba: Fixed in 7670964. `deleteSavedViewAction` now matches the `deleteTemplateAction` pattern: it returns `DeleteSavedViewResult` (`{ ok: true } | { ok: false; message }`) and wraps the repo call in try/catch. `SavedViewRowActions.onDelete` branches on the result, showing `toast.error(...)` on failure instead of unconditionally firing `toast.success("View deleted")`. Added a test asserting the error path returns `{ ok: false }` and skips `revalidatePath`.

PR #19, opened 2026-06-22
Migrate SQLite → Postgres + adopt @naeemba/next-starter auth

naeemba:
## Summary

Migrates the data layer from **SQLite/better-sqlite3 → PostgreSQL** (Drizzle + `postgres` driver) and adopts the in-house **`@naeemba/next-starter`** package for auth (magic-link + passkey + Google, sign-in locked to a single admin email).

This unblocks the Coolify build, which was failing because the SQLite connection was opened at module-import time and the data directory doesn't exist in the build container. The DB connection is now lazy, and there is no SQLite file at all.

Fresh start — **no data migration** (financial data lives in ledger journal files, untouched).

## What changed

- **DB:** lazy `postgres.js` Drizzle connection; app tables rewritten in `pg-core` (`userSetting`, `template`, `savedView`, `commodity_price`, `price_fetch_run`). `journalMain` folded into `userSetting`; `baseCurrency` nullable.
- **Migrations (two tracks):** auth tables are package-owned (`next-starter migrate`); app tables via `drizzle-kit migrate`, scoped with `tablesFilter`. FKs reference `user` from `@naeemba/next-starter/schema`. Deploy ordering wired into `prebuild`/`prestart`.
- **Repositories:** all five converted from synchronous better-sqlite3 to async Postgres. Fixed `isUniqueConflict` in the template/saved-view services to detect Postgres errors (`23505` / `duplicate key value`) — previously only matched SQLite text, so name-conflicts would have thrown.
- **Auth:** starter shims (`lib/auth.ts`, `lib/auth-client.ts`, `lib/auth-server.ts`, auth route, `/sign-in` + `/sign-in/error` + `/settings/passkeys` pages, `proxy.ts`). `email+password` and `/login` + `/signup` removed. The 13 `requireUser`/`getOptionalUser` callers preserved via a thin adapter.
- **Tests:** entire suite migrated to an in-process **PGlite** harness with per-test `DATA_DIR` isolation; stray worktrees excluded from the vitest glob.

## Verification

- `pnpm test` → **64 files / 345 tests passing**, pristine
- `pnpm type-check` → 0 errors
- `pnpm lint` → clean
- `pnpm build` → **succeeds** against a real Postgres (the original `/_not-found` DB-open failure is gone)
- Both migration tracks apply cleanly to a fresh Postgres 16 (10 tables, FK ordering correct)

Built spec-first (design + plan in `docs/superpowers/`), then implemented task-by-task with a fresh-context reviewer gating each task and a final whole-branch review (verdict: ready to merge, no Critical/Important findings).

## Before deploying on Coolify

- Add a **Postgres service**; set `DATABASE_URL` (`postgres://…`), `BETTER_AUTH_SECRET`, `BETTER_AUTH_URL`, `EMAIL_FROM`. Set `RESEND_API_KEY` for magic-link delivery in prod (without it, links only log to stdout). Optional: `GOOGLE_CLIENT_ID/SECRET` + `NEXT_PUBLIC_ENABLE_GOOGLE=1`.
- **Manual smoke not yet run** (needs live PG + browser): magic-link sign-in, passkey register/login, Google, and the singleAdmin lock.

## Deferred minor follow-ups

- `.env.example` omits `NEXT_PUBLIC_APP_ENV`/`PORT` (cosmetic)
- generated migration SQL missing trailing newline
- `updatedAt` uses `sql\`now()\`` in saved-views vs `new Date()` elsewhere (harmless inconsistency)


naeemba (commented):
## Review: SQLite → Postgres + next-starter auth migration

Solid, well-sequenced migration — repositories are cleanly async, the lazy `db` proxy correctly keeps `next build` DB-free, the price-run tombstone-up-front pattern is preserved, and async call sites were all updated (verified the 13 `requireUser` callers and every now-async repo/service consumer await correctly).

Most findings below are **correctness/wiring** issues from the route rename and removed config, plus some **reuse/structure** cleanup. Inline comments mark each. The most important one is **not in the diff**:

### 🔴 `components/AppShell/AppShell.tsx:13` — stale `AUTH_PATHS` orphans the sign-in page (out of diff)
This PR renames the auth route to `/sign-in` (+ `/sign-in/error`) and deletes `/login` + `/signup`, but `AppShell` still gates its chrome-free centered layout on `AUTH_PATHS = new Set(['/login', '/signup'])`. `app/layout.tsx` wraps **all** children in `<AppShell>`, so an unauthenticated user redirected to `/sign-in` now renders inside the full app chrome — `SidebarProvider` + `AppSidebar` + `AppHeader` (which calls `useSession`/`signOut` and assumes an authenticated session) — instead of the dedicated centered card. `SignInPage`'s own `classNames.main` can't undo the surrounding sidebar/header.
**Fix:** `const AUTH_PATHS = new Set(['/sign-in', '/sign-in/error'])`. Worth a test asserting the auth layout branch triggers for the sign-in path so the next route rename can't silently regress it.

### Verified and refuted (so you don't re-litigate)
- **`isUniqueConflict` is NOT broken under the prod `postgres` driver.** A natural worry is that it only inspects `e.cause` while porsager's `PostgresError` carries `code`/`message` on the error itself. But drizzle-orm 0.45.2 wraps *every* driver error in `DrizzleQueryError` via `queryWithCache` (shared `pg-core` base), setting `.cause` to the original `PostgresError`. So `cause.message`/`cause.code === '23505'` matches in production exactly as in the PGlite tests. It's still worth deduping (see inline), just not a bug.
- **No `journalMain`/`userSetting` backfill is fine** — this is a declared fresh start with no data migration; readers fall back (`DEFAULT_MAIN`, `env.DEFAULT_CURRENCY`).
- **`setMainFile` FK insert** is prod-safe (`requireUser` guarantees a real `user` row); only a concern for test fixtures that skip `insertUser`.

### Missing tests worth adding
- AppShell auth-layout branch for `/sign-in` (would have caught the bug above).
- A unique-conflict test that exercises the real `postgres` driver path (or at least asserts the shared helper against a `DrizzleQueryError`-shaped error), since PGlite is the only conflict coverage today.

lib/db/index.ts
```diff
@@ -1,9 +1,19 @@
...
 
-export const db = createDbConnection(dbPath);
+// Connect lazily on first property access, never at import time, so `next build`
+// can evaluate route modules without a database connection being established.
+const getDb = (): DbInstance => {
+  if (!instance) instance = createDbConnection(env.DATABASE_URL);
+  return instance;
+};
+
+export const db = new Proxy({} as DbInstance, {
```
naeemba: **Reuse — re-implements `@naeemba/next-starter/db`.** The adopted package already exports `db` (a lazy proxy), `createDb`, and `createDbOptionsFromEnv` from `@naeemba/next-starter/db`. This file plus `lib/db/connection.ts` hand-roll the same lazy-proxy + `postgres()`+`drizzle()` wiring, but read only `env.DATABASE_URL` — so the package's env-driven pool tuning (`createDbOptionsFromEnv`: prepare/max/idle-timeout) is ignored, and the app must carry its own `postgres`/`drizzle-orm` direct deps kept version-locked with the package. Prefer re-exporting the package's `db`/`createDb` here (the ~10 `@/lib/db` importers keep working) so the two bootstraps can't drift. (Same root issue as `lib/db/connection.ts:7`.)

lib/templates/service.ts
```diff
@@ -10,8 +10,22 @@ export type RenameResult =
   | { ok: true; template: Template }
   | { ok: false; reason: 'name-conflict' | 'not-found' };
 
-const isUniqueConflict = (e: unknown): boolean =>
-  e instanceof Error && /UNIQUE constraint failed/i.test(e.message);
+const isUniqueConflict = (e: unknown): boolean => {
```
naeemba: **Reuse + dead code — duplicated `isUniqueConflict`.** This 16-line helper is byte-for-byte identical to `lib/savedViews/service.ts:13`. Extract one shared helper (e.g. `lib/db/isUniqueConflict.ts`) and import in both — otherwise the next DB-error-shape change has to land in two places and one will be missed. Also drop the `/UNIQUE constraint failed/i` (SQLite) branch on line 16: SQLite is gone in this PR, so it's dead.

_(Note: the helper itself is correct for the prod driver — drizzle wraps the `PostgresError` in `DrizzleQueryError.cause`, so the `cause.code === '23505'` branch fires. This is cleanup, not a bug.)_

lib/env/index.ts
```diff
@@ -12,16 +12,22 @@ const envSchema = clientEnvSchema.extend({
     ),
   BETTER_AUTH_URL: z.string().url().default('http://localhost:3000'),
 
-  // Storage
-  DATA_DIR: z.string().default('./data'),
-  DATABASE_URL: z.string().optional(),
+  // Storage — Postgres connection string (postgres://user:pass@host:5432/db)
+  DATABASE_URL: z.string().url(),
```
naeemba, line 16: **Removed validation — `DATA_DIR` no longer validated or documented.** `lib/journal/layout.ts:11` still reads `process.env.DATA_DIR ?? './data'` (the root for all persistent journals), but it's gone from this schema and from `.env.example`. A misconfigured/empty `DATA_DIR` in production now silently falls back to `./data` instead of failing fast at startup as it did before. Add `DATA_DIR` to the schema (with a sensible default) so a bad deploy surfaces here rather than writing journals to the wrong place.

package.json
```diff
@@ -4,14 +4,17 @@
   "private": true,
   "scripts": {
     "dev": "dotenv -e .env -- next dev",
+    "prebuild": "next-starter migrate",
```
naeemba: **Migration wiring is asymmetric / fragile.** `prebuild` runs only the auth track (`next-starter migrate`), `prestart` runs both tracks, and `dev` runs neither. Consequences: a fresh `pnpm dev` clone hits `relation "..." does not exist` until someone runs `db:migrate` by hand; and any deploy that builds but doesn't start via `pnpm start` never applies the app-table track. Consider a single composite script — `"migrate": "next-starter migrate && drizzle-kit migrate"` — reused by `predev`/`prebuild`/`prestart`, so ordering (auth `user` table before app FKs) and completeness are defined once.

lib/test-utils/db.ts
```diff
@@ -1,52 +1,103 @@
...
+  insertUser: (id: string, name?: string, email?: string) => Promise<void>;
   tmpDir: string;
-  dbPath: string;
-  sqlite: Database.Database;
 };
 
+// Minimal stand-in for the package-owned auth `user` table plus the app tables.
+// Postgres DDL mirrors db/schema and @naeemba/next-starter/schema closely enough
+// for repository tests (FKs, uniqueness, defaults).
+const SCHEMA_SQL = `
```
naeemba: **Test quality — `SCHEMA_SQL` is a second source of truth.** This hand-written DDL re-spells the auth `user` table plus every app table, duplicating `db/migrations/0000_*.sql` and the package-owned auth schema. When a column/constraint changes in the real migration (or the starter's `user` shape changes), tests keep passing against stale DDL, so a schema/repo mismatch ships green — i.e. the suite never exercises the actual migrations. Prefer applying the real generated migration SQL (and the starter's) against PGlite in the harness.

lib/settings/getBaseCurrency.ts
```diff
@@ -19,7 +19,7 @@ export const getBaseCurrency = cache(async (): Promise<string> => {
   const user = await getOptionalUser();
   if (user) {
     const row = await userSettingRepository.get(user.id);
-    if (row) return row.baseCurrency;
+    if (row?.baseCurrency) return row.baseCurrency;
```
naeemba: **Behavior change — DB failure now errors the whole route.** Previously this resolved from local better-sqlite3 and effectively couldn't fail; it now performs a Postgres round-trip inside the React-cached function. If that query rejects, `getBaseCurrency()` rejects, and the ~20 pages that call it (mostly outside try/catch) error the entire route instead of degrading to `env.DEFAULT_CURRENCY`. Consider wrapping the `userSettingRepository.get` call so a DB hiccup falls back to the default rather than 500-ing every page.

lib/prices/repository.ts
```diff
@@ -22,7 +22,7 @@ export class CommodityPriceRepository {
   async insert(rows: CommodityPriceInput[]): Promise<void> {
     if (rows.length === 0) return;
     for (const r of rows) {
```
naeemba: **Efficiency (minor) — per-row upsert round-trips.** Under better-sqlite3 this loop was N local synchronous statements; under postgres.js each `await` is a separate network round-trip with no enclosing transaction, so a partial failure mid-loop leaves a partially-committed set. Drizzle accepts an array in `.insert(commodityPrice).values(rows).onConflictDoUpdate(...)` for a single batched upsert. Low urgency given the single-admin lock keeps row counts small, but it's a cheap win and restores all-or-nothing semantics.

lib/savedViews/repository.ts
```diff
@@ -56,26 +55,26 @@ export class SavedViewRepository {
     id: string,
     patch: SavedViewPatch
   ): Promise<SavedView | null> {
-    const updates: SavedViewPatch & { updatedAt: Date } = {
-      updatedAt: new Date(),
+    const updates: SavedViewPatch & {
+      updatedAt: Date | ReturnType<typeof sql>;
```
naeemba: **Cleanup (minor) — contrived patch type.** The `Date | ReturnType<typeof sql>` union just to let `set()` accept `sql\`now()\`` leaks Drizzle internals into the patch shape. Cleaner: add `$onUpdate(() => new Date())` (or `.$onUpdateFn`) to `updatedAt` in the schema so the repo never spells the bump, or keep `updatedAt: sql\`now()\`` with a narrow cast. Also note `setMainFile`/`upsertBaseCurrency` use `new Date()` for the same bump — worth picking one convention.

lib/db/index.ts
naeemba: Fixed in 733d33a. `lib/db/index.ts` now re-exports the package's lazy `db` proxy (and `createDb` via `connection.ts`) from `@naeemba/next-starter/db`, dropping the hand-rolled `postgres()`+`drizzle()` wiring. The package proxy reads `process.env.DATABASE_URL` and applies `createDbOptionsFromEnv()` pool tuning, so the two bootstraps can't drift. `DbInstance` is now `ReturnType<typeof createDb>`; verified no `db.query.*` relational usage exists, so the schema generic is irrelevant to the `.select()/.insert()` call sites and all ~10 importers + the PGlite test cast still type-check.

lib/templates/service.ts
naeemba: Fixed in 733d33a. Extracted to `lib/db/isUniqueConflict.ts`, imported by both `templates/service.ts` and `savedViews/service.ts`, and dropped the dead SQLite branch. Thanks for confirming the prod-driver path — kept the `cause.code === '23505'` / `duplicate key value` checks intact.

lib/env/index.ts
naeemba: Fixed in 03ed98c. Added `DATA_DIR: z.string().min(1).default('./data')` to the env schema (so an empty value now fails fast at startup) and documented it in `.env.example`. Left `layout.ts` reading `process.env.DATA_DIR` directly on purpose — the test harness overrides it per-case after `@/lib/env` has already parsed once.

package.json
naeemba: Fixed in bd9f0a6. Added a single `"migrate": "next-starter migrate && drizzle-kit migrate"` reused by `predev` (with dotenv), `prebuild`, and `prestart`, so ordering (auth `user` before app FKs) and completeness live in one place. A fresh `pnpm dev` now migrates both tracks, and a build-without-`pnpm start` deploy applies the app tables too.

lib/test-utils/db.ts
naeemba: Fixed in f7203b3. Dropped `SCHEMA_SQL` entirely; the harness now applies the **real** generated migrations to PGlite — the package-owned auth track first via `resolveMigrationsFolder()` (creates `user`), then the app's own `db/migrations`. Your hunch was exactly right: the hand-written `user` DDL had already drifted (camelCase `emailVerified`/`createdAt` vs the package's snake_case `email_verified`/`created_at`). All 347 tests pass against the shipped schema.

lib/settings/getBaseCurrency.ts
naeemba: Fixed in f4ddedd. Wrapped the `userSettingRepository.get` call in try/catch — a DB hiccup now logs and falls back to `env.DEFAULT_CURRENCY` instead of rejecting and 500-ing every page that reads the base currency.

lib/prices/repository.ts
naeemba: Fixed in 45ba646. Replaced the per-row `await` loop with a single `insert(commodityPrice).values(rows).onConflictDoUpdate(...)` — one statement, so the upsert is atomic (all-or-nothing) instead of partially committing on a mid-loop failure.

lib/savedViews/repository.ts
naeemba: Fixed in acd65f0. Added `\$onUpdate(() => sql`now()`)` to `updatedAt` on `savedView`/`template`, so `update()` no longer spells the bump — the contrived `Date | ReturnType<typeof sql>` union and template's manual `new Date()` are gone. On one convention: I used `sql`now()`` rather than `new Date()` so `updatedAt` shares `createdAt`'s clock source — a host `new Date()` diverges by the server's UTC offset and can land `updatedAt` *before* `createdAt` (this actually broke the savedViews tests until I switched). Since `\$onUpdate` doesn't fire on `onConflictDoUpdate`, the two upserts (`upsertBaseCurrency`, `setMainFile`) now spell `sql`now()`` too, so every `updatedAt` uses the DB clock.

naeemba:
## Re: 🔴 `AppShell.tsx` stale `AUTH_PATHS` orphans the sign-in page (from review body)

Fixed in 0e9a8b7. Confirmed the bug: `AppShell` still gated the chrome-free centered layout on the deleted `AUTH_PATHS = {/login, /signup}`, so an unauthenticated user redirected to `/sign-in` rendered inside the full app chrome (`SidebarProvider` + `AppSidebar` + `AppHeader`, which calls `useSession`/`signOut`).

Moved the predicate into a small `components/AppShell/authPaths.ts` keyed on `/sign-in` + `/sign-in/error`, and added `authPaths.test.ts` asserting those are auth paths while `/`, `/balance`, and the old `/login`/`/signup` are not — a pure-function test since the suite runs in the `node` env (no jsdom/testing-library). That guards the route-rename regression you flagged without pulling in a DOM-render harness.

lib/db/index.ts
naeemba: Verified. `lib/db/index.ts` now `export { db } from '@naeemba/next-starter/db'` and `connection.ts` re-exports `createDb as createDbConnection` with `DbInstance = ReturnType<typeof createDb>`. No second `postgres()`+`drizzle()` bootstrap remains, the signature stays compatible (`createDb(url, opts?)`), and `tsc --noEmit` passes with all importers resolving. ✅

lib/templates/service.ts
naeemba: Verified. `isUniqueConflict` now lives in `lib/db/isUniqueConflict.ts` and is imported by both `templates/service.ts` and `savedViews/service.ts`; the two inline copies and the dead `/UNIQUE constraint failed/i` SQLite branch are gone. ✅

lib/env/index.ts
naeemba: Verified. `DATA_DIR: z.string().min(1).default('./data')` is in the schema (an empty `DATA_DIR=` now fails `.min(1)` at parse, while a missing var defaults) and it's documented in `.env.example`. Leaving `layout.ts` on `process.env.DATA_DIR` for per-test override is reasonable. ✅

package.json
naeemba: Verified. Single `"migrate": "next-starter migrate && drizzle-kit migrate"` reused by `predev` (dotenv-wrapped), `prebuild`, and `prestart` — auth `user` track runs before the app FKs, and a build-without-`pnpm start` deploy now applies the app tables too. ✅

lib/test-utils/db.ts
naeemba: Verified. `SCHEMA_SQL` is gone; the harness applies the real package auth migrations via `resolveMigrationsFolder()` then `db/migrations`, splitting on `--> statement-breakpoint`. All 347 tests pass against the shipped schema, confirming the previously-drifted hand-written `user` DDL is no longer a second source of truth. ✅

lib/settings/getBaseCurrency.ts
naeemba: Verified. The `userSettingRepository.get` call is wrapped in try/catch — a DB error logs and falls back to `env.DEFAULT_CURRENCY` instead of rejecting and 500-ing the ~20 pages that read base currency. (`getOptionalUser()` stays outside, which is correct — it's the session lookup, not the flagged DB round-trip.) ✅

lib/savedViews/repository.ts
naeemba: Verified. `$onUpdate(() => sql`now()`)` added to `savedView`/`template`; `update()` is now `.set(patch)` (an empty patch still emits a valid SET because `$onUpdate` injects `updatedAt`), and `upsertBaseCurrency`/`setMainFile` use `sql`now()`. Consistent DB-clock convention throughout — agreed that a host `new Date()` could land `updatedAt` before `createdAt`. ✅

lib/prices/repository.ts
naeemba: The atomicity concern is resolved for the **fetch** path — `pairs` is a `Map` keyed by `symbol|quote` and every row in a run shares one `fetchedDate`, so that batch can't contain a duplicate conflict key.

But the batched upsert introduces a **regression on the legacy-import path**. `maybeMigrateLegacyFiles` → `parseLegacyPriceDb` (`lib/prices/migration.ts`) emits **one row per `P` directive**, and a legacy `price-db.ledger` routinely carries several same-day entries for one commodity (intraday quotes / repeated daily fetches). Those collapse to the same `(symbol, quote, fetchedDate)` key, so the single-statement `INSERT … VALUES (…),(…) … ON CONFLICT DO UPDATE` hits Postgres **`21000` — "ON CONFLICT DO UPDATE command cannot affect row a second time"** and the whole migration throws. The old per-row loop tolerated this (last write won), so this is a behavior change, not just a perf change.

Suggest de-duplicating by conflict key inside `insert()` (keeping the last occurrence, matching the upsert's last-wins intent) before building the statement:

```ts
const deduped = [
  ...new Map(
    rows.map((r) => [`${r.symbol}|${r.quote}|${r.fetchedDate}`, r])
  ).values(),
];
// …
.values(deduped)
```

A migration test feeding two same-day `P` lines for one symbol through `insert()` would lock it in — `migration.test.ts` currently exercises only the parser, not the insert path. Leaving this thread open for the follow-up.

naeemba:
Verified the AppShell fix (0e9a8b7). `components/AppShell/authPaths.ts` keys the centered auth layout on `{/sign-in, /sign-in/error}`, and those are exactly the two routes that exist under `app/sign-in/` — so an unauthenticated user on `/sign-in` no longer renders inside `SidebarProvider`/`AppSidebar`/`AppHeader`. `authPaths.test.ts` asserts the sign-in paths are auth pages and that `/`, `/balance`, and the deleted `/login`//`signup` are not, guarding the route-rename regression. Pulling the predicate into its own tested module is a clean call given the `node` test env. ✅

lib/prices/repository.ts
naeemba: Fixed in 52a28aa. Deduped by conflict key (`symbol|quote|fetchedDate`, last-wins via `new Map(...).values()`) inside `insert()` before building the batched statement, so the legacy-import path no longer hits Postgres 21000 when a `price-db.ledger` carries several same-day `P` directives for one commodity. Put the dedupe in the repo rather than the migration so the invariant holds for every caller. Added a repository test feeding two same-day rows for one symbol through `insert()` and asserting one row survives with the last price — `repository.test.ts` exercises the real insert path (the fetch path was already Map-deduped upstream, so this closes the remaining gap).

naeemba: Verified the dedupe fix (52a28aa). `insert()` now collapses rows through `new Map(rows.map((r) => [\`${r.symbol}|${r.quote}|${r.fetchedDate}\`, r]))` before building the statement — the key is exactly the `onConflictDoUpdate` target `(symbol, quote, fetchedDate)`, and `Map` keeps the **last** value per key, matching the upsert's last-wins intent. So the single `ON CONFLICT DO UPDATE` can no longer target the same row twice, and the legacy-import path (`parseLegacyPriceDb` emits one row per `P` directive with date-only `fetchedDate`, so same-day intraday quotes collide) no longer throws Postgres `21000`.

Confirmed against the code: `migration.ts` sets `fetchedDate: \`${y}-${mo}-${d}\`` (no time component), so multiple same-day `P` lines for one commodity do produce identical conflict keys — exactly the collision this guards.

The new `repository.test.ts` case ("dedupes same-day duplicate conflict keys in one batch (last-wins)") feeds two same-day BTC/EUR rows through the **real** PGlite insert path and asserts one row survives at the later price (61000). All 9 tests in the file pass locally. Putting the dedupe in the repo rather than the migration is the right call — the invariant now holds for every `insert()` caller, not just the legacy import. ✅

PR #20, opened 2026-06-22
feat(landing): add public marketing home page

naeemba:
## What

Adds a public, full-bleed **marketing landing page** at `/` and relocates the authenticated dashboard to `/dashboard`.

Previously `/` rendered the authenticated `Dashboard` inside the app shell. This makes `/` a public page aimed at converting logged-out visitors, while signed-in users are routed to their dashboard.

## Design

Editorial / terminal-inspired aesthetic that honors the plain-text accounting heritage:
- Self-contained deep "ink" dark theme (independent of the app's light/dark tokens), emerald accent reusing the app's "positive" green.
- **Fraunces** display serif paired with **JetBrains Mono** for a real `.ledger` journal block.
- Animated hero (journal → net-worth dashboard mockup with a self-drawing chart), scrolling principles marquee, feature grid mapped to real features (net worth, portfolio, transactions/payees, reconcile, multi-currency, cash flow), how-it-works, philosophy band, and final CTA.
- Staggered load + IntersectionObserver scroll reveals, with a `prefers-reduced-motion` fallback.

## Routing & wiring

- `proxy.ts`: `/` is now public; signed-in visitors are redirected to `/dashboard` via a cheap session-cookie check (no DB), keeping the landing fully decoupled from auth.
- `app/page.tsx` renders the landing; new `app/dashboard/page.tsx` hosts the `Dashboard`.
- `AppShell` renders `/` bare (no sidebar/header) and now takes `BaseCurrencyBanner` as a **server-passed slot** — the client shell no longer imports a server-only (`next/headers`) module, and the banner is scoped to app pages only.
- Dashboard nav item, sidebar brand link, and sign-in `callbackUrl` now point at `/dashboard`.

## Verification

- `pnpm type-check` ✅ · `pnpm lint` ✅ (also enforced by the pre-commit hook)
- Local render: `/` → HTTP 200 with all sections; request with a session cookie → 307 → `/dashboard`.

> Note: this branch was cut from `main` after PR #19 (the Postgres + starter migration) merged, since the landing relies on that auth/starter infrastructure.


proxy.ts
```diff
@@ -19,6 +33,8 @@ export function proxy(req: NextRequest) {
 //   /sign-in and /sign-in/error  — the auth UI itself
 //   /api/auth/*                  — better-auth's own handlers
 //   _next/* internals, favicon, static assets
+// The public landing at `/` still passes through the matcher but is short-
+// circuited above via PUBLIC_PATHS, so it renders without a session.
```
naeemba: The matcher/config comment references `PUBLIC_PATHS`, but no such constant exists — `/` is short-circuited by an inline `req.nextUrl.pathname === '/'` check above, not a `PUBLIC_PATHS` set. Either introduce a `PUBLIC_PATHS` constant (and check membership against it) or fix the comment to match the actual inline check. As written it sends a future reader looking for a constant that isn't there.

Related structure note: the landing-route special-casing is now spread as a magic `'/'` string across two layers (`proxy.ts` here and `AppShell.tsx`). The codebase already centralizes "which path gets which chrome" in the tested `components/AppShell/authPaths.ts` module (whose own comment warns that an un-centralized route decision can "silently strand" a page). Consider following that precedent — a small shared constant/predicate for the public landing path — rather than duplicating the literal in both the proxy and the shell.

components/AppShell/AppShell.tsx
```diff
@@ -14,12 +14,22 @@ import { usePathname } from 'next/navigation';
...
 };
 
-const AppShell = ({ children, headerSlot }: Props) => {
+const AppShell = ({ children, headerSlot, bannerSlot }: Props) => {
   const pathname = usePathname();
   const isAuthPage = isAuthPath(pathname);
 
+  // The marketing landing at `/` owns its own full-bleed chrome — no sidebar,
+  // header, or app-only banners.
+  if (pathname === '/') {
```
naeemba: This adds a third route-gating branch to AppShell (`/` -> bare chrome) but, unlike the sibling `isAuthPath` branch right below it, it has no test. The immediately-preceding commit on this file (`0e9a8b7 fix(shell): gate auth layout on /sign-in routes + test`) established the precedent that chrome-gating decisions get a colocated test, and `authPaths.test.ts` already exists for exactly that. Suggest extracting the public-landing check into a small tested predicate (mirroring `isAuthPath`/`authPaths.ts`) so the bare-landing decision is covered and `/` isn't a magic string duplicated here and in `proxy.ts`. As-is, a regression that strands the landing inside the sidebar shell would ship silently.

features/landing/Landing.tsx
```diff
@@ -0,0 +1,502 @@
...
+        className="lp-layer mx-auto w-full max-w-7xl px-5 py-24 sm:px-8 sm:py-32"
+      >
+        <Reveal>
+          <div className="lp-card lp-card--lit relative overflow-hidden">
+            {/* unsplash texture — degrades to the gradient underneath if it fails */}
+            <div
+              className="absolute inset-0 opacity-[0.16]"
+              style={{
+                backgroundImage:
+                  "url('https://images.unsplash.com/photo-1554260570-e9689a3418b8?auto=format&fit=crop&w=1600&q=70')",
```
naeemba: Hotlinking a production marketing page to `images.unsplash.com` is fragile: it adds an uncontrolled third-party request on the conversion-critical page, is subject to Unsplash rate-limits/ToS for hotlinking, and silently breaks (here it degrades to the gradient, which is good) when offline or behind a strict CSP/network. Prefer self-hosting the texture as a static asset under `public/` (or dropping it). It also can't be optimized by `next/image` as a CSS background. Not blocking since it degrades gracefully, but it's a reliability/best-practice smell on the most important public page.

features/landing/Landing.tsx
```diff
@@ -0,0 +1,502 @@
...
+    body: 'Net worth, balances, portfolio, cash flow, payees — every report you used to build by hand, now interactive and instant.',
+  },
+  {
+    k: '03',
+    title: 'Stay reconciled',
+    body: 'Add entries from templates, reconcile against statements, and keep your books honest from any device.',
+  },
+];
+
+export default function Landing() {
```
naeemba, line 83: nit: This public marketing landing inherits the root layout's generic `metadata` (`title: APP_NAME`, description "NextJS reporting tool for ledger-cli journals"). For a conversion-oriented public page, consider exporting a page-specific `metadata` (marketing-tuned title + description, and likely OpenGraph) from `app/page.tsx` so search/social previews reflect the landing rather than the app's internal description.

app/layout.tsx
```diff
@@ -22,8 +22,10 @@ export default function RootLayout({
   return (
     <html lang="en" className={cn('font-sans', geist.variable)}>
       <body>
-        <AppShell headerSlot={<BaseCurrencyPickerSlot />}>
-          <BaseCurrencyBanner />
+        <AppShell
+          headerSlot={<BaseCurrencyPickerSlot />}
+          bannerSlot={<BaseCurrencyBanner />}
```
naeemba, line 27: Minor: `BaseCurrencyBanner` is an async server component and is still rendered unconditionally here for every route, including the public `/` landing. On `/`, `AppShell` discards `bannerSlot` (the early `pathname === '/'` return), but the server has already executed the component — which calls `getOptionalUser()` (a session/cookie read) on every public landing render before its output is thrown away. It's harmless today (it short-circuits to `null` for logged-out visitors, and signed-in visitors are redirected away from `/` by the proxy), so this is wasted work rather than a bug. If you want the landing fully decoupled from auth as the PR description claims, consider not passing `bannerSlot` on public routes. Non-blocking.

proxy.ts
naeemba: Fixed in c8b912a. Introduced a shared, tested `components/AppShell/publicPaths.ts` module (`isPublicPath`/`PUBLIC_PATHS`, mirroring `authPaths.ts`). The proxy now checks `isPublicPath(pathname)` and the matcher comment now references the real predicate instead of a phantom constant — the `/` literal is no longer duplicated across the proxy and the shell.

components/AppShell/AppShell.tsx
naeemba: Fixed in c8b912a. Extracted the public-landing check into `isPublicPath` (`components/AppShell/publicPaths.ts`), mirroring `isAuthPath`, and added a colocated `publicPaths.test.ts` covering the bare-landing decision. AppShell now calls `isPublicPath(pathname)` instead of a magic `/` literal, so a regression that strands the landing inside the sidebar shell would now fail a test.

features/landing/Landing.tsx
naeemba: Fixed in c265539. Downloaded the texture to `public/landing-texture.jpg` and switched the background to `url('/landing-texture.jpg')`, removing the third-party hotlink. It still degrades to the gradient if it fails.

features/landing/Landing.tsx
naeemba: Fixed in d13dcf2. Exported a page-specific `metadata` from `app/page.tsx` with a marketing-tuned title + description and an OpenGraph block, overriding the root layout's generic app-internal metadata.

app/layout.tsx
naeemba: Rejecting (no code change), as flagged non-blocking. A clean fix isn't available at this layer: the root `app/layout.tsx` is a server component with no access to the request pathname, so it can't conditionally omit `bannerSlot` for `/`. Properly gating the slot would require splitting the landing into its own route group with a separate layout — structural complexity disproportionate to the actual cost here, which is a single cookie read (no DB) on a page that signed-in users never render (the proxy redirects them away from `/` before render, and `BaseCurrencyBanner` short-circuits to `null` for the logged-out audience). Happy to do the route-group split in a follow-up if you'd prefer the landing structurally decoupled.

PR #21, opened 2026-06-23
feat(auth): send magic-link mail via self-hosted Postal

naeemba:
## What

Routes magic-link auth emails through the self-hosted **Postal** server (`postal.raxel.studio`) instead of Resend.

- `lib/email-transport.ts` — `postalTransport`: POSTs the rendered email to Postal's HTTPS API (`/api/v1/send/message`) with an `X-Server-API-Key`.
- `lib/auth.ts` — passes `transport: postalTransport` to `createAuth`, so the starter skips its built-in Resend dispatch (no `RESEND_API_KEY` needed).

## Why

Self-hosting all transactional mail (3 domains) on Postal instead of paying Resend. SPF/DKIM/DMARC verified, delivering to inbox.

## Required env (set in Coolify before/with deploy)

```
POSTAL_API_URL = https://postal.raxel.studio
POSTAL_API_KEY = <Postal "API" credential key>
EMAIL_FROM     = Ledgeria <login@ledgeria.app>
```

⚠️ The transport throws if `POSTAL_API_KEY` is unset — set these env vars **before** the deploy serves sign-in, or magic-link sends will error.

## Test plan

- [ ] Set env vars on the ledgeria.app Coolify app
- [ ] Deploy
- [ ] Trigger a real sign-in → magic link arrives via Postal, lands in inbox

### DIFF
--- a/lib/email-transport.ts
+++ b/lib/email-transport.ts
@@ -1,0 +1,54 @@
+import 'server-only';
+import type { EmailTransport } from '@naeemba/next-starter/email';
+
+/**
+ * Magic-link delivery via the self-hosted Postal server (postal.raxel.studio)
+ * over its HTTPS API. When passed to `createAuth({ transport })`, the starter
+ * skips its built-in Resend dispatch entirely (no RESEND_API_KEY needed).
+ *
+ * Env:
+ *   POSTAL_API_URL  e.g. https://postal.raxel.studio
+ *   POSTAL_API_KEY  a Postal "API" credential key for the mail server
+ */
+export const postalTransport: EmailTransport = async ({
+  to,
+  from,
+  subject,
+  html,
+  text,
+}) => {
+  const baseUrl = process.env.POSTAL_API_URL;
+  const apiKey = process.env.POSTAL_API_KEY;
+  if (!baseUrl || !apiKey) {
+    throw new Error(
+      '[email] POSTAL_API_URL and POSTAL_API_KEY must be set to send mail.'
+    );
+  }
+
+  const res = await fetch(`${baseUrl}/api/v1/send/message`, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      'X-Server-API-Key': apiKey,
+    },
+    body: JSON.stringify({
+      to: [to],
+      from,
+      subject,
+      html_body: html,
+      plain_body: text,
+    }),
+  });
+
+  // Postal returns 200 with { status: 'success' | 'error' | 'parameter-error' }
+  const body = (await res.json().catch(() => null)) as {
+    status?: string;
+    data?: unknown;
+  } | null;
+
+  if (!res.ok || body?.status !== 'success') {
+    throw new Error(
+      `[email] Postal send failed (HTTP ${res.status}): ${JSON.stringify(body)}`
+    );
+  }
+};
--- a/lib/email-transport.ts
+++ b/lib/email-transport.ts
@@ -1,4 +1,5 @@
 import 'server-only';
+import { env } from '@/lib/env';
 import type { EmailTransport } from '@naeemba/next-starter/email';
 
 /**
@@ -17,19 +18,11 @@ export const postalTransport: EmailTransport = async ({
   html,
   text,
 }) => {
-  const baseUrl = process.env.POSTAL_API_URL;
-  const apiKey = process.env.POSTAL_API_KEY;
-  if (!baseUrl || !apiKey) {
-    throw new Error(
-      '[email] POSTAL_API_URL and POSTAL_API_KEY must be set to send mail.'
-    );
-  }
-
-  const res = await fetch(`${baseUrl}/api/v1/send/message`, {
+  const res = await fetch(`${env.POSTAL_API_URL}/api/v1/send/message`, {
     method: 'POST',
     headers: {
       'Content-Type': 'application/json',
-      'X-Server-API-Key': apiKey,
+      'X-Server-API-Key': env.POSTAL_API_KEY,
     },
     body: JSON.stringify({
       to: [to],
--- a/lib/email-transport.ts
+++ b/lib/email-transport.ts
@@ -36,7 +36,6 @@ export const postalTransport: EmailTransport = async ({
   // Postal returns 200 with { status: 'success' | 'error' | 'parameter-error' }
   const body = (await res.json().catch(() => null)) as {
     status?: string;
-    data?: unknown;
   } | null;
 
   if (!res.ok || body?.status !== 'success') {
--- a/lib/env/index.ts
+++ b/lib/env/index.ts
@@ -21,10 +21,12 @@ const envSchema = clientEnvSchema.extend({
   // lib/journal/layout.ts (so tests can override it per-case).
   DATA_DIR: z.string().min(1).default('./data'),
 
-  // Email (magic link). Required for Resend delivery in production; in dev the
-  // link is logged to stdout when RESEND_API_KEY is unset.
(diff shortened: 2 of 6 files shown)


lib/email-transport.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+export const postalTransport: EmailTransport = async ({
+  to,
+  from,
+  subject,
+  html,
+  text,
+}) => {
+  const baseUrl = process.env.POSTAL_API_URL;
+  const apiKey = process.env.POSTAL_API_KEY;
```
naeemba: These two required Postal env vars are read raw via `process.env` and are not added to the centralized, validated env schema in `lib/env/index.ts`. Every other server secret (`BETTER_AUTH_SECRET`, `DATABASE_URL`, `EMAIL_FROM`, even `DATA_DIR`) is declared there specifically so a missing/empty value fails fast at startup rather than silently breaking at runtime. As written, a missing `POSTAL_API_KEY` only surfaces the first time someone tries to sign in (the throw on line 22) — which is exactly the footgun the PR description warns about, and which the schema pattern exists to prevent.

Suggest adding to `envSchema`:
```ts
POSTAL_API_URL: z.string().url(),
POSTAL_API_KEY: z.string().min(1),
```
then `import { env } from '@/lib/env'` and read `env.POSTAL_API_URL` / `env.POSTAL_API_KEY` here. That also lets you drop the inline `if (!baseUrl || !apiKey)` guard.

Related cleanup (same root): now that delivery goes through Postal and `createAuth` no longer uses Resend, `RESEND_API_KEY` (and the `EMAIL_FROM` Resend-oriented comment) in `lib/env/index.ts` are dead — consider removing `RESEND_API_KEY` in this PR so the schema reflects reality.

naeemba: Fixed in e830b54. Added `POSTAL_API_URL` (url) and `POSTAL_API_KEY` (min 1) to the validated `envSchema`, switched the transport to read `env.POSTAL_API_URL` / `env.POSTAL_API_KEY`, and dropped the inline `if (!baseUrl || !apiKey)` guard. Also removed the now-dead `RESEND_API_KEY` and updated the Resend-oriented `EMAIL_FROM` comment to reflect Postal delivery.

lib/env/index.ts
```diff
@@ -21,10 +21,12 @@ const envSchema = clientEnvSchema.extend({
...
 
-  // Email (magic link). Required for Resend delivery in production; in dev the
-  // link is logged to stdout when RESEND_API_KEY is unset.
+  // Email (magic link). Delivered via the self-hosted Postal server (see
+  // lib/email-transport.ts). Both Postal vars are required so a missing/empty
+  // value fails fast at startup rather than the first time someone signs in.
   EMAIL_FROM: z.string().default('auth@example.com'),
-  RESEND_API_KEY: z.string().optional(),
+  POSTAL_API_URL: z.string().url(),
+  POSTAL_API_KEY: z.string().min(1),
```
naeemba, line 29: Making these two vars **required** (non-optional) breaks the test suite. `vitest-setup.ts` exists precisely to seed every required var that `lib/env` validates at module-load time (it already sets `BETTER_AUTH_SECRET`, `DATABASE_URL`, `DATA_DIR`, `BETTER_AUTH_URL`), but this PR adds `POSTAL_API_URL`/`POSTAL_API_KEY` to the schema without seeding them there.

Reproduced with a clean env (`env -i PATH=$PATH HOME=$HOME npx vitest run`): 6 test files fail to load with

```
Invalid environment configuration. Fix the following:
  - POSTAL_API_URL: Invalid input: expected string, received undefined
  - POSTAL_API_KEY: Invalid input: expected string, received undefined
```

via the import chain `lib/auth.ts -> lib/email-transport.ts -> lib/env/index.ts`. (It passes locally only when the maintainer's own shell exports these; CI / a fresh clone will fail.) This matches the established pattern violation: every other required var got a corresponding default in `vitest-setup.ts`.

Fix: add to `vitest-setup.ts`:
```ts
process.env.POSTAL_API_URL =
  process.env.POSTAL_API_URL ?? 'https://postal.test';
process.env.POSTAL_API_KEY = process.env.POSTAL_API_KEY ?? 'test-key';
```

lib/env/index.ts
```diff
@@ -21,10 +21,12 @@ const envSchema = clientEnvSchema.extend({
...
   DATA_DIR: z.string().min(1).default('./data'),
 
-  // Email (magic link). Required for Resend delivery in production; in dev the
-  // link is logged to stdout when RESEND_API_KEY is unset.
+  // Email (magic link). Delivered via the self-hosted Postal server (see
+  // lib/email-transport.ts). Both Postal vars are required so a missing/empty
+  // value fails fast at startup rather than the first time someone signs in.
   EMAIL_FROM: z.string().default('auth@example.com'),
-  RESEND_API_KEY: z.string().optional(),
+  POSTAL_API_URL: z.string().url(),
```
naeemba, line 28: `.env.example` is now stale and out of sync with this schema change (it isn't in the diff, so flagging here at the source of the drift). It still lists the removed `RESEND_API_KEY` and the old "optional in dev (link logs to stdout when RESEND_API_KEY unset)" comment, and does NOT document the two new **required** `POSTAL_API_URL` / `POSTAL_API_KEY` vars. Since they're now non-optional, anyone copying `.env.example` -> `.env` for local dev hits a hard startup failure with no hint of what to set. Please update `.env.example`: drop `RESEND_API_KEY`, and add `POSTAL_API_URL=https://postal.raxel.studio` and `POSTAL_API_KEY=` under the Email section.

lib/email-transport.ts
```diff
@@ -0,0 +1,47 @@
...
+      subject,
+      html_body: html,
+      plain_body: text,
+    }),
+  });
+
+  // Postal returns 200 with { status: 'success' | 'error' | 'parameter-error' }
+  const body = (await res.json().catch(() => null)) as {
+    status?: string;
+    data?: unknown;
```
naeemba: nit: the inline response type declares `data?: unknown` but only `status` is ever read — drop `data` to keep the shape minimal (`{ status?: string } | null`). Also, this transport has real branching (HTTP non-ok, non-JSON body, `status !== 'success'`) and currently has no test; a small unit test fetch-mocking Postal's success and error responses would lock in the error-throwing behavior. Not blocking given the repo only enforces coverage under `lib/journal/**`.

lib/env/index.ts
naeemba: Fixed in 2196ca3. Seeded `POSTAL_API_URL=https://postal.test` and `POSTAL_API_KEY=test-key` in `vitest-setup.ts` (idempotent `?? ` defaults, matching the other required vars). Verified the repro: `env -i PATH=$PATH HOME=$HOME npx vitest run` now loads all files instead of failing collection on the missing Postal vars.

lib/env/index.ts
naeemba: Fixed in b4e6590. Updated `.env.example`: dropped `RESEND_API_KEY` and its stale 'optional in dev' comment, and added the now-required `POSTAL_API_URL=https://postal.raxel.studio` / `POSTAL_API_KEY=` under the Email section.

lib/email-transport.ts
naeemba: Partially fixed in 9945071. Dropped the unused `data` field — the inline response type is now `{ status?: string } | null`. Leaving the unit test out for now since, as you noted, the repo only enforces coverage under `lib/journal/**` and this isn't blocking; happy to add a fetch-mocked success/error test in a follow-up if you'd prefer it in this PR.

PR #23, opened 2026-06-24
feat(storage): Garage object storage as source of truth for ledger files

naeemba:
## Summary

Makes **Garage** (self-hosted, S3-compatible object storage) the source of truth for ledger journal files. Local disk under `DATA_DIR` becomes an ephemeral cache. This unblocks stateless Coolify containers — journal files (the real financial data) now have a durable home instead of relying on container-local disk.

The `ledger` CLI only reads local filesystem paths, so the design bridges object storage ↔ local disk around every invocation:
- **Reads** pull the canonical journal into the local cache first (mirroring the `journals/<userId>/` prefix via `ListObjectsV2` + ETags, downloading only changed objects), then run `ledger`. The query cache key switches from file mtime to a content **fingerprint**.
- **Writes** do `pull → mutate → verify (ledger CLI) → push`, rolling the local mutation back on any verify/push failure so the cache is never left ahead of canonical. A lost-update guard (`StorageConflictError`) protects against concurrent overwrites.
- **Imports** replace canonical atomically (pull → reset local files keeping the manifest → write → verify → guarded push with orphan-delete) — no window where canonical is empty.

All Garage actions live in a single new `lib/storage/` directory. Backend is env-selected: `STORAGE_BACKEND=s3|memory` (default `memory`, so tests/dev/build need no infra; production sets `s3`).

## What's included
- `lib/storage/`: `ObjectStore` interface + `S3ObjectStore` (Garage) and `MemoryObjectStore` (tests/dev) backends, env-driven factory, manifest + content fingerprint, `download`(pull)/`save`(push)/`sync` public API.
- Read-path wiring: `runLedger` + `Transactions` keyed on the fingerprint; `getMaxMtime` → `getFingerprint`.
- Write-path wiring: `addTransaction`/`performEdit`/`performDelete`/imports sync to Garage, all serialized under the per-user lock.
- Env schema + `.env.example` (`STORAGE_BACKEND`, `S3_*`, conditional validation).
- Deployment runbook (`docs/deployment/garage.md`) + `deploy/garage/garage.toml` for standing Garage up on the new-raxel Coolify host.
- Design spec + implementation plan under `docs/superpowers/`.

## Security
Path-traversal hardening: `relPathFromKey` rejects absolute/`..` keys and `download.ts` re-checks containment before any write, so a malformed/compromised object key can't escape the user's journal dir (mirrors the existing ZIP-import guards).

## Testing
- 73 files / **379 tests** pass; type-check and lint clean.
- New tests cover: ETag-diff pull, lost-update conflict, serve-stale degraded read, path-traversal rejection, fingerprint cache key, and atomic import orphan-removal.
- Existing suite runs on the in-memory backend (`resetObjectStore()` isolation); no live Garage needed for CI.
- `pnpm build` and the live-Garage smoke test (`docs/deployment/garage.md`) are manual/CI gates (build's prebuild migrate needs Postgres).

### DIFF
--- a/features/transactions/Transactions.tsx
+++ b/features/transactions/Transactions.tsx
@@ -7,7 +7,7 @@ import {
 } from './applyTransactionFilters';
 import Help from '@/components/Help';
 import { requireUser } from '@/lib/auth/require-user';
-import { journalRepository, journalService } from '@/lib/journal';
+import { journalRepository } from '@/lib/journal';
 import { getJournalCacheTag } from '@/lib/journal/layout';
 import { type Transaction } from '@/lib/journal/parser';
 import { savedViewService } from '@/lib/savedViews';
@@ -16,7 +16,10 @@ import { unstable_cache } from 'next/cache';
 const buildLoader = (tag: string, fingerprint: string) =>
   unstable_cache(
     async (userId: string): Promise<Transaction[]> => {
-      const journal = await journalService.listTransactions(userId);
+      // getFingerprint (below) already pulled the canonical journal into the
+      // local cache, so read straight from the repository — going through
+      // journalService.listTransactions would pull a second time.
+      const journal = await journalRepository.list(userId);
       return journal.transactions;
     },
     ['journal-transactions', tag, fingerprint],
--- a/features/transactions/Transactions.tsx
+++ b/features/transactions/Transactions.tsx
@@ -13,19 +13,19 @@ import { type Transaction } from '@/lib/journal/parser';
 import { savedViewService } from '@/lib/savedViews';
 import { unstable_cache } from 'next/cache';
 
-const buildLoader = (tag: string, mtimeMs: number) =>
+const buildLoader = (tag: string, fingerprint: string) =>
   unstable_cache(
     async (userId: string): Promise<Transaction[]> => {
       const journal = await journalService.listTransactions(userId);
       return journal.transactions;
     },
-    ['journal-transactions', tag, String(mtimeMs)],
+    ['journal-transactions', tag, fingerprint],
     { revalidate: 60, tags: [tag] }
   );
 
 const loadTransactions = async (userId: string) => {
-  const mtimeMs = await journalRepository.getMaxMtime(userId);
-  return buildLoader(getJournalCacheTag(userId), mtimeMs)(userId);
+  const fingerprint = await journalRepository.getFingerprint(userId);
+  return buildLoader(getJournalCacheTag(userId), fingerprint)(userId);
 };
 
 const Transactions = async ({
--- a/lib/journal/repository.ts
+++ b/lib/journal/repository.ts
@@ -2,14 +2,10 @@ import { promises as fs } from 'fs';
 import path from 'path';
 import { eq, sql } from 'drizzle-orm';
 import { DEFAULT_MAIN, PRICE_DB_NAME, getJournalDir } from './layout';
-import {
-  parseJournal,
-  resolveIncludes,
-  type ParsedJournal,
-  type Transaction,
-} from './parser';
+import { parseJournal, type ParsedJournal, type Transaction } from './parser';
 import { userSetting } from '@/db/schema';
 import type { DbInstance } from '@/lib/db/connection';
+import { pull } from '@/lib/storage';
 
 export type JournalLayout = {
   dir: string;
@@ -108,14 +104,13 @@ export class JournalRepository {
     return transactions.find((t) => t.uid === uid) ?? null;
   }
 
-  /** Returns max mtimeMs across the user's include graph. Used as a cache-key
-   * input so any file change (internal or external) invalidates `unstable_cache`. */
-  async getMaxMtime(userId: string): Promise<number> {
-    const { mainPath } = await this.ensureLayout(userId);
-    const files = await resolveIncludes(mainPath);
-    if (files.length === 0) return 0;
-    const stats = await Promise.all(files.map((f) => fs.stat(f)));
-    return Math.max(...stats.map((s) => s.mtimeMs));
+  /** Pulls the canonical journal into the local cache and returns the content
+   * fingerprint. Used as the query cache-key input so any change (local or in
+   * Garage) invalidates `unstable_cache`. Also guarantees the local stub exists. */
+  async getFingerprint(userId: string): Promise<string> {
+    const { fingerprint } = await pull(userId);
+    await this.ensureLayout(userId);
+    return fingerprint;
   }
 
   private async findPriceDb(dir: string): Promise<string | null> {
--- a/lib/journal/repository.ts
+++ b/lib/journal/repository.ts
@@ -5,7 +5,7 @@ import { DEFAULT_MAIN, PRICE_DB_NAME, getJournalDir } from './layout';
 import { parseJournal, type ParsedJournal, type Transaction } from './parser';
 import { userSetting } from '@/db/schema';
 import type { DbInstance } from '@/lib/db/connection';
-import { pull, manifestRelName } from '@/lib/storage';
+import { pullLocked, manifestRelName } from '@/lib/storage';
 
 export type JournalLayout = {
   dir: string;
@@ -115,7 +115,7 @@ export class JournalRepository {
    * fingerprint. Used as the query cache-key input so any change (local or in
    * Garage) invalidates `unstable_cache`. Also guarantees the local stub exists. */
   async getFingerprint(userId: string): Promise<string> {
-    const { fingerprint } = await pull(userId);
+    const { fingerprint } = await pullLocked(userId);
     await this.ensureLayout(userId);
     return fingerprint;
   }
--- a/lib/journal/repository.ts
+++ b/lib/journal/repository.ts
@@ -5,7 +5,7 @@ import { DEFAULT_MAIN, PRICE_DB_NAME, getJournalDir } from './layout';
 import { parseJournal, type ParsedJournal, type Transaction } from './parser';
 import { userSetting } from '@/db/schema';
 import type { DbInstance } from '@/lib/db/connection';
-import { pull } from '@/lib/storage';
+import { pull, clearRemote } from '@/lib/storage';
 
 export type JournalLayout = {
   dir: string;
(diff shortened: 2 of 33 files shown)


lib/storage/download.ts
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Mirrors the user's remote prefix into the local journal dir. Downloads only
+ * objects whose ETag differs from the local manifest, deletes local files no
+ * longer present remotely, rewrites the manifest, and returns the fingerprint.
+ *
+ * If the store is unreachable but a manifest exists, serves the stale local
+ * cache (warn + fingerprint from manifest). With no manifest, the error
+ * propagates.
+ */
+export const pullToLocal = async (
```
naeemba, line 24: **Concurrent reads race on the shared local journal dir (no lock on the pull path).**

`pullToLocal` mutates the per-user local cache in place: it downloads changed objects, `fs.rm`s local files that are no longer in `remoteRelSet`, and rewrites `.manifest.json`. Writes serialize this via `withUserLock`, but the read path does not — `getFingerprint` (runLedger), `listTransactions`, and `findTransaction` all call `pull(userId)` outside any lock.

A single Dashboard render fires 8+ `runLedger` calls through `Promise.all` (see `features/dashboard/Dashboard.tsx`), and each one calls `getFingerprint` → `pull` → `pullToLocal` concurrently against the **same** directory. Interleavings produce:
- torn `.manifest.json` (see separate comment on `writeManifest` non-atomicity),
- one pull deleting a file another pull is mid-download on,
- `ledger` reading a half-written `main.ledger`.

A concurrent write's `pushFromLocal` can also interleave with an in-flight read pull on the same files.

Suggest serializing the pull under the same per-user lock (`withUserLock`) used by writes, or coalescing the per-request pulls into one. At minimum the read-path pull must hold the lock so it can't interleave with a write's push or with another read's pull.

lib/storage/manifest.ts
```diff
@@ -0,0 +1,61 @@
...
+    return {};
+  }
+};
+
+export const writeManifest = async (
+  userId: string,
+  m: Manifest
+): Promise<void> => {
+  await fs.mkdir(getJournalDir(userId), { recursive: true });
+  await fs.writeFile(manifestPath(userId), JSON.stringify(m, null, 2), 'utf-8');
```
naeemba: **`writeManifest` is a non-atomic write — a crash or concurrent pull tears the manifest.**

Everywhere else the journal cache uses `writeFileAtomic` (tmp + rename, see `JournalRepository.writeFileAtomic`) precisely so a reader never sees a half-written file. The manifest, which is the lost-update guard's source of truth, is written with a plain `fs.writeFile`. Two concerns:

1. If two pulls run concurrently (see the lock comment on `download.ts`), their `fs.writeFile` calls to the same `.manifest.json` interleave and can leave truncated/interleaved JSON; `readManifest`'s `JSON.parse` then throws and is swallowed by the `catch {}`, silently resetting the manifest to `{}` — which disables the conflict guard and makes the next push blindly overwrite canonical.
2. A process crash mid-write leaves a partial manifest with the same effect.

Use the existing atomic-write pattern (tmp file + rename) here too, and consider reusing `JournalRepository.writeFileAtomic` rather than re-implementing the bare write.

lib/storage/save.ts
```diff
@@ -0,0 +1,86 @@
...
+    for (const e of entries) {
+      const childRel = rel ? path.join(rel, e.name) : e.name;
+      if (e.isDirectory()) await walk(path.join(abs, e.name), childRel);
+      else if (e.name !== manifestRelName && !e.name.endsWith('.tmp'))
+        out.push(childRel);
+    }
+  };
+  await walk(dir, '');
+  return out;
+};
```
naeemba: **Duplicated `listLocalRelPaths` helper.**

This exact recursive walker (same name, same exclusion of `manifestRelName` and `*.tmp`) is defined again verbatim in `lib/storage/download.ts` (lines 17–37). Hoist it into `manifest.ts` (or a `lib/storage/local.util.ts`) and import it in both `save.ts` and `download.ts` so the two copies can't drift.

features/transactions/Transactions.tsx
```diff
@@ -13,19 +13,19 @@ import { type Transaction } from '@/lib/journal/parser';
...
-    ['journal-transactions', tag, String(mtimeMs)],
+    ['journal-transactions', tag, fingerprint],
     { revalidate: 60, tags: [tag] }
   );
 
 const loadTransactions = async (userId: string) => {
-  const mtimeMs = await journalRepository.getMaxMtime(userId);
-  return buildLoader(getJournalCacheTag(userId), mtimeMs)(userId);
+  const fingerprint = await journalRepository.getFingerprint(userId);
+  return buildLoader(getJournalCacheTag(userId), fingerprint)(userId);
```
naeemba, line 31: nit: **Redundant pull on the read path — every cache miss pulls twice.**

`loadTransactions` calls `journalRepository.getFingerprint(userId)` (which does a full `pull`) to build the cache key, then on a miss the loader calls `journalService.listTransactions(userId)` which pulls *again*. `runLedger` has the same shape (`getFingerprint` pulls, then `getLayout`). A pull is a `ListObjectsV2` round-trip plus per-object ETag diffing, so this doubles remote I/O on every cache miss. Consider having `getFingerprint` return the fingerprint and letting downstream reads reuse the already-pulled cache rather than pulling a second time inside `listTransactions`/`findTransaction`.

naeemba:
Initial review of PR #23 (Garage object storage as source of truth).

The storage layer is well-structured (clean ObjectStore interface, memory/s3 backends, env-gated factory, path-traversal hardening that mirrors the existing ZIP guards, good test coverage). My main concern is concurrency on the **read** path:

1. **No lock on the pull path (download.ts).** Reads call `pull` outside `withUserLock`; a single Dashboard render fans out 8+ `runLedger` calls via `Promise.all`, each pulling concurrently against the same local journal dir (download + delete + manifest rewrite). This races reads against each other and against a concurrent write's push.
2. **`writeManifest` is non-atomic (manifest.ts).** A torn/partial manifest is swallowed by `readManifest`'s `catch {}` → silently resets to `{}` → disables the lost-update guard. Use the existing tmp+rename atomic write.
3. **Duplicated `listLocalRelPaths` (save.ts / download.ts).** Hoist to a shared util.

Plus one nit on redundant double-pulling per cache miss.

Nothing blocking the design — the lock/atomicity items are the ones worth fixing before this carries real financial data.

lib/storage/download.ts
naeemba: Fixed in 45b1dfe. Added a lock-acquiring `pullLocked` in lib/storage/sync.ts and routed the read paths through it: `getFingerprint` (repository.ts, used by runLedger and the Transactions loader), `listTransactions`, and `findTransaction` (service.ts). They now serialize under the same per-user `withUserLock` the write paths use, so a read's pull can no longer interleave with another pull, with a concurrent write's push, or with a manifest write. The raw `pull` is kept for the write paths only — they already hold the lock and the mutex is non-reentrant, so locking inside `pull` itself would deadlock them.

lib/storage/manifest.ts
naeemba: Fixed in 668e9b5. `writeManifest` now writes atomically (tmp + rename on the same filesystem), matching the rest of the journal cache, so a crash or a concurrent reader can never observe a half-written `.manifest.json` (which would fail `JSON.parse`, reset to {}, and disable the conflict guard). The `*.tmp` scratch file is already excluded by `listLocalRelPaths`, so it won't leak into uploads or deletions.

lib/storage/save.ts
naeemba: Fixed in 668e9b5. Hoisted `listLocalRelPaths` into lib/storage/manifest.ts (single source of truth) and removed the verbatim copies from `download.ts` and `save.ts`, which now import it.

features/transactions/Transactions.tsx
naeemba: Fixed in 0f09400. The Transactions loader now reads straight from `journalRepository.list(userId)` instead of `journalService.listTransactions` — `getFingerprint` already pulled the canonical journal to build the cache key, so the second pull was redundant. `findTransaction` and the export route still call the service passthroughs (they have no prior `getFingerprint`), so those keep their own pull.

lib/journal/repository.ts
```diff
@@ -108,14 +111,13 @@ export class JournalRepository {
...
-  async getMaxMtime(userId: string): Promise<number> {
-    const { mainPath } = await this.ensureLayout(userId);
-    const files = await resolveIncludes(mainPath);
-    if (files.length === 0) return 0;
-    const stats = await Promise.all(files.map((f) => fs.stat(f)));
-    return Math.max(...stats.map((s) => s.mtimeMs));
+  /** Pulls the canonical journal into the local cache and returns the content
+   * fingerprint. Used as the query cache-key input so any change (local or in
+   * Garage) invalidates `unstable_cache`. Also guarantees the local stub exists. */
+  async getFingerprint(userId: string): Promise<string> {
```
naeemba, line 130: **`getFingerprint` is not request-deduped — a single render fires N serialized `ListObjectsV2` pulls.**

The read-path lock fix (45b1dfe) routes every read's pull through `pullLocked`, which now correctly serializes them. But `Dashboard.tsx` fires ~8 concurrent reads through `Promise.all` (7× `runLedger` + `getRecentTransactions`/`getJournalStats`), and each calls `journalRepository.getFingerprint` → `pullLocked` → a full `ListObjectsV2` + per-object ETag diff. Post-fix these no longer race, but they now run **back-to-back under the lock**: one dashboard render does ~8 sequential remote list round-trips that all return identical data, on the critical render path.

Wrap `getFingerprint` in React's `cache()` so the concurrent callers in one request coalesce to a single pull. There's precedent in the codebase (`lib/settings/getBaseCurrency.ts`, `getAvailableCurrencies.ts`, `getMissingRateCommodities.ts` all use `import { cache } from 'react'` for exactly this request-scoped dedup). This collapses the N pulls per render to one and also removes the self-contention the lock now introduces.

naeemba: Fixed in c9baf43. Wrapped the read-path pull in React's `cache()` keyed by userId (`cachedPull`), so the ~8 concurrent `getFingerprint` callers in one render coalesce to a single `pullLocked` -> `ListObjectsV2` round-trip instead of running back-to-back under the read lock. Followed the existing `import { cache } from 'react'` precedent (getBaseCurrency etc.). Tests + lint + tsc green.

PR #24, opened 2026-06-24
feat(auth): studio-grade passwordless auth + multi-user signup

naeemba:
## What & why

Replaces the minimal starter sign-in screen with a bespoke, studio-grade **passwordless** auth experience, adds a matching **sign-up** screen, and **opens the app to multi-user self-registration**. Forms are email-only (passwordless magic link + passkey + optional Google) — no unnecessary data to fill in.

Design spec: `docs/superpowers/specs/2026-06-24-auth-forms-redesign-design.md`
Implementation plan: `docs/superpowers/plans/2026-06-24-auth-forms-redesign.md`

## Changes

- **New `features/auth/` module**
  - Pure, unit-tested logic: `callbackUrl.ts` (open-redirect–safe same-origin callbacks), `authCopy.ts` (mode-aware copy incl. the check-spam warning), `authState.ts` (per-method status reducer + 30s resend cooldown).
  - Presentational: `SentNotice.tsx` (“Check your inbox” + **spam/junk warning** + resend), `BrandPanel.tsx` (split-editorial brand panel).
  - Interactive: `AuthForm.tsx` (wires `@/lib/auth-client`), `AuthScreen.tsx` (responsive split layout, collapses to a centered card on mobile, dark-mode aware).
- **Routes:** `app/sign-in/page.tsx` rewritten as a thin shell (drops the starter `SignInPage`); new `app/sign-up/page.tsx`.
- **Multi-user:** removed the `singleAdmin` allowlist in `lib/auth.ts`. The data layer is already per-user keyed (`ensureLayout(userId)` auto-provisions each user's journal), so opening registration required no data-isolation work.

## Spam warning (requested)

After a magic link is sent, the form shows: *"Don't see it? Check your spam or junk folder — and add our address to your contacts."* plus a resend with a 30s cooldown and a "use a different email" escape.

## Review notes

Built via subagent-driven development: each task TDD'd + independently reviewed, then a final whole-branch review. Two bugs were caught and fixed in review:
- resend cooldown was never actually enforced (timer state never disabled);
- passkey button wasn't hidden on browsers without WebAuthn (now gated via `useWebAuthnSupported()`, along with the "or" divider to avoid an orphaned separator).

## Known limitations

- **Sign-up name field dropped:** the installed better-auth `magicLink` type doesn't accept `name`, so sign-up is email-only (pre-authorized fallback in the spec). `authCopy` keeps a `showNameField` placeholder for when the signature supports it.
- **Fully open registration:** a deployed instance accepts any sign-up once merged. No rate-limit/CAPTCHA yet — revisit if abuse appears.
- A few deferred Minor cleanups are tracked in the SDD ledger (dead `canResend`/`showNameField`, duplicated "10 minutes" copy).

## Testing

- type-check clean, lint clean, **399 tests pass** (79 files), incl. static-markup component tests (node-env Vitest, no new test deps).
- ⚠️ **Manual dev-server smoke test not run** — sandbox has no local Postgres (`pnpm dev` predev migrate needs `DATABASE_URL`). Please run locally to verify visuals / dark mode / responsive collapse and the live magic-link/passkey flows.

### DIFF
--- a/features/auth/SentNotice.tsx
+++ b/features/auth/SentNotice.tsx
@@ -1,0 +1,52 @@
+'use client';
+
+import { MailCheck } from 'lucide-react';
+import { sentCopy } from './authCopy';
+import { Button } from '@/components/ui/button';
+
+interface SentNoticeProps {
+  email: string;
+  canResend: boolean;
+  onResend: () => void;
+  onUseDifferentEmail: () => void;
+}
+
+export function SentNotice({
+  email,
+  canResend,
+  onResend,
+  onUseDifferentEmail,
+}: SentNoticeProps) {
+  const copy = sentCopy(email);
+  return (
+    <div className="flex flex-col gap-4 text-center" aria-live="polite">
+      <div className="mx-auto flex size-12 items-center justify-center rounded-full bg-muted">
+        <MailCheck className="size-6 text-foreground" aria-hidden />
+      </div>
+      <div className="space-y-1">
+        <h2 className="text-lg font-semibold">{copy.heading}</h2>
+        <p className="text-sm text-muted-foreground">
+          We sent a sign-in link to{' '}
+          <span className="font-medium text-foreground">{email}</span>. It
+          expires in 10 minutes.
+        </p>
+      </div>
+      <p className="rounded-lg bg-muted/60 px-3 py-2 text-xs text-muted-foreground">
+        {copy.spam}
+      </p>
+      <div className="flex flex-col gap-2">
+        <Button
+          type="button"
+          variant="outline"
+          onClick={onResend}
+          disabled={!canResend}
+        >
+          Resend link
+        </Button>
+        <Button type="button" variant="ghost" onClick={onUseDifferentEmail}>
+          Use a different email
+        </Button>
+      </div>
+    </div>
+  );
+}
--- a/features/auth/SentNotice.tsx
+++ b/features/auth/SentNotice.tsx
@@ -17,7 +17,7 @@ export function SentNotice({
   onResend,
   onUseDifferentEmail,
 }: SentNoticeProps) {
-  const copy = sentCopy(email);
+  const copy = sentCopy();
   return (
     <div className="flex flex-col gap-4 text-center" aria-live="polite">
       <div className="mx-auto flex size-12 items-center justify-center rounded-full bg-muted">
--- a/features/auth/authCopy.ts
+++ b/features/auth/authCopy.ts
@@ -4,7 +4,6 @@ export interface AuthCopy {
   heading: string;
   subheading: string;
   submitLabel: string;
-  showNameField: boolean;
   altPrompt: string;
   altLinkLabel: string;
   altHref: string;
@@ -18,7 +17,6 @@ const COPY: Record<AuthMode, AuthCopy> = {
     heading: 'Welcome back',
     subheading: 'Sign in with a magic link, passkey, or Google.',
     submitLabel: 'Send magic link',
-    showNameField: false,
     altPrompt: 'New here?',
     altLinkLabel: 'Sign up',
     altHref: '/sign-up',
@@ -27,7 +25,6 @@ const COPY: Record<AuthMode, AuthCopy> = {
     heading: 'Create your account',
     subheading: 'No password needed — we email you a secure sign-in link.',
     submitLabel: 'Create account',
-    showNameField: true,
     altPrompt: 'Already have an account?',
     altLinkLabel: 'Sign in',
     altHref: '/sign-in',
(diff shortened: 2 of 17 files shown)


features/auth/authState.ts
```diff
@@ -0,0 +1,53 @@
...
+        errors: { ...state.errors, [action.method]: action.message },
+      };
+    case 'reset':
+      return initialAuthState;
+  }
+}
+
+export const RESEND_COOLDOWN_MS = 30_000;
+
+export function canResend(lastSentAt: number | null, now: number): boolean {
```
naeemba, line 50: nit: `canResend` (and `RESEND_COOLDOWN_MS` as its sole consumer) is dead — `AuthForm` reimplements the cooldown imperatively with two `setTimeout`s driving `resendAllowed` (AuthForm.tsx:38-50) and never calls this pure predicate. You now have two parallel cooldown implementations; the timer-based one in the component is the one that actually gates the button. Consider deriving `resendAllowed` from `canResend(state.lastSentAt, now)` (e.g. tick `now` on an interval, or compute it in the effect) so the tested pure function is the single source of truth, or drop `canResend` if the imperative version is intended to stay. The PR description already flags this as deferred.

features/auth/SentNotice.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const copy = sentCopy(email);
+  return (
+    <div className="flex flex-col gap-4 text-center" aria-live="polite">
+      <div className="mx-auto flex size-12 items-center justify-center rounded-full bg-muted">
+        <MailCheck className="size-6 text-foreground" aria-hidden />
+      </div>
+      <div className="space-y-1">
+        <h2 className="text-lg font-semibold">{copy.heading}</h2>
+        <p className="text-sm text-muted-foreground">
+          We sent a sign-in link to{' '}
```
naeemba, line 29: nit: this inlines the same sentence that `sentCopy(email).body` already produces ("We sent a sign-in link to {email}. It expires in 10 minutes."), so `copy.body` and `copy.expires` are computed but unused, and the "10 minutes" / "sign-in link" strings are now spelled in two places (here and authCopy.ts:44/46). Render `{copy.body}` (or drop the unused fields from `sentCopy`) so the copy lives in one module. Acknowledged as a deferred cleanup in the PR description.

features/auth/authCopy.ts
```diff
@@ -0,0 +1,48 @@
+export type AuthMode = 'sign-in' | 'sign-up';
+
+export interface AuthCopy {
+  heading: string;
+  subheading: string;
+  submitLabel: string;
+  showNameField: boolean;
```
naeemba: nit: `showNameField` is part of the public `AuthCopy` type and set to `true` for sign-up, but no consumer reads it — `AuthForm` renders an email-only form and never branches on it. Given the documented limitation that the installed `magicLink` signature can't accept `name`, this field is currently dead. Either drop it until the signature supports a name (keeps the type honest) or add a `// reserved for…` comment so a future reader doesn't assume the name field is wired. Acknowledged in the PR's Known limitations.

naeemba (commented):
Initial review of PR #24. Overall this is clean, well-tested work that faithfully ports the starter's auth patterns (`resolveCallbackUrl`, passkey `window.location.assign` on success, the `useWebAuthnSupported` gate) into a bespoke `features/auth/` module. Verified the starter does NOT export those callback helpers, so the local port is justified (not a duplicate of package API). The `singleAdmin` removal correctly opens registration (the starter's allowlist is gated solely on that option) and the 600s magic-link expiry matches the "10 minutes" copy. A few cleanups around dead/duplicated copy fields below — all non-blocking nits, most already acknowledged in the PR description.

features/auth/authState.ts
naeemba: Fixed in 503fd81 — the resend effect now derives `resendAllowed` from `canResend(lastSentAt, Date.now())` (both the immediate-disable and the delayed-enable timers route through it), so the tested pure predicate is the single source of truth. `RESEND_COOLDOWN_MS` is now consumed only to compute the timer delay.

features/auth/SentNotice.tsx
naeemba: Fixed in 503fd81 — dropped the unused `body`/`expires` fields from `sentCopy()`. The sign-in-link / "10 minutes" sentence now lives only in `SentNotice` (which keeps the bolded email span), so the copy is no longer duplicated.

features/auth/authCopy.ts
naeemba: Fixed in 503fd81 — removed the dead `showNameField` from the `AuthCopy` type and both copy entries until the installed `magicLink` signature can accept a name.

PR #26, opened 2026-06-24
feat(account): self-service account deletion + journal backup export

naeemba:
Phase 7 self-service **account deletion** plus a journal **backup export** (both were unchecked Phase 7 items). Built via spec → 10-task plan → subagent-driven TDD; full whole-branch review passed.

## Flow
On `/settings`, a red **Danger Zone**: download a `.zip` backup → hard irreversible notice → emailed **6-digit verification code** → confirm → permanent wipe → sign out → `/account/deleted`.

## What it wipes (order is load-bearing)
1. **Garage** objects (`clearRemote` — source of truth first)
2. **Local** journal dir (`rm -rf {DATA_DIR}/journals/{userId}`)
3. **`db.delete(user)`** — FK-cascades `session`/`account`/`passkey` + `userSetting`/`savedView`/`template`/`accountDeletionChallenge`

A mid-failure leaves only inert orphans (data with no owning user); re-running completes the wipe.

## Verification code
SHA-256 hashed at rest, constant-time compared, 10-min expiry, 5-attempt cap (then invalidated), 30s resend throttle — all enforced server-side. The purge is unreachable without a valid, non-expired challenge.

## Notable pieces
- New `accountDeletionChallenge` table + migration (PK = userId, FK cascade).
- `lib/account-deletion/` — repository + service (issue/verify state machine) + `purgeUserData` orchestration + Zod schema, with the project's Repository+Service convention.
- `GET /api/account/export` — streams the user's journal dir as a `.zip` (reuses `pullLocked` + `listLocalRelPaths` + `adm-zip`). *(Also delivers the Phase 7 backup item.)*
- `features/settings/actions/` — `requestAccountDeletion` + `deleteAccount` (one-action-per-file, `requireUser`-gated, self-scoped).
- `features/settings/DangerZone.tsx` + public `/account/deleted` goodbye page.

## Security review (whole-branch, opus)
Ready to merge — no Critical/Important. Every entry point is `requireUser`-scoped; `db.delete(user)` is always `where(user.id == caller.id)`; all 7 cascade FKs confirmed in the live schema + migration; no plaintext code stored/logged; export returns a generic 500 without leaking internals.

## Tests
421 passing (84 files), type-check + `eslint .` clean. New unit tests cover the repository, code schema, purge orchestration (Garage + local + DB), the service state machine (throttle/expiry/attempt-cap/happy path), and the backup-zip round-trip.

> Note: manual dev-server smoke (click-through) was not run in the dev environment (no local Postgres). Worth a quick local pass before relying on it — and use a throwaway user, since it really deletes.

### DIFF
--- a/components/AppShell/publicPaths.ts
+++ b/components/AppShell/publicPaths.ts
@@ -5,7 +5,7 @@
 // place. A future addition (e.g. /pricing) can be made here once, instead of
 // duplicating a magic route literal across the proxy and the shell where a
 // missed copy could silently strand a public page inside the app chrome.
-export const PUBLIC_PATHS = new Set(['/']);
+export const PUBLIC_PATHS = new Set(['/', '/account/deleted']);
 
 export const isPublicPath = (pathname: string): boolean =>
   PUBLIC_PATHS.has(pathname);
--- a/components/AppShell/publicPaths.ts
+++ b/components/AppShell/publicPaths.ts
@@ -1,11 +1,28 @@
 // Paths that are publicly reachable without a session AND render their own
-// full-bleed chrome (no sidebar, header, or app-only banners). Today that is
-// just the marketing landing at `/`. Kept in its own tested module — mirroring
-// authPaths.ts — so the "which path gets which chrome" decision lives in one
-// place. A future addition (e.g. /pricing) can be made here once, instead of
-// duplicating a magic route literal across the proxy and the shell where a
-// missed copy could silently strand a public page inside the app chrome.
+// full-bleed chrome (no sidebar, header, or app-only banners). Two routes today:
+// the marketing landing at `/` and the post-deletion goodbye page at
+// `/account/deleted`. Kept in its own tested module — mirroring authPaths.ts —
+// so the "which path gets which chrome" decision lives in one place. A future
+// addition (e.g. /pricing) can be made here once, instead of duplicating a
+// magic route literal across the proxy and the shell where a missed copy could
+// silently strand a public page inside the app chrome.
 export const PUBLIC_PATHS = new Set(['/', '/account/deleted']);
 
 export const isPublicPath = (pathname: string): boolean =>
   PUBLIC_PATHS.has(pathname);
+
+// The subset of public paths that should *also* bounce a signed-in visitor to
+// their dashboard. This is deliberately narrower than PUBLIC_PATHS: the
+// marketing landing has no use for an authenticated user, so the proxy sends
+// them to `/dashboard` on a cheap cookie check.
+//
+// `/account/deleted` is intentionally NOT here. The deletion flow can leave a
+// stale session cookie behind when `authClient.signOut()` fails (the exact case
+// commit 9352138 tolerates), and a presence-only cookie check would then bounce
+// the just-deleted user to `/dashboard` instead of the goodbye page —
+// nullifying that fix. The goodbye page must stay reachable even with a stale
+// cookie present, so it is public-for-everyone but never redirected.
+export const BOUNCE_SIGNED_IN_PATHS = new Set(['/']);
+
+export const bouncesSignedInToDashboard = (pathname: string): boolean =>
+  BOUNCE_SIGNED_IN_PATHS.has(pathname);
--- a/features/settings/DangerZone.tsx
+++ b/features/settings/DangerZone.tsx
@@ -31,6 +31,10 @@ const DangerZone = () => {
       }
       setPhase('code-sent');
       toast.success('Verification code sent to your email.');
+    } catch {
+      // The server logs the real cause (e.g. email transport down); surface a
+      // generic failure so the user knows the request did not go through.
+      toast.error('Could not send a verification code. Please try again.');
     } finally {
       setPending(false);
     }
@@ -52,6 +56,10 @@ const DangerZone = () => {
       }
       switch (res.reason) {
         case 'no-code':
+          setError('Request a new code.');
+          setPhase('idle');
+          setCode('');
+          break;
         case 'expired':
           setError('That code expired. Request a new one.');
           setPhase('idle');
@@ -70,6 +78,11 @@ const DangerZone = () => {
           );
           break;
       }
+    } catch {
+      // Deletion is irreversible and may partially complete server-side (e.g.
+      // remote-storage cleanup failing mid-purge). Never fail silently — the
+      // server logs the real cause; surface a generic error so the user knows.
+      setError('Something went wrong. Please try again.');
     } finally {
       setPending(false);
     }
--- a/features/settings/DangerZone.tsx
+++ b/features/settings/DangerZone.tsx
@@ -42,7 +42,11 @@ const DangerZone = () => {
     try {
       const res = await deleteAccountAction(code);
       if (res.ok) {
-        await authClient.signOut();
+        try {
+          await authClient.signOut();
+        } catch {
+          // account is already deleted server-side; the session is invalid regardless
+        }
         router.push('/account/deleted');
         return;
       }
--- a/features/settings/DangerZone.tsx
+++ b/features/settings/DangerZone.tsx
@@ -1,0 +1,166 @@
+'use client';
+
+import { useState } from 'react';
+import { toast } from 'sonner';
+import { deleteAccountAction, requestAccountDeletionAction } from './actions';
+import { Alert, AlertDescription, AlertTitle } from '@/components/ui/alert';
+import { Button, buttonVariants } from '@/components/ui/button';
+import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
+import { Input } from '@/components/ui/input';
+import { Label } from '@/components/ui/label';
+import { authClient } from '@/lib/auth-client';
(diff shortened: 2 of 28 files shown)


features/settings/DangerZone.tsx
```diff
@@ -0,0 +1,170 @@
...
+      toast.success('Verification code sent to your email.');
+    } finally {
+      setPending(false);
+    }
+  };
+
+  const confirmDelete = async () => {
+    setPending(true);
+    setError(null);
+    try {
```
naeemba, line 46: `confirmDelete` has no `catch`. If the server action rejects — `purgeUserData` throwing because `clearRemote` (Garage) fails mid-purge, or `requireUser`/network failing — the promise rejects, `finally` re-enables the button, but **no error is shown to the user**. For an irreversible, partially-completing operation that is the worst case to fail silently on: the user clicks "Permanently delete", Garage deletion fails, nothing visibly happens, and they have no idea whether data was wiped. Add a `catch` that surfaces a generic error via `setError`/`toast.error` (the server already logs the real cause). Same shape applies to `sendCode` below.

lib/account-deletion/service.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    if (!constantTimeEqual(hashCode(code), challenge.codeHash)) {
+      const attempts = await this.repo.incrementAttempts(userId);
+      if (attempts >= MAX_ATTEMPTS) {
+        await this.repo.delete(userId);
+        return { ok: false, reason: 'too-many-attempts' };
+      }
+      return {
+        ok: false,
```
naeemba, line 83: `issueCode` upserts the challenge (fresh `createdAt`) **before** `sendCode`. If `sendCode` throws — Postal transport down, `EMAIL_FROM` unset (`index.ts` throws here), transient SMTP error — the row is already persisted with a current `createdAt`, so the next request within 30s returns `throttled`. Net effect of a transient email failure: the user gets no email, gets no error (the action rejection is swallowed client-side, see DangerZone), and is locked out of re-requesting for 30s. Consider sending the email first (or rolling back the upsert on send failure) so a failed send does not arm the throttle.

features/settings/DangerZone.tsx
```diff
@@ -0,0 +1,170 @@
...
+        } catch {
+          // account is already deleted server-side; the session is invalid regardless
+        }
+        router.push('/account/deleted');
+        return;
+      }
+      switch (res.reason) {
+        case 'no-code':
+        case 'expired':
+          setError('That code expired. Request a new one.');
```
naeemba, line 64: nit: `no-code` and `expired` share the message "That code expired." `no-code` is reachable when the challenge was never issued or was already consumed (e.g. a double-click after a successful purge, or attempting verify before requesting). "That code expired" is slightly misleading there — a neutral "Request a new code." covers both reasons more accurately.

naeemba (commented):
Reviewed the account-deletion + backup-export feature. The structure follows the repo's Repository + Service + one-action-per-file convention cleanly: `AccountDeletionChallengeRepository` maps exactly one table, `purgeUserData` is the orchestration unit, the export route mirrors the sibling `app/api/*/export/route.ts` shape (`requireUser` first, `force-dynamic`, generic 500), and the verification state machine is well-covered by unit tests. Hashed-at-rest + constant-time compare + expiry/attempt-cap/throttle are all enforced server-side.

A couple of client-side robustness gaps around the destructive path are noted inline.

features/settings/DangerZone.tsx
naeemba: Fixed in b9d7fbb. Added catch handlers to both confirmDelete and sendCode that surface a generic error (setError for delete, toast.error for send) so a rejected server action is never silent. The server still logs the real cause.

lib/account-deletion/service.ts
naeemba: Fixed in f53b35b. issueCode now wraps sendCode in a try/catch and deletes the challenge row on send failure before re-throwing, so a transient email failure no longer arms the 30s resend throttle. Added a service test covering the rollback (challenge gone after failed send, immediate retry succeeds).

features/settings/DangerZone.tsx
naeemba: Fixed in b9d7fbb. Split the no-code case from expired: no-code now shows the neutral "Request a new code." while expired keeps "That code expired. Request a new one."

lib/account-deletion/index.ts
```diff
@@ -0,0 +1,38 @@
+import 'server-only';
+import { purgeUserData } from './purge';
+import { AccountDeletionChallengeRepository } from './repository';
+import { AccountDeletionService } from './service';
+import { APP_NAME } from '@/lib/app';
+import { db } from '@/lib/db';
+import { postalTransport } from '@/lib/email-transport';
+
+const sendCode = async (email: string, code: string): Promise<void> => {
+  const from = process.env.EMAIL_FROM;
```
naeemba: Reads `process.env.EMAIL_FROM` directly and re-implements an env check, bypassing the validated `env` object in `lib/env/index.ts` — which already declares `EMAIL_FROM` (with a default of `'auth@example.com'`). The sibling `lib/email-transport.ts` is the canonical reference: it pulls `env.POSTAL_API_URL` / `env.POSTAL_API_KEY` from `@/lib/env`. This module should do the same: `import { env } from '@/lib/env'` and use `env.EMAIL_FROM`.

Two concrete consequences of diverging:
- The `if (!from) throw` branch is effectively dead — `env.EMAIL_FROM` always has a value because the schema supplies a default — so the guard reads as live error-handling but can never fire once you go through `env`. If the intent is actually to fail loud when `EMAIL_FROM` is unset (rather than silently send from the `auth@example.com` placeholder the env default would use), that decision belongs in the env schema (drop the `.default(...)` / make it required), not re-litigated here.
- Reading `process.env` directly skips the startup validation the rest of the app relies on, so a misconfiguration surfaces at first-delete-request time instead of at boot.

naeemba: Fixed in bfff0f8. sendCode now imports { env } from @/lib/env and reads env.EMAIL_FROM, matching the canonical lib/email-transport.ts. Since env.EMAIL_FROM always resolves (schema default), the dead if (!from) throw guard is removed — env startup validation is now the single source of truth.

components/AppShell/publicPaths.ts
```diff
@@ -5,7 +5,7 @@
 // place. A future addition (e.g. /pricing) can be made here once, instead of
 // duplicating a magic route literal across the proxy and the shell where a
 // missed copy could silently strand a public page inside the app chrome.
-export const PUBLIC_PATHS = new Set(['/']);
+export const PUBLIC_PATHS = new Set(['/', '/account/deleted']);
```
naeemba, line 9: Adding `/account/deleted` to `PUBLIC_PATHS` collides with the proxy's signed-in redirect and silently defeats the "always redirect to the goodbye page even if signOut fails" fix (9352138).\n\n`proxy.ts` treats every `isPublicPath` route as: if a session cookie is present → redirect to `/dashboard`. And `getSessionCookie` (re-exported from `better-auth/cookies`) is a pure cookie-**presence** check — no DB validation.\n\nTrace the deletion flow in `DangerZone.confirmDelete`:\n1. `deleteAccountAction` succeeds → user row is gone server-side.\n2. `await authClient.signOut()` is wrapped in try/catch *specifically* to tolerate failure.\n3. `router.push('/account/deleted')`.\n\nIf `signOut()` throws (network blip, transport down — the exact case the catch was added for), the session cookie is **not cleared**. The push to `/account/deleted` hits the proxy, which sees `isPublicPath('/account/deleted') === true` **and** a stale session cookie → redirects to `/dashboard`. The deleted user never sees the goodbye page and instead lands on the dashboard with a cookie pointing at a now-deleted user (which then fails server-side auth and bounces again). The whole point of commit 9352138 is nullified on the one path it was meant to cover.\n\nTwo ways out, pick one:\n- Have the proxy treat `/account/deleted` as public-but-not-redirect-when-signed-in (i.e. it's public for *everyone*, including a stale-cookie holder), separate from the "marketing landing → bounce signed-in users to /dashboard" behavior that `/` wants. The current `PUBLIC_PATHS` set conflates "no chrome" with "bounce signed-in users", and the goodbye page wants the former but not the latter.\n- Or guarantee the cookie is cleared before navigating (e.g. hard-navigate via `window.location` only after a confirmed sign-out, or clear the cookie server-side as part of the purge), so the signed-in branch can never fire here.\n\nWorth a test asserting the goodbye page is reachable with a session cookie still present.

naeemba: Fixed in e777558. Split the two concerns that PUBLIC_PATHS was conflating: PUBLIC_PATHS still means "public, no app chrome" (used by both AppShell and the proxy), and a new narrower BOUNCE_SIGNED_IN_PATHS (just `/`) drives the proxy's signed-in -> /dashboard redirect. The proxy now only bounces signed-in visitors on bounce paths, so /account/deleted stays reachable even when a stale session cookie lingers after a failed signOut() — preserving commit 9352138. Added tests asserting the goodbye page is public but never bounced.

PR #27, opened 2026-06-25
Phase 7 hardening: rate limiting, CSP/security headers, per-user quota

naeemba:
## Phase 7 — Multi-user hardening: rate limiting, CSP/security headers, per-user quota

Closes three open Phase 7 items now that the app is genuinely multi-user (open registration + self-service account deletion). Spec: `docs/superpowers/specs/2026-06-25-phase7-hardening-design.md`; plan: `docs/superpowers/plans/2026-06-25-phase7-hardening.md`.

### 1. Rate limiting (`lib/rate-limit/`)
- In-memory fixed-window limiter behind a one-method store interface (swappable for Redis/Postgres later), keyed per-user (every guarded surface is authenticated).
- Applied to `POST /api/upload` (**HTTP 429 + `Retry-After`**) and all 11 authenticated mutating server actions (transactions, templates, saved-views, saved base-currency, account-deletion request/verify).
- Guards sit **after** `requireUser()` (unauth → still 401) and **before** side effects; each returns a value that's already a member of that action's existing result union — no UI/type changes.
- Out of scope (already covered): auth endpoints (better-auth built-in), account-deletion email-code throttle (existing DB throttle), price refresh (daily server-side cap), anonymous currency-cookie toggles.

### 2. CSP + security headers (`proxy.ts`, `lib/security/headers.ts`)
- Strict **nonce-based CSP** emitted from the Next.js 16 `proxy.ts`: `script-src` is nonce + `strict-dynamic` (no `unsafe-inline`); `style-src 'unsafe-inline'` is the only inline allowance (Recharts/Base-UI inline `style=`). Plus HSTS, `X-Frame-Options: DENY`/`frame-ancestors 'none'`, `X-Content-Type-Options: nosniff`, `Referrer-Policy`, `Permissions-Policy`.
- Per-request nonce threaded onto request headers (so Next nonces its bootstrap scripts) and the full set onto the response; existing better-auth routing preserved.
- **Auth pages fix (folded in):** `/sign-in`, `/sign-in/error`, `/sign-up` now pass through the proxy via an `isAuthPath` early-return, so the credential forms get clickjacking protection without a redirect loop.

### 3. Per-user quota (`lib/journal/quota.ts`)
- `JOURNAL_QUOTA_MB` (default 100) enforced centrally in `JournalService`: imports are rejected **before** the journal dir is wiped (`replaceFromZip` checks extracted bytes before the lock; `replaceFromSingleFile` before `resetLocalJournal`), and `addTransaction` checks before append — the snapshot/verify/rollback/push flow is untouched. Mapped to **HTTP 413** in both upload branches.

### Verification
- **438 tests pass** (88 files; +13 new), type-check + lint clean.
- New unit tests: rate-limit store, `buildSecurityHeaders`, quota helper, quota enforcement.
- Final whole-branch review (Opus): **Ready to merge — 0 Critical, 0 Important.** Remaining findings cosmetic.

### ⚠️ Before merge — manual browser smoke (needs a configured env)
The strict nonce CSP couldn't be verified in-browser from the CI/worktree (no `DATABASE_URL`/Postgres there). Please run `pnpm dev` and confirm no CSP violations break Recharts charts, the Cmd+K palette / dialogs (Base-UI), or sonner toasts.

### DIFF
--- a/lib/journal/quota.test.ts
+++ b/lib/journal/quota.test.ts
@@ -2,6 +2,7 @@ import { promises as fs } from 'fs';
 import os from 'os';
 import path from 'path';
 import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { PRICE_DB_NAME } from './layout';
 import { getJournalDirSize, journalQuotaBytes } from './quota';
 
 let dataDir: string;
@@ -28,6 +29,14 @@ describe('journalQuotaBytes', () => {
     process.env.JOURNAL_QUOTA_MB = '5';
     expect(journalQuotaBytes()).toBe(5 * 1024 * 1024);
   });
+
+  it.each(['abc', '0', '-5'])(
+    'falls back to 100 MB (fail closed) for invalid value %j',
+    (value) => {
+      process.env.JOURNAL_QUOTA_MB = value;
+      expect(journalQuotaBytes()).toBe(100 * 1024 * 1024);
+    }
+  );
 });
 
 describe('getJournalDirSize', () => {
@@ -41,5 +50,13 @@ describe('getJournalDirSize', () => {
     await fs.writeFile(path.join(dir, 'main.ledger'), 'a'.repeat(100));
     await fs.writeFile(path.join(dir, 'sub', 'inc.ledger'), 'b'.repeat(50));
     expect(await getJournalDirSize(userId)).toBe(150);
+  });
+
+  it('excludes the auto-managed price DB from the total', async () => {
+    const dir = path.join(dataDir, 'journals', userId);
+    await fs.mkdir(dir, { recursive: true });
+    await fs.writeFile(path.join(dir, 'main.ledger'), 'a'.repeat(100));
+    await fs.writeFile(path.join(dir, PRICE_DB_NAME), 'p'.repeat(500));
+    expect(await getJournalDirSize(userId)).toBe(100);
   });
 });
--- a/lib/journal/quota.test.ts
+++ b/lib/journal/quota.test.ts
@@ -1,0 +1,44 @@
+import { promises as fs } from 'fs';
+import os from 'os';
+import path from 'path';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { getJournalDirSize, journalQuotaBytes } from './quota';
+
+let dataDir: string;
+const userId = 'user-quota';
+
+beforeEach(async () => {
+  dataDir = await fs.mkdtemp(path.join(os.tmpdir(), 'quota-'));
+  process.env.DATA_DIR = dataDir;
+});
+
+afterEach(async () => {
+  await fs.rm(dataDir, { recursive: true, force: true });
+  delete process.env.JOURNAL_QUOTA_MB;
+});
+
+describe('journalQuotaBytes', () => {
+  it('defaults to 100 MB', () => {
+    delete process.env.JOURNAL_QUOTA_MB;
+    expect(journalQuotaBytes()).toBe(100 * 1024 * 1024);
+  });
+
+  it('reads JOURNAL_QUOTA_MB from the environment', () => {
+    process.env.JOURNAL_QUOTA_MB = '5';
+    expect(journalQuotaBytes()).toBe(5 * 1024 * 1024);
+  });
+});
+
+describe('getJournalDirSize', () => {
+  it('returns 0 when the dir does not exist', async () => {
+    expect(await getJournalDirSize(userId)).toBe(0);
+  });
+
+  it('sums nested file sizes', async () => {
+    const dir = path.join(dataDir, 'journals', userId);
+    await fs.mkdir(path.join(dir, 'sub'), { recursive: true });
+    await fs.writeFile(path.join(dir, 'main.ledger'), 'a'.repeat(100));
+    await fs.writeFile(path.join(dir, 'sub', 'inc.ledger'), 'b'.repeat(50));
+    expect(await getJournalDirSize(userId)).toBe(150);
+  });
+});
--- a/lib/journal/quota.test.ts
+++ b/lib/journal/quota.test.ts
@@ -14,6 +14,7 @@ beforeEach(async () => {
 
 afterEach(async () => {
   await fs.rm(dataDir, { recursive: true, force: true });
+  delete process.env.DATA_DIR;
   delete process.env.JOURNAL_QUOTA_MB;
 });
 
--- a/lib/journal/quota.ts
+++ b/lib/journal/quota.ts
@@ -1,23 +1,40 @@
 import { promises as fs } from 'fs';
 import path from 'path';
 import 'server-only';
-import { getJournalDir } from './layout';
+import { PRICE_DB_NAME, getJournalDir } from './layout';
 import { listLocalRelPaths } from '@/lib/storage/manifest';
+
+const DEFAULT_QUOTA_MB = 100;
 
 /**
  * Per-user cumulative journal-dir cap in bytes. Read lazily from process.env so
  * tests can override it, mirroring DATA_DIR in layout.ts. The Zod-validated env
  * in @/lib/env is the production source of truth.
+ *
+ * Falls back to DEFAULT_QUOTA_MB when the runtime value is missing or
+ * non-numeric so a bad env var defaults *closed* (still enforcing a cap) rather
+ * than silently disabling the quota (Number('x') → NaN → every check false).
  */
-export const journalQuotaBytes = (): number =>
-  Number(process.env.JOURNAL_QUOTA_MB ?? 100) * 1024 * 1024;
+export const journalQuotaBytes = (): number => {
+  const mb = Number(process.env.JOURNAL_QUOTA_MB);
+  return (Number.isFinite(mb) && mb > 0 ? mb : DEFAULT_QUOTA_MB) * 1024 * 1024;
+};
 
-/** Total bytes of all files under the user's journal dir (0 if absent). */
+/**
+ * Total bytes of user-authored files under the journal dir (0 if absent).
+ *
+ * Excludes the auto-managed price DB (`price-db.ledger`): it is system-generated
+ * price data from refreshPrices, not user content. Counting it would let a large
+ * price DB block addTransaction even when the user's own content is under quota,
+ * while imports (which don't count it) would still succeed — an inconsistency the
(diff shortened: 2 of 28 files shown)


lib/journal/service.ts
```diff
@@ -274,6 +292,19 @@ export class JournalService {
       }
     }
 
+    const extractedBytes = entries.reduce(
+      (sum, entry) => sum + entry.getData().length,
```
naeemba: `replaceFromZip` decompresses every entry **twice**: once here in the quota `reduce` (`entry.getData()`), then again inside the lock when writing (`await fs.writeFile(target, entry.getData())` ~line 320). adm-zip does not cache decompressed buffers, so this doubles CPU/allocation on every archive import.

More importantly, this first pass fully decompresses the *entire* archive into memory before the quota can reject it. A malicious 25 MB zip (the upload cap) can decompress to far more than 100 MB here — the `extractedBytes` sum walks every entry's full uncompressed payload regardless of how far over quota it goes, so the quota check meant to protect against oversized imports is itself a zip-bomb amplification path.

Suggest decompressing each entry once (e.g. build `const datas = entries.map((e) => ({ entry: e, data: e.getData() }))` and reuse), and short-circuiting the size accumulation as soon as the running total exceeds `journalQuotaBytes()` so you stop decompressing past the limit.

lib/journal/service.ts
```diff
@@ -152,6 +153,15 @@ export class JournalService {
       const snapshot = await this.repo.readFile(mainPath);
       // Leading "\n\n" guards against imported files that lack a trailing newline.
       const block = `\n\n${formatTransaction(draft)}\n`;
+      const projected =
+        (await getJournalDirSize(userId)) + Buffer.byteLength(block);
```
naeemba, line 157: The two quota enforcement paths disagree on what counts toward the cap. Imports (`replaceFromSingleFile`/`replaceFromZip`) check only the size of the **new payload**. But `addTransaction` checks `getJournalDirSize(userId)`, which sums *every* file under the journal dir via `listLocalRelPaths` — including the auto-managed `price-db.ledger` (`PRICE_DB_NAME` from layout.ts), which is system-generated price data, not user content.

Consequence: a user whose price DB has grown large from `refreshPrices` can be blocked from adding a transaction (413 / "would exceed your N MB journal limit") even though their actual journal content is well under quota — and re-importing the same content would succeed because imports don't count the price DB. The enforcement is inconsistent and the failure is confusing (the user can't shrink the price DB).

Consider excluding `price-db.ledger` from `getJournalDirSize` (or from the quota math) so both paths measure the same thing: user-authored journal bytes.

lib/journal/quota.ts
```diff
@@ -0,0 +1,28 @@
...
+import { getJournalDir } from './layout';
+import { listLocalRelPaths } from '@/lib/storage/manifest';
+
+/**
+ * Per-user cumulative journal-dir cap in bytes. Read lazily from process.env so
+ * tests can override it, mirroring DATA_DIR in layout.ts. The Zod-validated env
+ * in @/lib/env is the production source of truth.
+ */
+export const journalQuotaBytes = (): number =>
+  Number(process.env.JOURNAL_QUOTA_MB ?? 100) * 1024 * 1024;
```
naeemba: `journalQuotaBytes()` reads `process.env.JOURNAL_QUOTA_MB` directly and runs it through `Number(...)`. If the env var is ever a non-numeric string, `Number('x')` is `NaN`, and every quota comparison (`bytes > journalQuotaBytes()`) evaluates to `false` — i.e. the quota **fails open** and stops enforcing entirely, silently.

In production this is guarded by the Zod `JOURNAL_QUOTA_MB` validation in `lib/env/index.ts`, so it's low-risk. But this helper deliberately bypasses that validated `env` (reading `process.env` lazily "so tests can override"), so the validated value never actually gates this path. Consider clamping/falling back: `const mb = Number(process.env.JOURNAL_QUOTA_MB); return (Number.isFinite(mb) && mb > 0 ? mb : 100) * 1024 * 1024;` so a bad runtime value defaults closed rather than disabling the cap.

proxy.ts
```diff
@@ -38,20 +72,21 @@ export function proxy(req: NextRequest) {
     );
     return NextResponse.redirect(target);
```
naeemba: nit: the two `NextResponse.redirect(...)` branches in `proxy()` (the bounce-to-dashboard at line 60 and the no-session redirect at line 73) return the redirect **without** the security headers, since only `withSecurityHeaders()` attaches them. The CSP/HSTS/X-Frame-Options set is therefore absent on 3xx responses. Impact is low (a redirect has no rendered body to attack and the browser follows it to a header-bearing page), but if the goal is a uniform header baseline you could thread the redirect through the same helper (e.g. set headers on the redirect response before returning).

lib/security/headers.ts
```diff
@@ -0,0 +1,31 @@
...
+    "object-src 'none'",
+  ].join('; ');
+
+  return {
+    'Content-Security-Policy': csp,
+    'Strict-Transport-Security': 'max-age=63072000; includeSubDomains',
+    'X-Content-Type-Options': 'nosniff',
+    'X-Frame-Options': 'DENY',
+    'Referrer-Policy': 'strict-origin-when-cross-origin',
+    'Permissions-Policy': 'camera=(), microphone=(), geolocation=()',
```
naeemba: nit: spec divergence — the design doc's `Permissions-Policy` table lists `camera=(), microphone=(), geolocation=(), interest-cohort=()`, but the implemented header drops `interest-cohort=()` (the opt-out of FLoC topic sampling). Either restore it to match the spec or update the spec; harmless either way, just flagging the mismatch.

lib/journal/service.ts
naeemba: Fixed in ebe62e3. `replaceFromZip` now decompresses each entry exactly once (buffers stored in `decompressed[]` and reused for the write), and the quota accumulation short-circuits the moment the running total crosses `journalQuotaBytes()` so an oversized/malicious archive stops decompressing past the limit.

lib/journal/service.ts
naeemba: Fixed in 07fd52d. `getJournalDirSize` now skips `price-db.ledger`, so both enforcement paths measure the same thing (user-authored bytes). A large auto-generated price DB can no longer block `addTransaction` while imports would have succeeded.

lib/journal/quota.ts
naeemba: Fixed in 07fd52d. `journalQuotaBytes()` now uses `Number.isFinite(mb) && mb > 0 ? mb : DEFAULT_QUOTA_MB`, so a missing/non-numeric runtime value defaults closed at 100 MB instead of going to NaN and silently disabling the cap.

proxy.ts
naeemba: Fixed in 4e01008. Both redirect branches now route through `redirectWithSecurityHeaders()`, which attaches the full security-header baseline (CSP/HSTS/X-Frame-Options/etc.) to the 3xx response for a uniform header set.

lib/security/headers.ts
naeemba: Fixed in 4e01008. Restored `interest-cohort=()` to `Permissions-Policy` to match the design spec.

lib/journal/quota.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  it('defaults to 100 MB', () => {
+    delete process.env.JOURNAL_QUOTA_MB;
+    expect(journalQuotaBytes()).toBe(100 * 1024 * 1024);
+  });
+
+  it('reads JOURNAL_QUOTA_MB from the environment', () => {
+    process.env.JOURNAL_QUOTA_MB = '5';
+    expect(journalQuotaBytes()).toBe(5 * 1024 * 1024);
+  });
+});
```
naeemba, line 40: The two regressions fixed in `07fd52d` have no test guarding them, so a future refactor can silently reintroduce either:

1. **fail-closed env handling** — `journalQuotaBytes()` is the security-relevant change (non-numeric/zero/negative `JOURNAL_QUOTA_MB` must fall back to 100 MB rather than `NaN`-disabling the cap). `quota.test.ts` only covers the missing-and-valid cases; add cases for `JOURNAL_QUOTA_MB = 'abc'`, `'0'`, and `'-5'` all returning `100 * 1024 * 1024`.\n\n2. **price-db exclusion** — `getJournalDirSize` now skips `price-db.ledger`, but the `sums nested file sizes` test never writes one. Add a case that writes `price-db.ledger` alongside `main.ledger` and asserts only the user file is counted — that is the exact behavior the fix relies on for the `addTransaction` path.

lib/journal/service.ts
```diff
@@ -152,6 +153,15 @@ export class JournalService {
...
+      if (projected > journalQuotaBytes()) {
+        return {
+          ok: false,
+          fieldErrors: {},
+          formError: `This transaction would exceed your ${process.env.JOURNAL_QUOTA_MB ?? 100} MB journal limit.`,
```
naeemba: nit: after the `07fd52d` fail-closed fix, the cap enforced by `journalQuotaBytes()` and the number shown to the user can disagree. The check uses the clamped value (bad env → 100 MB), but this message prints the raw `process.env.JOURNAL_QUOTA_MB ?? 100`, so a non-numeric/zero/negative env value would render e.g. "would exceed your abc MB journal limit" while the real cap is 100. Same literal is duplicated in `app/api/upload/route.ts` (two branches). Consider sourcing the displayed number from `journalQuotaBytes() / 1024 / 1024` so message and enforcement can't drift. Low risk in prod since Zod validates the env, hence nit.

naeemba: Fixed in 930becf. Added `journalQuotaMb()` (the clamped MB value) and sourced all three user-facing messages from it — the `addTransaction` formError and both `app/api/upload/route.ts` branches now print the same number that `journalQuotaBytes()` enforces, so display and enforcement can no longer drift on a bad env value.

lib/journal/quota.test.ts
naeemba: Fixed in 80d4cd9. Added the missing guards in `quota.test.ts`: an `it.each` over `'abc'`, `'0'`, `'-5'` asserting `journalQuotaBytes()` falls back to 100 MB (fail closed), and a `getJournalDirSize` case that writes `price-db.ledger` alongside `main.ledger` and asserts only the user file is counted. All 8 quota tests green.

PR #28, opened 2026-06-25
Encrypted journals P1: crypto core + storage seam (zero-knowledge foundation)

naeemba:
First of four slices toward **zero-knowledge encrypted journals at rest** (Phase 7). This PR lands the server-side cryptographic foundation **with no UI** — it is a **verified no-op** until P2's onboarding wizard exists (no `userCrypto` row can be created yet, and with no session DEK the storage seam is pure passthrough).

## What's here

- **`lib/crypto/fileCrypto.ts`** — per-file AES-256-GCM envelope `[magic "LEJ1"(4)][version(1)][nonce(12)][ciphertext][tag(16)]`, with a per-file subkey `HKDF-SHA256(DEK, info=relPath)` and AAD = `magic ‖ version ‖ relPath`. Path is bound two independent ways (subkey + AAD), so cross-file swap/replay fails closed.
- **`lib/crypto/sessionKeys.ts`** — in-RAM `Map<userId, DEK>` + `LockedError`. DEK is 32-byte-validated and defensively copied; never persisted, never logged.
- **`db/schema/userCrypto.ts`** + migration `0002` + **`lib/crypto/userCryptoRepository.ts`** — stores only **opaque client-created wrap blobs** (passphrase/recovery wraps, salts, Argon2 params). FK `ON DELETE cascade` so account deletion wipes it. The server never unwraps these.
- **`lib/crypto/journalCipher.ts`** + edits to **`lib/storage/download.ts`/`save.ts`** — encrypt/decrypt at the Garage push/pull seam, **DEK-driven and DB-free**: encrypt iff a session DEK is present; decrypt iff the body carries the `LEJ1` magic (else passthrough). `runLedger`/`verify`/`repository` are untouched.
- **`app/api/crypto/unlock` + `/lock`** + `decodeDek` + an `UNLOCK` rate-limit policy — receive the raw DEK over TLS into the in-RAM map; drop it on lock.

## Design & threat model

Zero-knowledge **at rest** (not full E2E — the server must see plaintext to run `ledger`). Garage/backups/Postgres hold only ciphertext + opaque wraps; the DEK lives only in RAM for the session. Spec: `docs/superpowers/specs/2026-06-25-encrypted-journals-v2-design.md`. Plan: `docs/superpowers/plans/2026-06-25-encrypted-journals-p1.md`.

## Safety property (verified in review)

**Every** mutation *and* import path in `JournalService` calls `pull()` before any write, and `pull` decrypts — so an encryption-enabled but **locked** user hits `LockedError` before any plaintext can be written over ciphertext. There is no silent-downgrade path on any flow.

## Tests

Built TDD via subagent-driven development, 5 tasks each independently reviewed plus a whole-branch opus review (**ready to merge, 0 Critical / 0 Important**). Full suite: **95 files, 467 tests passing**. Pre-existing storage tests unchanged and green (the no-DEK path is passthrough).

## Deferred (non-blocking, scheduled for P2)

- Add a 32-byte guard in `fileCrypto` and freeze the exported `MAGIC` (currently guarded upstream by `setSessionDek`).
- Re-export `NewUserCrypto` from the schema barrel; mock rate-limit in the unlock route test.

## Follow-ups (next slices)

P2 onboarding wizard + per-session unlock screen + plaintext→ciphertext migration; P3 Settings→Security; fast-follow passkey-PRF unlock.


lib/storage/download.ts
```diff
@@ -73,8 +74,9 @@ export const pullToLocal = async (
       }
```
naeemba, line 74: Locked-read safety gap: the unchanged-ETag fast path `continue`s before `decryptFromDownload` runs, so it never throws `LockedError`. For an encryption-enabled but *locked* user whose local plaintext cache is still warm (they unlocked earlier in this process, then locked — the DEK is dropped from RAM but the plaintext file on `getJournalDir` is not shredded), a subsequent `pull` matches the remote ciphertext ETag, sees the local file exists, and serves the stale plaintext instead of failing closed.

The PR description states "an encryption-enabled but locked user hits LockedError before any plaintext can be written" — that holds only when the local cache is absent. It is not reachable in production yet (no `userCrypto` row can exist until the P2 wizard), and shred-on-lock is explicitly deferred to P2, so this is a forward-looking note rather than a blocker. Flagging so the P2 lock flow (a) shreds the local dir on `dropSessionDek` and/or (b) makes this skip path lock-aware, otherwise the "locked" guarantee is only true on a cold cache.

db/schema/index.ts
```diff
@@ -6,4 +6,5 @@ export { commodityPrice, type CommodityPrice } from './commodityPrice';
 export { priceFetchRun, type PriceFetchRun } from './priceFetchRun';
 export { savedView, type SavedView } from './savedView';
 export { template, type Template } from './template';
+export { userCrypto, type UserCrypto } from './userCrypto';
```
naeemba: nit: `NewUserCrypto` isn't re-exported from the `db/schema` barrel, so `userCryptoRepository.ts` reaches past it into `@/db/schema/userCrypto` for the insert type while pulling `UserCrypto` via the barrel would be the consistent path. The sibling `accountDeletionChallenge` only needs the select type so it isn't a perfect precedent, but for consistency add `type NewUserCrypto` here (the PR's own "Deferred" list already notes this). Cheap to land now.

lib/crypto/transport.ts
```diff
@@ -0,0 +1,13 @@
+const DEK_BYTES = 32;
```
naeemba: nit: `DEK_BYTES = 32` and the 32-byte length validation are now defined independently here and in `sessionKeys.ts` (and the magic length lives in `fileCrypto.ts`). Consider hoisting a single `DEK_BYTES` constant (e.g. into `sessionKeys.ts` or a small `crypto/constants.ts`) and importing it in both, so the DEK size can't drift between the transport decoder and the in-RAM store.

app/api/crypto/unlock/route.test.ts
```diff
@@ -0,0 +1,52 @@
...
+import {
+  __resetSessionKeysForTest,
+  hasSessionDek,
+} from '@/lib/crypto/sessionKeys';
+
+const existsMock = vi.fn();
+
+vi.mock('@/lib/auth/require-user', () => ({
+  requireUser: vi.fn(async () => ({ id: 'alice' })),
+}));
```
naeemba, line 14: The real `rateLimit`/`UNLOCK` limiter is exercised here unmocked, sharing the process-global in-memory `MemoryStore` singleton keyed `unlock:alice`. With `max: 10/min` this file's three POSTs stay under budget, but any other test in the suite that unlocks the same user id contributes to the same window, so this is order-dependent shared global state. The PR's own "Deferred" list flags mocking the limiter here — worth doing, and while you're in here add a test for the 429 branch (the rate-limit path in the route is currently untested).

lib/crypto/transport.ts
naeemba: Done in 2454010 — hoisted a single `DEK_BYTES` into `lib/crypto/constants.ts` and imported it in `transport.ts`, `sessionKeys.ts`, and `fileCrypto.ts` (the HKDF subkey length), so the size can no longer drift between the transport decoder, the in-RAM store, and the envelope derivation.

db/schema/index.ts
naeemba: Done in 55ac931 — added `type NewUserCrypto` to the `db/schema` barrel and switched `userCryptoRepository.ts` to import both types via the barrel instead of reaching into `@/db/schema/userCrypto`.

app/api/crypto/unlock/route.test.ts
naeemba: Done in 1c7f0a7 — mocked `@/lib/rate-limit` in the unlock route test so it no longer shares the process-global `MemoryStore` window with any other test that unlocks the same user id (removes the order-dependent shared global state), and added a test for the 429 branch asserting status 429 and a `Retry-After` header.

lib/storage/download.ts
naeemba: Not changing this in P1 — agreeing with your own framing that this is a forward-looking note, not a blocker (no `userCrypto` row can exist until the P2 wizard, so the locked-warm-cache path is unreachable in production today), and that shred-on-lock is explicitly a P2 deliverable.

I prototyped option (b) — making the skip path lock-aware by resolving encryption-enabled state once via `getUserCryptoRepository().exists(userId)` and failing closed with `LockedError` before serving the warm cache. It works, but it pulls a DB dependency into the storage seam (`pullToLocal`), which breaks the layering and fails the seam's pure unit tests (`seam-encryption.test.ts`, `download.test.ts` run with no DB and would now hit Postgres). The correct fix is option (a): shred the local journal dir on `dropSessionDek`, which keeps the seam DB-free and makes the "locked" guarantee hold on a warm cache too. Tracking that as part of the P2 lock flow rather than bolting a DB call onto the storage layer here.

PR #29, opened 2026-06-25
Encrypted journals P2: client crypto, gated wizard, unlock, migration

naeemba:
Builds on P1 (#28) to make **zero-knowledge encryption real and usable**. After this, a user is gated into a setup wizard that generates their key client-side, sets a passphrase + recovery code, and encrypts their existing journal; subsequent sessions require an unlock.

> ⚠️ **Live e2e/visual acceptance is pending** (see checklist) — this needs a real Postgres + Garage `pnpm dev` env, which the build worktree lacks. All static verification (495 tests, type-check, lint) is green and the whole-branch review is clean.

## What's here
- **Client crypto** (`features/crypto/lib/clientCrypto.ts`): 32-byte DEK generation, Argon2id passphrase KEK (`hash-wasm`), recovery-code HKDF KEK, AES-GCM wrap/unwrap (WebCrypto), grouped Base32 recovery codes. The passphrase/recovery code **never leave the browser**; only opaque wraps are uploaded; only the raw DEK is POSTed (over TLS) to P1's `/api/crypto/unlock`.
- **Setup** (`setupCrypto` action + `setupSchema`) stores only opaque wraps in `userCrypto`.
- **Migration** (`JournalService.enableEncryption`, under `withUserLock`): re-encrypts the existing plaintext journal, idempotent via the `LEJ1` magic, **provably round-trips** back to the original plaintext (test wipes local → pulls → asserts byte-identical).
- **The gate**: `proxy.ts` forwards `x-pathname`; `CryptoGate` (root-layout server component) redirects un-set-up → `/crypto/setup`, locked → `/crypto/unlock`; loop-free; `LockedError` at the data layer is the hard backstop.
- **UI (au-* design, matches sign-in)**: 4-step setup wizard (Why → passphrase → recovery code shown once → encrypting), per-session unlock screen (passphrase + recovery), Lock control in the header, drop-DEK-on-sign-out.

## Notable fixes caught in review
- **Double-encryption bug (data-loss-level)**: the migration wrote ciphertext then `push`ed, and P1's `encryptForUpload` seam had no `isCiphertext` guard → double-wrapped, unrecoverable journals. Fixed with a guard (`lib/crypto/journalCipher.ts`), proven by a wipe+pull round-trip test (also hardens P1's seam generally).
- **CSP blocked WebAssembly**: P1's strict nonce CSP had no `'wasm-unsafe-eval'`, which would have made `hash-wasm` Argon2id throw in-browser — the wizard wouldn't work at all. Added the narrow keyword (permits WASM compile only, **not** JS `eval`; strict-dynamic/nonce JS protection preserved).

## Verification
Built TDD via subagent-driven development — 11 tasks, each implemented + independently reviewed, plus a whole-branch opus review: **Ready to merge, 0 Critical / 0 Important.** Full suite **495 tests passing**, type-check + lint clean.

## Your e2e acceptance checklist (before merge)
- [ ] Fresh un-set-up user → gated to `/crypto/setup`; wizard walks Why → passphrase → recovery (shown once, copy + download, "I've saved it" gates Continue) → encrypting → lands on dashboard.
- [ ] **CSP/WASM**: no console CSP error during Argon2id (the fix above should cover it — confirm).
- [ ] Journal reports still render after setup; the object at rest in Garage carries the `LEJ1` magic (ciphertext).
- [ ] Lock → routed to `/crypto/unlock`; **passphrase** unlocks; **recovery code** also unlocks (fresh session each).
- [ ] Sign out → sign in → re-unlock required.
- [ ] Wizard visual quality matches sign-in / unlock.

## Non-blocking follow-ups (P3 candidates)
- Partial-setup recoverability: if `setupCrypto` succeeds but `postDek` fails, "Retry" gets stuck (gate no-ops on `/crypto/*`). Recommend the gate redirect already-set-up users away from `/crypto/setup`.
- P1 carryover: 32-byte `dek` guard in `fileCrypto`; freeze the exported `MAGIC`.

### DIFF
--- a/features/crypto/SetupWizard.tsx
+++ b/features/crypto/SetupWizard.tsx
@@ -488,7 +488,9 @@ function RecoveryStep({ code, onNext }: { code: string; onNext: () => void }) {
     const a = document.createElement('a');
     a.href = url;
     a.download = 'ledger-recovery-code.txt';
+    document.body.appendChild(a);
     a.click();
+    document.body.removeChild(a);
     URL.revokeObjectURL(url);
   }
 
@@ -664,12 +666,20 @@ export function SetupWizard() {
   const [step, setStep] = useState<Step>('why');
   const [recoveryCode, setRecoveryCode] = useState<string | null>(null);
   const [fatalError, setFatalError] = useState<string | null>(null);
+  const [fatalRetryStep, setFatalRetryStep] = useState<Step>('passphrase');
 
   // Advance from passphrase → recovery: runs the full orchestration.
   async function handlePassphraseNext(passphrase: string) {
-    const code = await runSetup(passphrase);
-    setRecoveryCode(code);
-    setStep('recovery');
+    try {
+      const code = await runSetup(passphrase);
+      setRecoveryCode(code);
+      setStep('recovery');
+    } catch (err) {
+      setFatalRetryStep('passphrase');
+      setFatalError(
+        err instanceof Error ? err.message : CRYPTO_COPY.errors.setupFailed
+      );
+    }
   }
 
   // Advance from recovery → encrypting: kick off finalization.
@@ -678,12 +688,14 @@ export function SetupWizard() {
     try {
       const result = await finalizeEncryption();
       if (!result.ok) {
+        setFatalRetryStep('recovery');
         setFatalError(result.error ?? CRYPTO_COPY.errors.setupFailed);
         return;
       }
       // Hard navigate so the session gate sees `ready`.
       window.location.assign('/dashboard');
     } catch (err) {
+      setFatalRetryStep('recovery');
       setFatalError(
         err instanceof Error ? err.message : CRYPTO_COPY.errors.generic
       );
@@ -743,7 +755,7 @@ export function SetupWizard() {
                     className="au-btn au-btn--ghost"
                     onClick={() => {
                       setFatalError(null);
-                      setStep('recovery');
+                      setStep(fatalRetryStep);
                     }}
                   >
                     Retry
--- a/features/crypto/SetupWizard.tsx
+++ b/features/crypto/SetupWizard.tsx
@@ -40,10 +40,14 @@ const mono = JetBrains_Mono({
 
 const ARGON = { m: 65536, t: 3, p: 1 } as const;
 
-// Fixed setup orchestration — runs exactly once on advancing from step 2 → 3.
-// Generates DEK client-side, derives both wrapping keys, posts to server,
-// unlocks the session, and returns the one-time recovery code.
-async function runSetup(passphrase: string): Promise<string> {
+// Fixed setup orchestration — runs on advancing from step 2 → 3.
+// Generates DEK client-side, derives both wrapping keys, and persists the
+// wrapped keys server-side. Returns the raw DEK and the one-time recovery code
+// so the caller can surface the code immediately and unlock the session
+// separately — a failure to unlock must not discard the recovery secret.
+async function runSetup(
+  passphrase: string
+): Promise<{ dek: Uint8Array; code: string }> {
   const dek = generateDek();
   const salt = crypto.getRandomValues(new Uint8Array(16));
   const { code, bytes } = generateRecoveryCode();
@@ -59,8 +63,7 @@ async function runSetup(passphrase: string): Promise<string> {
     wrapRecovery,
   });
   if (!res.ok) throw new Error(res.error);
-  await postDek(dek); // unlock this session so migration can run
-  return code;
+  return { dek, code };
 }
 
 type Step = 'why' | 'passphrase' | 'recovery' | 'encrypting';
@@ -667,11 +670,18 @@ export function SetupWizard() {
   const [recoveryCode, setRecoveryCode] = useState<string | null>(null);
   const [fatalError, setFatalError] = useState<string | null>(null);
   const [fatalRetryStep, setFatalRetryStep] = useState<Step>('passphrase');
+  // Held in memory so a transient unlock failure can be retried at finalize
+  // time without re-running setupCrypto (which would now reject with
+  // "already set up" and orphan the displayed recovery code).
+  const dekRef = useRef<Uint8Array | null>(null);
 
-  // Advance from passphrase → recovery: runs the full orchestration.
+  // Advance from passphrase → recovery: persists the wrapped keys, then
+  // surfaces the recovery code. The session unlock (postDek) is deferred to
+  // finalize so a network blip here can never discard the one-time code.
   async function handlePassphraseNext(passphrase: string) {
     try {
-      const code = await runSetup(passphrase);
+      const { dek, code } = await runSetup(passphrase);
+      dekRef.current = dek;
       setRecoveryCode(code);
       setStep('recovery');
     } catch (err) {
@@ -682,10 +692,19 @@ export function SetupWizard() {
     }
   }
 
-  // Advance from recovery → encrypting: kick off finalization.
+  // Advance from recovery → encrypting: unlock the session, then finalize.
+  // Both steps are idempotent and safe to re-run on "Retry".
   async function handleRecoveryNext() {
+    const dek = dekRef.current;
+    if (!dek) {
+      // Lost the in-memory DEK (e.g. a reload landed back here) — the row
+      // already exists, so route through the normal unlock flow instead.
+      window.location.assign('/crypto/unlock');
+      return;
+    }
     setStep('encrypting');
     try {
+      await postDek(dek); // unlock this session so migration can run
       const result = await finalizeEncryption();
       if (!result.ok) {
         setFatalRetryStep('recovery');
--- a/features/crypto/SetupWizard.tsx
+++ b/features/crypto/SetupWizard.tsx
@@ -1,0 +1,767 @@
+'use client';
+
+import {
+  KeyRound,
+  Copy,
+  Download,
+  CheckCircle2,
+  Loader2,
+  ShieldCheck,
+} from 'lucide-react';
+import { type FormEvent, useState, useRef } from 'react';
+import '@/features/auth/auth.css';
+import { finalizeEncryption } from '@/features/crypto/actions/finalizeEncryption';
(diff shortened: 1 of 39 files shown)


features/crypto/SetupWizard.tsx
```diff
@@ -0,0 +1,779 @@
...
+  );
+  const wrapRecovery = await wrapDek(dek, await recoveryHkdfKey(bytes));
+  const res = await setupCrypto({
+    wrapPassphrase,
+    passSalt: toBase64(salt),
+    argonParams: ARGON,
+    wrapRecovery,
+  });
+  if (!res.ok) throw new Error(res.error);
+  await postDek(dek); // unlock this session so migration can run
```
naeemba: **Stuck setup + permanent recovery-code loss on a partial failure.** If `setupCrypto` (the DB write) succeeds but `postDek` (line 62) throws — a transient network blip is enough — the `userCrypto` row already exists, so `cryptoStatus` reports `locked`. The wizard surfaces the error and "Retry" re-enters `runSetup`, which calls `setupCrypto` again; that now returns `Encryption is already set up` (setupCrypto.ts L16) and the user is wedged on `/crypto/setup` (the gate no-ops on crypto paths). The user never reached the recovery step, so the recovery code generated on L49 is gone forever — they can only ever recover via the passphrase they happened to set. The PR notes a variant of this as a non-blocking follow-up, but the silent, permanent loss of the recovery secret makes it worth addressing before merge: either have the gate redirect already-set-up users off `/crypto/setup` to `/crypto/unlock`, or make `runSetup` idempotent/resumable (detect the existing row and continue to the recovery step rather than re-creating).

features/crypto/SetupWizard.tsx
```diff
@@ -0,0 +1,779 @@
...
+        err instanceof Error ? err.message : CRYPTO_COPY.errors.setupFailed
+      );
+    }
+  }
+
+  // Advance from recovery → encrypting: kick off finalization.
+  async function handleRecoveryNext() {
+    setStep('encrypting');
+    try {
+      const result = await finalizeEncryption();
```
naeemba, line 708: **Journal can stay plaintext-at-rest while the account is marked encryption-ready.** `enableEncryption` (the bulk migration of existing journal files) is invoked only from `finalizeEncryption`, which runs only from the wizard's `handleRecoveryNext`. If the browser is closed (or `finalizeEncryption` fails) after `postDek` on line 62 succeeds but before finalization, the `userCrypto` row exists and a future session can unlock to `ready` — but the existing journal in Garage was never re-encrypted. The user now believes their journal is encrypted while it sits in plaintext at rest until the next mutating push happens to re-encrypt it via the `encryptForUpload` seam. Consider re-running the idempotent `enableEncryption` on unlock/first-ready (it short-circuits via the `LEJ1` magic when already ciphertext), or persisting a `migrated` flag and reconciling it in the gate.

features/crypto/lib/unlockFlow.ts
```diff
@@ -0,0 +1,63 @@
+import {
+  derivePassphraseKek,
+  fromBase64,
+  parseRecoveryCode,
+  recoveryHkdfKey,
+  toBase64,
+  unwrapDek,
+} from './clientCrypto';
+
+type Material = {
```
naeemba: nit: This local `Material` type re-spells the response shape of `app/api/crypto/material/route.ts` (passSalt / argonParams / wrapPassphrase / wrapRecovery), which is itself a projection of `SetupCryptoInput` (lib/crypto/setupSchema.ts) minus `userId`. The `argonParams` shape is now hand-written in three places (here, setupSchema, and the `ARGON` const in SetupWizard). Consider deriving from the schema (e.g. `Pick<SetupCryptoInput, 'passSalt' | 'argonParams' | 'wrapPassphrase' | 'wrapRecovery'>` or a shared exported type for the GET payload) so the material route and its client consumer can't drift.

features/crypto/SetupWizard.tsx
naeemba: Fixed in 558fd04. Two-part fix matching both your suggestions:

1. **Gate / page redirect**: \`app/crypto/setup/page.tsx\` now checks \`cryptoStatus\` server-side and redirects already-set-up users to \`/crypto/unlock\`, so a reload onto \`/crypto/setup\` after the row exists no longer re-enters the wizard and wedges on the "already set up" rejection.
2. **In-session resumability**: \`runSetup\` no longer calls \`postDek\` before returning — it persists the wrapped keys and returns \`{ dek, code }\`. The recovery code is committed to state (and the DEK held in a ref) the instant the row is written, so a transient \`postDek\` blip can no longer discard the one-time code. \`postDek\` is deferred to the finalize step where it is retried idempotently; if the in-memory DEK is gone (reload), finalize routes to \`/crypto/unlock\` instead.

features/crypto/SetupWizard.tsx
naeemba: Fixed in 13ffa96. Took the "re-run idempotent enableEncryption on unlock" route you suggested: after a successful unlock in \`UnlockScreen\`, it now calls \`finalizeEncryption()\` (which runs \`enableEncryption\`) before navigating. That short-circuits already-ciphertext files via the LEJ1 magic, so for the common already-encrypted case it is a cheap no-op, but it brings any journal that was left plaintext-at-rest (setup abandoned after the row was written but before migration) up to ciphertext without waiting for the next mutating push. The call is best-effort (\`.catch(() => {})\`) so it never blocks entry — the DEK is already in-session and the storage seam re-encrypts on push regardless.

features/crypto/lib/unlockFlow.ts
naeemba: Fixed in 5583262. Added an exported \`CryptoMaterial = Pick<SetupCryptoInput, "passSalt" | "argonParams" | "wrapPassphrase" | "wrapRecovery">\` to \`lib/crypto/setupSchema.ts\`. \`unlockFlow.ts\` now imports that type instead of re-spelling \`Material\`, and the GET \`/api/crypto/material\` route types its response object as \`CryptoMaterial\` so the producer and consumer are pinned to the schema and cannot drift. (The \`ARGON\` const in SetupWizard is a value, not a type, so it stays.)

features/crypto/UnlockScreen.tsx
```diff
@@ -0,0 +1,301 @@
...
+        await unlockWithRecovery(value);
+      }
+      // Reconcile any journal that was left plaintext-at-rest if a prior setup
+      // was abandoned after the userCrypto row was written but before the bulk
+      // migration ran. enableEncryption is idempotent — it short-circuits on
+      // already-ciphertext files via the LEJ1 magic — so this is a cheap no-op
+      // for the common already-encrypted case. Best-effort: a failure here
+      // must not block entry, since the DEK is now in-session and the data
+      // layer re-encrypts on the next mutating push regardless.
+      await finalizeEncryption().catch(() => {});
```
naeemba, line 169: The reconcile fix is correct, but the comment calling this a "cheap no-op" understates its cost. `finalizeEncryption()` → `enableEncryption()` runs `withUserLock` + `pull` + a **full `push`**, and `pushFromLocal` (lib/storage/save.ts:54-58) unconditionally re-uploads *every* local file via `store.put` regardless of whether any byte changed — there is no content/etag short-circuit. So in the common already-encrypted case this still does a full re-download and full re-upload of the entire journal to Garage on **every single unlock**, not a no-op. The `isCiphertext` LEJ1 guard only avoids re-*encrypting*; it does not avoid re-*uploading*.

Since the only case that actually needs reconciling is "row written but bulk migration never ran" (a rare partial-setup), gating this on a persisted `migrated` flag on `userCrypto` (the alternative you noted in the original thread) would avoid the per-unlock full pull+push for the 99% already-migrated case. Worth doing before merge, or at minimum correct the "cheap no-op" comment so the cost is not misrepresented.

features/crypto/SetupWizard.tsx
naeemba: Verified and resolving. `runSetup` now returns `{ dek, code }` and no longer calls `postDek` before returning, so the recovery code is committed to state the instant the row is written; `postDek` is deferred to `handleRecoveryNext` and `dekRef` carries the DEK across retries. `app/crypto/setup/page.tsx` redirects any non-`unset` user to `/crypto/unlock`, so a reload after the row exists no longer wedges on the "already set up" rejection. Both failure paths from the original report are closed.

features/crypto/lib/unlockFlow.ts
naeemba: Verified and resolving. `CryptoMaterial = Pick<SetupCryptoInput, ...>` is now the single source of truth in `lib/crypto/setupSchema.ts`; both the GET `/api/crypto/material` route response and `unlockFlow.ts` import it, so producer and consumer cannot drift.

features/crypto/UnlockScreen.tsx
naeemba: Fixed in 5f187ae — took the persisted `migrated` flag route. Added a nullable `migratedAt` timestamp to `userCrypto` (migration `0003_thankful_tag.sql`). `enableEncryption` now stamps it via `repo.markMigrated` on success, and `finalizeEncryption` short-circuits to a single DB read (`repo.hasMigrated`) before any storage I/O when it is already set — so the 99% already-migrated case on every unlock no longer does a full pull+push. Only the rare partial-setup user (row written but bulk migration never ran) pays the one-time cost. The misleading "cheap no-op" comment in `UnlockScreen` is corrected to describe the single-read short-circuit. Added repo tests covering `hasMigrated`/`markMigrated` (incl. idempotency).

features/crypto/SetupWizard.tsx
naeemba: Follow-up in 5f187ae. The unlock-time reconcile that closed this thread (13ffa96) is now gated on a persisted `migratedAt` flag on `userCrypto`, so the common already-encrypted case is a single DB read with no per-unlock journal pull+push — while still bringing any plaintext-at-rest journal from an abandoned setup up to ciphertext the first time that user unlocks. Original plaintext-at-rest concern remains closed and the per-unlock cost it introduced is eliminated.

PR #30, opened 2026-06-25
feat(crypto): encrypted journals P3 — Settings → Security

naeemba:
## Encrypted Journals — P3: Settings → Security

Lets a user manage their encryption from **Settings → Security**, all zero-knowledge:

- **Change passphrase** — with a recovery-code "forgot passphrase" fallback.
- **Rotate recovery code** — new code shown once (copy + download + acknowledge).
- **Reset encryption** — destructive, guarded by an **emailed 6-digit code** (equal strength to account deletion). Wipes the encrypted journal + `userCrypto` row → status `unset` → gate routes to `/crypto/setup`. **Keeps the account.**

**Architecture:** all re-wrapping is client-side. The browser re-obtains the DEK by unwrapping with the *current* secret (exactly as unlock does), then wraps the **same DEK** under a new KEK and uploads only the new opaque wrap. The DEK never changes → journal data is never re-encrypted. The server only ever stores opaque wraps.

Plan: `docs/superpowers/plans/2026-06-25-encrypted-journals-p3.md`. Built via subagent-driven development (TDD per task).

### Commits
- `feat(crypto): userCrypto re-wrap + delete repository methods`
- `feat(crypto): client re-wrap orchestration (change passphrase, rotate recovery)`
- `feat(crypto): change-passphrase + rotate-recovery server actions`
- `feat(crypto): resetUserEncryption wipe service (keeps account)`
- `feat(crypto): emailed-code challenge for reset-encryption` (new `encryptionResetChallenge` table, migration `0004`)
- `feat(crypto): settings security section (change passphrase, rotate recovery, reset)`
- `fix(crypto): mark resetUserEncryption server-only`

### Quality gates
- Each task spec+quality reviewed; whole-branch **opus** review: **READY TO MERGE — 0 Critical / 0 Important**.
- Four binding invariants verified in-code end-to-end: **zero-knowledge**, **DEK-never-changes**, **proof-before-mutation** (wrong secret fails the unwrap with a friendly error, no server mutation), **non-replayable reset** (challenge row explicitly deleted since reset doesn't cascade-delete the user; challenge fully independent from the account-deletion challenge).
- Full suite **551 passing** (110 files), type-check + lint clean.
- Reset challenge: sha256-hashed code, constant-time compare, 10-min TTL, 5-attempt cap, 30s resend throttle, email-rollback on send failure. Migration `0004` creates only `encryptionResetChallenge` (FK cascade to user).

### ⚠️ Live e2e acceptance — REQUIRED before merge (worktree has no Postgres/Garage)
- [ ] **Change passphrase** (current path **and** recovery "forgot" path) → re-unlock with the **new** passphrase → existing journal **still decrypts** / reports render identically (proves DEK identity — only provable against a real DB + Garage).
- [ ] **Rotate recovery** → old code rejected, new code unlocks; new code revealed only after successful persist.
- [ ] **Reset encryption** → email delivers code → confirm → journal wiped, hard-nav to `/crypto/setup`, can re-set-up; Resend works; wrong/expired/too-many-attempts surfaced.
- [ ] **SecuritySection** visual rendering in the settings page (shadcn, matches Danger Zone).

### Deferred non-blocking fast-follows
- `rewrapFlow` tests: snapshot-assert production Argon params `{m:65536,t:3,p:1}` + cover the wrong-recovery-code error path.
- change/rotate action tests: assert `revalidatePath('/', 'layout')` is called.
- `confirmEncryptionResetAction` returns `reason:'too-many-attempts'` on an HTTP rate-limit hit (cosmetic message conflation; no security impact).
- `ChangePassphraseCard`: remove unused `formRef`; rotate button label is static in the recovery-authorizer path.

### Out of scope (next)
- Passkey-PRF unlock fast-follow — add/remove passkey would also live in this Security section once PRF lands.

### DIFF
--- a/features/crypto/lib/rewrapFlow.ts
+++ b/features/crypto/lib/rewrapFlow.ts
@@ -1,0 +1,80 @@
+import {
+  derivePassphraseKek,
+  generateRecoveryCode,
+  parseRecoveryCode,
+  recoveryHkdfKey,
+  toBase64,
+  unwrapDek,
+  wrapDek,
+  fromBase64,
+} from './clientCrypto';
+
+const ARGON = { m: 65536, t: 3, p: 1 } as const;
+
+type Material = {
+  passSalt: string;
+  argonParams: { m: number; t: number; p: number };
+  wrapPassphrase: string;
+  wrapRecovery: string;
+};
+
+export type Authorizer =
+  | { kind: 'passphrase'; passphrase: string }
+  | { kind: 'recovery'; code: string };
+
+const getMaterial = async (): Promise<Material> => {
+  const res = await fetch('/api/crypto/material');
+  if (!res.ok) throw new Error('Encryption is not set up.');
+  return res.json();
+};
+
+/** Re-obtain the DEK client-side by unwrapping with the current secret. */
+export const obtainDek = async (
+  authorizer: Authorizer
+): Promise<Uint8Array> => {
+  const m = await getMaterial();
+  if (authorizer.kind === 'passphrase') {
+    const kek = await derivePassphraseKek(
+      authorizer.passphrase,
+      fromBase64(m.passSalt),
+      m.argonParams
+    );
+    return unwrapDek(m.wrapPassphrase, kek).catch(() => {
+      throw new Error('Incorrect passphrase.');
+    });
+  }
+  const kek = await recoveryHkdfKey(parseRecoveryCode(authorizer.code));
+  return unwrapDek(m.wrapRecovery, kek).catch(() => {
+    throw new Error('Incorrect recovery code.');
+  });
+};
+
+export const changePassphrase = async (
+  authorizer: Authorizer,
+  newPassphrase: string
+): Promise<{
+  wrapPassphrase: string;
+  passSalt: string;
+  argonParams: { m: number; t: number; p: number };
+}> => {
+  const dek = await obtainDek(authorizer);
+  const salt = crypto.getRandomValues(new Uint8Array(16));
+  const wrapPassphrase = await wrapDek(
+    dek,
+    await derivePassphraseKek(newPassphrase, salt, ARGON)
+  );
+  return {
+    wrapPassphrase,
+    passSalt: toBase64(salt),
+    argonParams: { ...ARGON },
+  };
+};
+
+export const rotateRecovery = async (
+  authorizer: Authorizer
+): Promise<{ wrapRecovery: string; code: string }> => {
+  const dek = await obtainDek(authorizer);
+  const { code, bytes } = generateRecoveryCode();
+  const wrapRecovery = await wrapDek(dek, await recoveryHkdfKey(bytes));
+  return { wrapRecovery, code };
+};
--- a/features/crypto/lib/rewrapFlow.ts
+++ b/features/crypto/lib/rewrapFlow.ts
@@ -8,25 +8,13 @@ import {
   wrapDek,
   fromBase64,
 } from './clientCrypto';
+import { getMaterial } from './cryptoMaterial';
 
 const ARGON = { m: 65536, t: 3, p: 1 } as const;
-
-type Material = {
-  passSalt: string;
-  argonParams: { m: number; t: number; p: number };
-  wrapPassphrase: string;
-  wrapRecovery: string;
-};
 
 export type Authorizer =
   | { kind: 'passphrase'; passphrase: string }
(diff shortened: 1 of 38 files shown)


features/crypto/lib/rewrapFlow.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export type Authorizer =
+  | { kind: 'passphrase'; passphrase: string }
+  | { kind: 'recovery'; code: string };
+
+const getMaterial = async (): Promise<Material> => {
+  const res = await fetch('/api/crypto/material');
+  if (!res.ok) throw new Error('Encryption is not set up.');
+  return res.json();
+};
```
naeemba: Duplicated helper + duplicated type. This re-declares a private `Material` type (lines 14-19) and a private `getMaterial()` (lines 25-29) that are behaviorally identical to the ones already shipped in `features/crypto/lib/unlockFlow.ts` (lines 11-15). `unlockFlow` imports and reuses the exported `CryptoMaterial` type from `@/lib/crypto/setupSchema` (`Pick<…, 'passSalt' | 'argonParams' | 'wrapPassphrase' | 'wrapRecovery'>`) — the exact shape this local `Material` re-spells. The P3 plan itself says "Reuse P2 verbatim: `getMaterial` … from `features/crypto/lib/`", so this drifts from the intended design.

Suggest: export `getMaterial` from `unlockFlow.ts` (or extract both flows' `getMaterial` into a small shared `cryptoMaterial.ts`) and import `CryptoMaterial` instead of re-declaring `Material`. Keeps a single source of truth for the `/api/crypto/material` contract so the two unwrap paths can't drift.

features/settings/ChangePassphraseCard.tsx
```diff
@@ -0,0 +1,166 @@
...
+const MIN_LENGTH = 8;
+
+const ChangePassphraseCard = () => {
+  const [useForgot, setUseForgot] = useState(false);
+  const [currentSecret, setCurrentSecret] = useState('');
+  const [newPass, setNewPass] = useState('');
+  const [confirmPass, setConfirmPass] = useState('');
+  const [error, setError] = useState<string | null>(null);
+  const [pending, setPending] = useState(false);
+  const formRef = useRef<HTMLFormElement>(null);
```
naeemba: nit: `formRef` is declared and attached to the `<form>` (line 78) but never read. Drop the ref and the `useRef` import. (Already listed as a deferred fast-follow in the PR description.)

features/settings/RotateRecoveryCard.tsx
```diff
@@ -0,0 +1,252 @@
...
+    }
+  };
+
+  const handleCopy = async () => {
+    if (!newCode) return;
+    try {
+      await navigator.clipboard.writeText(newCode);
+      setCopied(true);
+      if (copyTimeout.current) clearTimeout(copyTimeout.current);
+      copyTimeout.current = setTimeout(() => setCopied(false), 2500);
```
naeemba, line 75: nit: `copyTimeout` is set via `setTimeout` but never cleared on unmount. If the card unmounts within the 2.5s window (e.g. the user clicks Done right after Copy), the pending `setCopied(false)` fires on an unmounted component. Consider clearing it in a `useEffect` cleanup: `useEffect(() => () => { if (copyTimeout.current) clearTimeout(copyTimeout.current); }, [])`.

features/crypto/lib/rewrapFlow.ts
naeemba: Fixed in 3c20714. Extracted a shared `getMaterial` into `features/crypto/lib/cryptoMaterial.ts` (importing the `CryptoMaterial` contract type from `@/lib/crypto/setupSchema`) and switched both `unlockFlow.ts` and `rewrapFlow.ts` to import it — removing the duplicated `Material` type and `getMaterial()`. Single source of truth for the `/api/crypto/material` contract.

features/settings/ChangePassphraseCard.tsx
naeemba: Fixed in 188e24c. Dropped the unused `formRef` and the `useRef` import.

features/settings/RotateRecoveryCard.tsx
naeemba: Fixed in 188e24c. Added a `useEffect` cleanup that clears `copyTimeout.current` on unmount, so `setCopied(false)` can't fire after the card unmounts within the 2.5s window.

PR #31, opened 2026-06-25
feat(crypto): encrypted journals — passkey (PRF) unlock

naeemba:
## Encrypted journals — passkey (PRF) unlock

Adds a third way to unlock the zero-knowledge encrypted journal: the user's **passkey**, alongside the existing passphrase and one-time recovery code. Purely additive — the DEK never changes; passphrase/recovery wraps are untouched. Multiple passkeys are supported (each carries its own wrap).

Spec: `docs/superpowers/specs/2026-06-25-encrypted-journals-passkey-unlock-design.md`
Plan: `docs/superpowers/plans/2026-06-25-encrypted-journals-passkey-unlock.md`

### How it works
A passkey carries an additional opaque wrap of the per-user DEK, derived from the authenticator's **WebAuthn PRF** output: `PRF → HKDF-SHA256(info "ledger-passkey-v1") → AES-GCM-256 KEK`. PRF-based unlock needs **no server-side assertion verification** — the assertion uses a random challenge and the DEK either unwraps client-side or it doesn't (same model as the passphrase path). Reuses the existing login passkeys via the PRF extension.

### What's included
- `cryptoPasskeyWrap` one-to-many table + migration `0005`; `PasskeyWrapRepository`.
- `derivePrfKek` + WebAuthn PRF assertion helpers (`assertPrfForCredential`, `assertPrfAny`).
- `CryptoMaterial` extended with `passkeys[]`; `/api/crypto/material` projects only `{credentialId, prfSalt, wrap}`.
- `enablePasskeyUnlock` / `disablePasskeyUnlock` server actions.
- Client flows `buildPasskeyWrap` (proof-before-mutation) + `unlockWithPasskey`.
- Settings → Security **Passkey unlock** card; **Unlock with passkey** button on `/crypto/unlock`.
- Enables the PRF extension at passkey registration (requires `@naeemba/next-starter` ≥ 0.9.0, bumped here).

### Invariants (verified end-to-end in final review)
- **Zero-knowledge** — server stores/returns only opaque wraps+salts; no DEK/PRF logged; material projection strips `id`/`userId`/`label`/`createdAt` (test-asserted).
- **DEK never changes** — enable/disable only add/remove a wrap; unlock recovers the same DEK (round-trip test).
- **Proof-before-mutation** — enabling requires the passphrase/recovery secret client-side (`obtainDek`); the typed secret never reaches a server action.
- **No last-method guard needed** — passphrase + recovery always remain.

### Tests
571 passing (114 files), type-check + lint clean. Final whole-branch review (opus): **0 Critical / 0 Important.**

### ⚠️ Live acceptance (merge gate — needs Postgres + Garage + a PRF-capable authenticator)
Folds into the encrypted-journals e2e already owed for P2/P3.
- [ ] A newly-registered passkey returns a PRF result on assertion (confirms starter 0.9.0 forwards `registration.extensions`).
- [ ] Settings → Security → Enable unlock (prove with passphrase). Lock. Unlock via Touch ID/PIN → journal decrypts.
- [ ] Second passkey on another device; both unlock independently.
- [ ] Disable one passkey → it no longer unlocks; the other still does; passphrase + recovery still work.
- [ ] Confirm the `/api/auth/passkey/list-user-passkeys` response shape matches the Settings card's parsing.

### Deferred follow-ups (non-blocking)
- Optional setup-wizard passkey step (enable from day one) — intentionally deferred; modifies the wizard state machine, best tested live.
- Cosmetic: migration trailing newline; unlock error preserving root cause.

### DIFF
--- a/features/crypto/lib/webauthn.ts
+++ b/features/crypto/lib/webauthn.ts
@@ -1,0 +1,67 @@
+import { fromBase64 } from './clientCrypto';
+
+export const base64urlToBytes = (s: string): Uint8Array<ArrayBuffer> => {
+  const pad = s.length % 4 === 0 ? '' : '='.repeat(4 - (s.length % 4));
+  const b64 = (s + pad).replace(/-/g, '+').replace(/_/g, '/');
+  return Uint8Array.from(atob(b64), (c) =>
+    c.charCodeAt(0)
+  ) as Uint8Array<ArrayBuffer>;
+};
+
+export const bytesToBase64url = (b: Uint8Array): string =>
+  btoa(String.fromCharCode(...b))
+    .replace(/\+/g, '-')
+    .replace(/\//g, '_')
+    .replace(/=+$/, '');
+
+export type PrfAssertion = { credentialId: string; prfOutput: Uint8Array };
+
+const readPrf = (cred: PublicKeyCredential): Uint8Array => {
+  const first = cred.getClientExtensionResults().prf?.results?.first;
+  if (!first) throw new Error('This device does not support passkey unlock.');
+  return new Uint8Array(first as ArrayBuffer);
+};
+
+/** Single-credential PRF assertion — used when enabling a specific passkey. */
+export const assertPrfForCredential = async (
+  credentialId: string,
+  salt: Uint8Array
+): Promise<PrfAssertion> => {
+  const cred = (await navigator.credentials.get({
+    publicKey: {
+      challenge: crypto.getRandomValues(new Uint8Array(32)),
+      allowCredentials: [
+        { id: base64urlToBytes(credentialId), type: 'public-key' },
+      ],
+      userVerification: 'required',
+      extensions: { prf: { eval: { first: salt as unknown as BufferSource } } },
+    },
+  })) as PublicKeyCredential | null;
+  if (!cred) throw new Error('Passkey prompt was dismissed.');
+  return { credentialId, prfOutput: readPrf(cred) };
+};
+
+/** Multi-credential PRF assertion — used at unlock; the user picks any enrolled passkey. */
+export const assertPrfAny = async (
+  creds: { credentialId: string; prfSalt: string }[]
+): Promise<PrfAssertion> => {
+  const evalByCredential: Record<string, { first: BufferSource }> = {};
+  for (const c of creds) {
+    evalByCredential[c.credentialId] = {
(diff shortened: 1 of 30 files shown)


features/crypto/lib/webauthn.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export const base64urlToBytes = (s: string): Uint8Array<ArrayBuffer> => {
+  const pad = s.length % 4 === 0 ? '' : '='.repeat(4 - (s.length % 4));
+  const b64 = (s + pad).replace(/-/g, '+').replace(/_/g, '/');
+  return Uint8Array.from(atob(b64), (c) =>
+    c.charCodeAt(0)
+  ) as Uint8Array<ArrayBuffer>;
+};
+
+export const bytesToBase64url = (b: Uint8Array): string =>
```
naeemba: nit: `bytesToBase64url` is exported but never used by production code — the only importer is `webauthn.test.ts`. The unlock/enable paths use `cred.id` (already a base64url string) and the stored `credentialId` directly, so nothing round-trips bytes → base64url. Consider dropping the export (and its dedicated round-trip test), or keep only `base64urlToBytes` which is actually used.

lib/crypto/passkeyWrapSchema.ts
```diff
@@ -0,0 +1,25 @@
...
+  .regex(/^[A-Za-z0-9+/]+={0,2}$/)
+  .min(1)
+  .max(512);
+
+// WebAuthn credential id: base64url charset.
+const b64url = z
+  .string()
+  .regex(/^[A-Za-z0-9_-]+={0,2}$/)
+  .min(1)
+  .max(512);
```
naeemba, line 18: nit: the `b64url` charset here (`[A-Za-z0-9_-]`) hard-asserts that better-auth's `credentialID` is base64url. If the starter ever hands back a standard-base64 id (containing `+` or `/`), `enablePasskeyUnlockAction`/`disablePasskeyUnlockAction` would silently reject it with a generic "Invalid request." even though the client-side `base64urlToBytes` would still decode it fine (it only translates `-_`, leaving `+/` intact). This is exactly live-acceptance item 5 ("confirm the list-user-passkeys response shape"); worth a code comment pointing the reader at that gate, since a format mismatch here fails closed and opaquely.

PR #32, opened 2026-06-26
feat(crypto): wizard recovery-gate hardening + inline passkey step

naeemba:
## What & why

Two safeguards for the zero-knowledge encrypted-journals onboarding wizard, both aimed at the same risk: a user can lose their journal forever if they lose *all* unlock methods at once. The encryption is zero-knowledge by design — there is no operator backdoor — so the defences are (a) making sure the recovery code is genuinely saved, and (b) making a second unlock method (passkey) easy to add during onboarding.

This finishes the deferred "Task 11" passkey step from PR #31.

## Changes

**1. Recovery-gate hardening** — the recovery step's "I have saved my recovery code" checkbox could be ticked without ever copying or downloading the code. Now the user must click **Copy** or **Download** before the checkbox unlocks, and "Enable encryption" stays disabled until both.

**2. Inline passkey step** — a new optional step (`why → passphrase → recovery → passkey → encrypting`) lets the user enroll a passkey for unlock without leaving onboarding:
- **Registers a new PRF passkey inline** if the user has none (most magic-link users), and offers to enroll existing passkeys.
- Enrollment is **additive** — the DEK never changes, the journal is not re-encrypted; each passkey is a separate wrap of the same DEK. It runs *before* finalize (the `enable` action needs only the `userCrypto` row + auth, not an unlocked session).
- Always **skippable**; the recovery code remains the mandatory fallback.
- Two new helpers carry the logic and are unit-tested: `registerPasskey`, `enrollPasskeyForUnlock`. The `postDek`+`finalizeEncryption` call moved from the recovery handler to the new passkey handler.

## Design / plan

- Spec: `docs/superpowers/specs/2026-06-26-wizard-recovery-gate-and-passkey-step-design.md`
- Plan: `docs/superpowers/plans/2026-06-26-wizard-recovery-gate-and-passkey-step.md`

## Testing

- `pnpm test` — 576/576 green (helper unit tests for both new functions; static-render test for the new component — the wizard component imports server actions so it is intentionally not import-tested, matching repo convention).
- `pnpm type-check` + `pnpm lint` — clean.
- Built via subagent-driven development: per-task review + a final whole-branch opus review (crypto model verified intact; 0 Critical / 0 Important after fixes).

## Notes for reviewer

- ⚠️ **Live WebAuthn/PRF e2e acceptance is still un-run** (needs a real Postgres + Garage dev env) — same gate as the rest of the encrypted-journals epic. To verify live: add-this-device (two biometric prompts: register, then PRF assert) → enrolled → Continue → lock → unlock with the passkey; plus the Skip path and a no-PRF-device error path.
- Spec drift (deferred, non-blocking): the `registerPasskey` re-fetch fallback and a `passkey.errors.unsupported` string from the spec were not implemented — `addPasskey` always returns `credentialID` in the current better-auth client, and the unsupported case folds into the generic `registerFailed`/`cancelled` copy.
- Incidental: the branch's first push surfaced that `main` had stale deps (PR #31 bumped `package.json` to starter 0.9.0 + `hash-wasm` without re-installing); resolved via `pnpm install`.

### DIFF
--- a/features/crypto/PasskeyStep.tsx
+++ b/features/crypto/PasskeyStep.tsx
@@ -62,16 +62,31 @@ export function PasskeyStep({
   async function handleAdd() {
     setError(null);
     setBusy(ADD);
+    let credentialId: string;
     try {
-      const { credentialId } = await registerPasskey('This device');
-      await enrollPasskeyForUnlock(dek, credentialId, 'This device');
-      setRows((r) => [
-        ...r.filter((x) => x.credentialId !== credentialId),
-        { credentialId, name: 'This device', enabled: true },
-      ]);
-      setEnrolledCount((c) => c + 1);
+      ({ credentialId } = await registerPasskey('This device'));
     } catch (err) {
       setError(friendlyError(err, copy.errors.registerFailed));
+      setBusy(null);
+      return;
+    }
+    // The passkey now exists server-side. Surface it as an un-enrolled row
+    // immediately so a failed enroll below leaves a retryable "Enable unlock"
+    // entry here instead of an invisible passkey only enrollable via Settings.
+    setRows((r) => [
+      ...r.filter((x) => x.credentialId !== credentialId),
+      { credentialId, name: 'This device', enabled: false },
+    ]);
+    try {
+      await enrollPasskeyForUnlock(dek, credentialId, 'This device');
+      setRows((r) =>
+        r.map((x) =>
+          x.credentialId === credentialId ? { ...x, enabled: true } : x
+        )
+      );
+      setEnrolledCount((c) => c + 1);
+    } catch (err) {
+      setError(friendlyError(err, copy.errors.enrollFailed));
     } finally {
       setBusy(null);
     }
--- a/features/crypto/PasskeyStep.tsx
+++ b/features/crypto/PasskeyStep.tsx
@@ -38,10 +38,11 @@ export function PasskeyStep({
 
   useEffect(() => {
     void (async () => {
-      // fetchUserPasskeys swallows its own errors; only getMaterial can throw.
+      // fetchUserPasskeys swallows its own errors (returns null on failure);
+      // this is an optional step so treat a load failure as "no passkeys".
       // Resolve it separately so a getMaterial failure just means
       // "list passkeys as not-enrolled" without re-fetching the list.
-      const passkeys = await fetchUserPasskeys();
+      const passkeys = (await fetchUserPasskeys()) ?? [];
       const enrolled = await getMaterial()
         .then((m) => new Set(m.passkeys.map((p) => p.credentialId)))
         .catch(() => new Set<string>());
--- a/features/crypto/PasskeyStep.tsx
+++ b/features/crypto/PasskeyStep.tsx
@@ -3,11 +3,21 @@
 import { CheckCircle2, KeyRound, Loader2, Plus } from 'lucide-react';
 import { useEffect, useState } from 'react';
 import { CRYPTO_COPY } from './cryptoCopy';
+import { getMaterial } from './lib/cryptoMaterial';
 import { enrollPasskeyForUnlock, registerPasskey } from './lib/passkeyFlow';
 
 type Row = { credentialId: string; name: string; enabled: boolean };
 
 const ADD = '__add__';
+
+function friendlyError(err: unknown, fallback: string): string {
+  if (typeof DOMException !== 'undefined' && err instanceof DOMException) {
+    return err.name === 'NotAllowedError'
+      ? CRYPTO_COPY.passkey.errors.cancelled
+      : fallback;
+  }
+  return err instanceof Error && err.message ? err.message : fallback;
+}
 
 async function fetchPasskeys(): Promise<
   { credentialID: string; name?: string }[]
@@ -39,14 +49,33 @@ export function PasskeyStep({
 
   useEffect(() => {
     void (async () => {
-      const passkeys = await fetchPasskeys();
-      setRows(
-        passkeys.map((p) => ({
-          credentialId: p.credentialID,
-          name: p.name ?? 'Passkey',
-          enabled: false,
-        }))
-      );
+      try {
+        const [passkeys, material] = await Promise.all([
+          fetchPasskeys(),
+          getMaterial(),
+        ]);
+        const enrolled = new Set(material.passkeys.map((p) => p.credentialId));
+        setRows(
+          passkeys.map((p) => ({
+            credentialId: p.credentialID,
+            name: p.name ?? 'Passkey',
+            enabled: enrolled.has(p.credentialID),
+          }))
+        );
+        setEnrolledCount(
+          passkeys.filter((p) => enrolled.has(p.credentialID)).length
+        );
+      } catch {
+        // getMaterial failure: fall back to listing passkeys as not-enrolled.
+        const passkeys = await fetchPasskeys();
+        setRows(
+          passkeys.map((p) => ({
+            credentialId: p.credentialID,
+            name: p.name ?? 'Passkey',
+            enabled: false,
+          }))
+        );
+      }
     })();
   }, []);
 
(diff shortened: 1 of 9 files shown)


features/crypto/PasskeyStep.tsx
```diff
@@ -0,0 +1,214 @@
...
+function friendlyError(err: unknown, fallback: string): string {
+  if (typeof DOMException !== 'undefined' && err instanceof DOMException) {
+    return err.name === 'NotAllowedError'
+      ? CRYPTO_COPY.passkey.errors.cancelled
+      : fallback;
+  }
+  return err instanceof Error && err.message ? err.message : fallback;
+}
+
+async function fetchPasskeys(): Promise<
```
naeemba: The `fetchPasskeys` helper and the `Row`/`AuthPasskey` shapes are copied almost verbatim from the canonical sibling `features/settings/PasskeyUnlockCard.tsx` (its `fetchPasskeys` at lines 20-27, `type Row` / `type AuthPasskey` at lines 17-18). The PR introduces a second spelling of the same `GET /api/auth/passkey/list-user-passkeys` call and the same row shape. Extract a single shared helper (e.g. `fetchUserPasskeys()` returning `{ credentialID: string; name?: string }[]`) into `features/crypto/lib/passkeyFlow.ts` or a colocated `*.util.ts`, and a shared `Row`/`AuthPasskey` type, and have both this step and the settings card consume it — otherwise the endpoint contract and the row-mapping logic can drift between the two. This is the structure/no-reinvention bucket; it also covers the duplicated types.

features/crypto/PasskeyStep.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+  useEffect(() => {
+    void (async () => {
+      try {
+        const [passkeys, material] = await Promise.all([
+          fetchPasskeys(),
+          getMaterial(),
+        ]);
+        const enrolled = new Set(material.passkeys.map((p) => p.credentialId));
+        setRows(
```
naeemba: nit: the `catch` fallback calls `fetchPasskeys()` a second time. Since `fetchPasskeys` already swallows its own errors and returns `[]`, the only thing inside the `Promise.all` that can reject is `getMaterial()` — so the recovery here is purely "list passkeys but treat all as not-enrolled." You already have the passkeys from the failed `Promise.all`? No — they're lost when `Promise.all` rejects, hence the re-fetch. Consider awaiting them separately (`const passkeys = await fetchPasskeys();` then a guarded `getMaterial()`), which avoids the duplicate network round-trip and the duplicated `setRows` mapping block.

features/crypto/lib/passkeyFlow.ts
```diff
@@ -27,6 +29,41 @@ export const buildPasskeyWrap = async (
...
+ */
+export const registerPasskey = async (
+  name: string
+): Promise<{ credentialId: string }> => {
+  const res = await authClient.passkey.addPasskey({ name });
+  if (!res || res.error || !res.data) {
+    throw new Error(res?.error?.message ?? 'Could not create a passkey.');
+  }
+  return {
+    credentialId: (res.data as { credentialID: string }).credentialID,
```
naeemba: The `as { credentialID: string }` cast is unchecked against `data: unknown` (that's how the starter types `addPasskey`'s `data`). The only guard is `!res.data`, which passes for any non-null object. If a better-auth version ever returns success `data` that lacks `credentialID` (or nests it differently), this silently resolves `credentialId: undefined`; `enrollPasskeyForUnlock` then builds and persists a wrap keyed by `undefined`, yielding a passkey that shows as "Enabled" in the UI but can never assert/unlock — a silent zero-knowledge lockout-fallback failure. The PR notes acknowledge the spec's re-fetch fallback was dropped on the assumption `addPasskey` "always returns credentialID." Given the stakes, add a runtime guard: `const id = (res.data as { credentialID?: unknown }).credentialID; if (typeof id !== 'string' || !id) throw new Error('Could not create a passkey.'); return { credentialId: id };`

features/crypto/PasskeyStep.tsx
naeemba: Fixed in 5b8fdfe. Extracted a shared `fetchUserPasskeys()` helper and an `AuthPasskey` type into `features/crypto/lib/passkeyFlow.ts`; both `PasskeyStep` and `PasskeyUnlockCard` now consume them, removing the duplicated `list-user-passkeys` call and row shape.

features/crypto/PasskeyStep.tsx
naeemba: Fixed in 5b8fdfe. The load effect now awaits `fetchUserPasskeys()` separately (it swallows its own errors) and wraps `getMaterial()` in a `.catch(() => new Set())`. The duplicate re-fetch and the duplicated `setRows` mapping block are gone; there's a single mapping path now.

features/crypto/lib/passkeyFlow.ts
naeemba: Fixed in 5b8fdfe. Added a runtime guard: `const id = (res.data as { credentialID?: unknown }).credentialID; if (typeof id !== 'string' || !id) throw new Error('Could not create a passkey.');`. A missing/non-string id now throws instead of silently resolving `credentialId: undefined`.

features/settings/PasskeyUnlockCard.tsx
```diff
@@ -39,7 +32,7 @@ const PasskeyUnlockCard = () => {
     setLoading(true);
     try {
       const [passkeys, material] = await Promise.all([
-        fetchPasskeys(),
+        fetchUserPasskeys(),
```
naeemba, line 35: The shared `fetchUserPasskeys()` swallows HTTP/network errors and returns `[]`, but this card's original `fetchPasskeys()` **threw** on `!res.ok` (`throw new Error('Could not load passkeys')`), and `refresh()`'s `catch` surfaced "Could not load your passkeys." to the user. After this refactor, a non-OK `list-user-passkeys` response no longer reaches that catch — the card silently renders an empty list as if the user has no passkeys, hiding a real load failure (only a `getMaterial()` rejection still trips the catch). The swallow-and-return-`[]` contract is correct for the wizard's optional step, but it's a behavior regression here. Either keep the settings card on a throwing fetch, or have `fetchUserPasskeys` signal failure distinctly (e.g. return `null` on error) so this card can still show its load error.

naeemba: Fixed in 07a3543. `fetchUserPasskeys()` now returns `null` on a non-OK/network error (was `[]`), so callers can distinguish "no passkeys" from "couldn't load." `PasskeyUnlockCard.refresh()` throws on `null`, restoring the "Could not load your passkeys." error path; the optional wizard step coalesces `null` to `[]` and keeps treating a load failure as "no passkeys."

naeemba: Fixed in 07a3543 (verified this round). `fetchUserPasskeys()` now returns `null` on a non-OK/network error; `PasskeyUnlockCard.refresh()` throws on `null` (line 38) and surfaces "Could not load your passkeys." again. The optional wizard step coalesces `null` to `[]`. Resolving.

features/crypto/PasskeyStep.tsx
```diff
@@ -0,0 +1,193 @@
...
+        ...r.filter((x) => x.credentialId !== credentialId),
+        { credentialId, name: 'This device', enabled: true },
+      ]);
+      setEnrolledCount((c) => c + 1);
+    } catch (err) {
+      setError(friendlyError(err, copy.errors.registerFailed));
+    } finally {
+      setBusy(null);
+    }
+  }
```
naeemba, line 93: nit: partial-failure gap in `handleAdd`. `registerPasskey` creates+persists the passkey server-side first; if the subsequent `enrollPasskeyForUnlock` then fails (the second biometric prompt is dismissed, or the device has no PRF), the catch surfaces an error but the just-created passkey is never added to `rows` (which was loaded at mount). The user is now left with an account passkey that is invisible in this step and cannot be re-enrolled here without a reload — only via Settings. Consider, on enroll failure after a successful register, appending the new credential to `rows` as an un-enrolled entry so the inline "Enable unlock" path can retry it. Non-blocking: the recovery code remains the mandatory fallback and the passkey is still enrollable from Settings.

PR #33, opened 2026-06-26
feat(phase7): structured logging (pino + GlitchTip) + audit log

naeemba:
Closes the final two Phase 7 items in `PLAN.md`: **structured logging + an error-tracking destination** and **audit log of journal mutations + security events**. One combined PR; logging is the foundation the audit log emits through.

Spec: `docs/superpowers/specs/2026-06-26-structured-logging-and-audit-log-design.md`
Plan: `docs/superpowers/plans/2026-06-26-structured-logging-and-audit-log.md`

## Part A — Structured logging
- **`lib/log/`** — single pino instance as the one node-runtime logging entry point, with **mandatory secret redaction** (`REDACT_PATHS` censors passphrase/recoveryCode/dek/wrap/password/token/authorization/cookie/secret at top-level and nested) and a documented hard rule: never log journal content/amounts/names. Pretty in dev, plain JSON to stdout in prod/test (transport gated on `NODE_ENV==='development'` to avoid a vitest worker-thread hang).
- **GlitchTip error tracking** via `@sentry/nextjs`, **fully disabled when `SENTRY_DSN`/`NEXT_PUBLIC_SENTRY_DSN` are unset** (no init, no network — dev/test need none). Wired through `instrumentation.ts` (`register` + `onRequestError`), `sentry.{server,edge}.config.ts`, `instrumentation-client.ts`. Error boundaries capture via `Sentry.captureException`. CSP `connect-src` appends the GlitchTip origin only when the DSN is set.
- Migrated all **24 server-side `console.*`** calls to the logger and added a **`no-console` lint rule** (with sanctioned overrides for client components / config / tests).
- New env vars (all optional, safe defaults): `LOG_LEVEL`, `SENTRY_DSN`, `SENTRY_ENVIRONMENT`, `NEXT_PUBLIC_SENTRY_DSN`.

## Part B — Audit log
- **`auditLog`** Postgres table (migration `0006`), `userId` FK `ON DELETE cascade` (rows purge with the account-deletion feature).
- **`lib/audit/`** — `AuditRepository` + **best-effort `AuditService.record()` that never throws** (validation failure or DB-insert rejection is logged and swallowed, so an audit-write problem can never fail or roll back a user's actual action) + Zod action enum + request-meta helper.
- Instrumented at the action/route boundary (journal core untouched):
  - journal mutations `tx.add` / `tx.edit` / `tx.delete` with journal-dir **byte deltas** + `targetUid`; `journal.import`.
  - security events `crypto.unlock` / `lock` / `enable` / `passphrase-change` / `recovery-rotate` / `reset`.
- `account.delete` is **logger-only** (its audit row would cascade-delete with the user).
- Store-only — no viewer UI yet (`listByUser` exists for a future `/settings` Activity card).

## Review & quality
Built via subagent-driven development: each of the 9 tasks implemented + reviewed (spec + quality), then a whole-branch Opus review. The four binding invariants were verified end-to-end: zero-knowledge, audit-never-throws, Sentry DSN-gated, no-console enforced.

**One must-fix from the final review, resolved (b962314):** the tx actions originally persisted `result.message` into `auditLog.detail`, and for parse failures that string embeds `verify.firstLine` (ledger stderr, path-redacted only) — a durable journal-content leak into the audit table. Now all actions persist a **structured `reason` discriminant** (`invalid`/`quota`/`write-failed`/`parse-failed`/`stale`), never ledger text; `formError` stays on the return value for user-facing display only.

`588/588 tests green · type-check clean · lint clean.`

## Not in this PR (follow-ups, reviewer-triaged non-blocking)
- ⚠️ **Live e2e is un-run** — needs real Postgres + Garage + a GlitchTip DSN (same standing gate as the rest of the encrypted-journals epic). Manual smoke checklist is in the plan.
- Mock `@/lib/audit` in the crypto-action tests to silence swallowed-error log noise (those actions hit the real `auditService` singleton in tests — harmless, and actually demonstrates the never-throws contract, but noisy).
- Record `journal.import` quota-exceeded / hard-error failure paths (only the parse-failure import path is audited today).
- Minor: `headers.ts` `new URL()` try/catch; `listByUser` ordering assertion; restore a `download.test.ts` warn assertion.


db/schema/auditLog.ts
```diff
@@ -0,0 +1,29 @@
+import { sql } from 'drizzle-orm';
+import { user } from '@naeemba/next-starter/schema';
+import { integer, jsonb, pgTable, text, timestamp } from 'drizzle-orm/pg-core';
+
+export const auditLog = pgTable('auditLog', {
```
naeemba: **Missing index on `userId` for a high-growth table.** `auditLog` receives a row on every journal mutation and every security event, so it will be the largest and most write/append-heavy table in the schema, yet it has no index on `userId`. `listByUser` (repository.ts) filters by `userId` and orders by `createdAt`, which on Postgres will be a full table scan — Postgres does **not** auto-create an index for FK columns. The sibling per-user tables `savedView` and `template` both index `userId` (via a composite `uniqueIndex` on `(userId, name)`). Add at least an index on `userId` (ideally `(userId, createdAt desc)` to serve the planned Activity-card query) and regenerate the migration.

features/crypto/actions/setupCrypto.ts
```diff
@@ -13,6 +14,12 @@ export async function setupCrypto(input: unknown): Promise<Result> {
 
   const repo = getUserCryptoRepository();
   if (await repo.exists(user.id)) {
+    await auditService.record(user.id, {
+      action: 'crypto.enable',
+      result: 'failure',
+      detail: { reason: 'Encryption is already set up' },
+      ...(await auditRequestMeta()),
```
naeemba, line 21: nit: inconsistent `detail.reason` vocabulary across the audit table. Here (and in `changePassphrase`/`rotateRecovery`) the reason is a full user-facing sentence — `{ reason: 'Encryption is already set up' }` / `{ reason: 'Encryption is not set up.' }` — whereas `crypto.unlock` uses short discriminants (`'not-set-up'`, `'decode'`) and `crypto.reset` uses (`'too-many-attempts'`, `'invalid'`). The PR's stated invariant is a *structured reason discriminant*. These aren't a content leak (they're static app strings), but a future Activity card / analytics over `detail.reason` will have to special-case prose vs. codes. Suggest a stable short code here too, e.g. `'already-set-up'` / `'not-set-up'`.

features/transactions/actions/createTransaction.ts
```diff
@@ -25,7 +27,17 @@ export async function createTransactionAction(
     return { ok: false, formError: 'Transaction payload is not valid JSON' };
   }
 
+  const bytesBefore = await getJournalDirSize(user.id);
```
naeemba, line 30: nit: every successful and failed `tx.add` now walks the journal dir and `stat`s every file **twice** (`getJournalDirSize` before and after the mutation), and the same pattern is added to edit/delete/import. `getJournalDirSize` is safe (it swallows readdir/stat errors so it won't break the action), but it duplicates work the quota check inside `journalService` likely already performs, and runs even on the `reason: 'invalid'` early-return path where nothing was written. Journals are usually small so the cost is negligible today, but consider having the service return the byte delta it already computes for quota enforcement rather than re-walking the directory at the action layer.

db/schema/auditLog.ts
naeemba: Fixed in 914ae9f. Added a composite index `auditLog_user_createdAt` on `(userId, createdAt desc)` in `db/schema/auditLog.ts` — exactly to serve `listByUser` (filter userId, order createdAt desc) and the planned Activity-card query without a full table scan. Regenerated migration `0007_kind_zuras.sql`.

features/crypto/actions/setupCrypto.ts
naeemba: Fixed in cf61d4b. Replaced the prose `detail.reason` values with stable short discriminants: `already-set-up` in `setupCrypto`, and `not-set-up` in `changePassphrase`/`rotateRecovery` — matching the codes already used by `crypto.unlock` (`not-set-up`/`decode`) and `crypto.reset`. The user-facing `error`/`message` strings are intentionally left as prose.

features/transactions/actions/createTransaction.ts
naeemba: Declining this one, as raised (nit / "consider"). Two reasons:

1. The action layer and the service layer measure different things on purpose. The service computes the per-write byte delta solely for *quota enforcement* (does this write fit under the cap), and short-circuits — it does not return before/after for a write that it rejects, and on the `reason: invalid` early-return path it never reaches the quota check at all. The audit record wants the *observed directory size before and after the action* for forensics, which is a strictly different quantity (it includes concurrent writes, prior state, deletions). Threading the quota-internal delta out through the service contract would conflate the two and still leave the early-return path with nothing to record.

2. As noted in the comment, `getJournalDirSize` swallows readdir/stat errors and journals are small, so the cost is negligible. The proposed refactor changes the `journalService` return type across add/edit/delete/import (and all four action call sites + their tests) for no measurable benefit. Not worth the churn now; happy to revisit if the directory walk ever shows up in profiling.

app/api/upload/route.ts
```diff
@@ -51,6 +57,18 @@ export async function POST(req: NextRequest): Promise<NextResponse> {
           { status: 413 }
         );
```
naeemba, line 66: **`journal.import` is the only mutating action that records success but never failure — forensic-coverage gap and an inconsistency with the rest of the audit surface.**

In this route, an audit row is written only on the two success branches (zip / single). The quota-rejection early-returns (`if (result.quotaExceeded)` here and at the single-file branch) and the `catch (e)` at the bottom return *without* calling `auditService.record`. Every other audited mutation records both outcomes — `createTransaction`/`updateTransaction`/`deleteTransaction` emit `result: 'failure'` with `detail.reason` on the failure path, and the crypto actions/routes record `not-set-up`/`decode`/`too-many-attempts`/`invalid` failures. So a user repeatedly trying to import an over-quota or malformed archive leaves no trace in the audit log, while a single rejected `tx.add` does.

Suggest recording the failure paths too, mirroring the tx actions:
- quota branch → `result: 'failure'`, `detail: { kind, reason: 'quota' }` (the `bytesBefore`/`bytesAfter` are already in scope; reuse the short `reason` code vocabulary `quota` already used by the journal service).
- `catch` branch → `result: 'failure'`, `detail: { reason: 'write-failed' }` (best-effort; the record call itself never throws).

Note this also keeps the `reason` vocabulary consistent with the codes settled in cf61d4b.

naeemba: Fixed in 35fb7d1. The upload route now records `result: 'failure'` on all three previously-untraced paths: both quota-rejection early-returns (`detail: { kind, reason: 'quota' }`) and the `catch` block (`detail: { reason: 'write-failed' }`), reusing `bytesBefore`/`bytesAfter` already in scope and the settled reason-code vocabulary. `journal.import` now mirrors the tx actions for both outcomes.

features/transactions/actions/createTransaction.ts
naeemba: Declining this nit (as already noted in the thread). The action-layer before/after directory sizes are an intentionally distinct quantity from the service's quota-internal byte delta — the audit record wants the observed on-disk size for forensics (including concurrent writes/prior state), the service short-circuits on rejection and never reaches the quota check on the `reason: invalid` path. `getJournalDirSize` swallows readdir/stat errors and journals are small, so the cost is negligible; threading the quota delta through the service contract would change add/edit/delete/import return types and all four call sites + tests for no measurable benefit. Happy to revisit if the directory walk ever shows up in profiling.

app/api/upload/route.ts
naeemba: Confirmed resolved in 35fb7d1. Verified in app/api/upload/route.ts: all three previously-untraced failure paths now emit result: failure — the zip quota branch (detail { kind: zip, reason: quota }), the single-file quota branch (detail { kind: single, reason: quota }), and the catch block (detail { reason: write-failed }). Each reuses bytesBefore and the now-once-computed meta and re-stats bytesAfter, mirroring the success branches. Reason codes match the settled vocabulary; no double-record on any path. journal.import now records both outcomes like the tx actions.

PR #34, opened 2026-06-26
feat(activity): per-user Activity viewer (/settings/activity + summary card)

naeemba:
## Summary

Implements the Phase 7 "per-user Activity viewer UI" Future bullet: a read-only, filterable, keyset-paginated view of each user's own audit log.

- **`/settings/activity`** — server-component page. Filter tabs for **type** (All / Transactions / Imports / Security) and **result** (All / Success / Failures), driven entirely by URL search params. Rows expand via native `<details>` (no client JS) to show uid, byte deltas, IP, user-agent, and the raw `detail` metadata. "Load older" keyset pagination.
- **Activity card on `/settings`** — last 3 events + "View all activity →", placed between Security and Danger Zone.
- Reuses the existing `auditLog` table and `AuditService` — **no schema change**.

## Implementation

Built across 6 TDD tasks (subagent-driven, each individually reviewed):
1. Pure `describeAuditEvent` — maps `(action, result, detail)` → friendly label + icon.
2. `AuditRepository.listByUser` extended to an options object: **composite `(createdAt, id)` keyset cursor** + `actions[]` + `result` filters. Backward-compatible.
3. `AuditService.listForUser` — translates the 3 user-facing type groups → raw `crypto.*`/`tx.*`/`journal.import` actions; reads are not best-effort (errors propagate).
4. URL param + cursor encode/decode helpers (rejects untrusted/garbage input).
5. Page UI: `ActivityRow`, `ActivityList`, route shell + skeleton.
6. Settings summary card + wiring + PLAN.md update.

## Verification

- **602/602 tests pass** (full suite); type-check + lint clean.
- Final whole-branch review (opus): **ready to merge**. All cross-cutting risks verified clean — keyset pagination correct end-to-end (no skip/repeat/loop), per-user isolation airtight (`requireUser()` → `eq(userId)`; URL params only narrow *within* the user), untrusted input rejected + queries parameterized, and the **zero-knowledge invariant holds** (audit `detail` is small metadata only — audited every `record()` site; viewer adds no new leak).

## Out of scope (deferred)

The audit-log **retention/pruning cron** was intentionally deferred — the PLAN.md Phase 7 Future bullet stays open. Table growth is fine at current volume.

## Follow-ups (non-blocking, from review)

- Tighten `describe.ts` `COPY` typing toward exhaustiveness over `AUDIT_ACTIONS`.
- `ActivityRow`: pretty-print multi-key `detail` JSON + add byte units.
- Add param tests for zero/negative-ms cursor and `'imports'`/`'all'` parseType.

Spec: `docs/superpowers/specs/2026-06-26-activity-viewer-design.md`
Plan: `docs/superpowers/plans/2026-06-26-activity-viewer.md`

### DIFF
--- a/features/activity/ActivityRow.tsx
+++ b/features/activity/ActivityRow.tsx
@@ -1,12 +1,6 @@
+import formatWhen from './formatWhen.util';
 import type { AuditLog } from '@/db/schema/auditLog';
 import { describeAuditEvent } from '@/lib/audit/describe';
-import getDefaultDateLocale from '@/utils/getDefaultDateLocale';
-
-const formatWhen = (d: Date): string =>
-  new Date(d).toLocaleString(getDefaultDateLocale(), {
-    dateStyle: 'medium',
-    timeStyle: 'short',
-  });
 
 const DetailLine = ({ label, value }: { label: string; value: string }) => (
   <div className="flex gap-2">
--- a/features/activity/ActivityRow.tsx
+++ b/features/activity/ActivityRow.tsx
@@ -1,0 +1,59 @@
+import type { AuditLog } from '@/db/schema/auditLog';
+import { describeAuditEvent } from '@/lib/audit/describe';
+import getDefaultDateLocale from '@/utils/getDefaultDateLocale';
+
+const formatWhen = (d: Date): string =>
+  new Date(d).toLocaleString(getDefaultDateLocale(), {
+    dateStyle: 'medium',
+    timeStyle: 'short',
+  });
+
+const DetailLine = ({ label, value }: { label: string; value: string }) => (
+  <div className="flex gap-2">
+    <span className="w-24 shrink-0 text-muted">{label}</span>
+    <span className="break-all tabular-nums">{value}</span>
+  </div>
+);
+
+const ActivityRow = ({ row }: { row: AuditLog }) => {
+  const { label, icon } = describeAuditEvent(row);
+  const hasBytes = row.bytesBefore !== null || row.bytesAfter !== null;
+  const detailJson =
+    row.detail && Object.keys(row.detail as object).length > 0
+      ? JSON.stringify(row.detail)
+      : null;
+
+  return (
+    <details className="border-b border-border py-2 text-sm">
+      <summary className="flex cursor-pointer list-none items-center gap-3">
+        <span
+          className={`shrink-0 font-medium ${icon === 'success' ? 'text-positive' : 'text-negative'}`}
+          aria-hidden
+        >
+          {icon === 'success' ? '✓' : '✗'}
+        </span>
+        <span className="flex-1">{label}</span>
+        <time
+          className="shrink-0 text-muted"
+          dateTime={new Date(row.createdAt).toISOString()}
+        >
+          {formatWhen(row.createdAt)}
+        </time>
+      </summary>
+      <div className="mt-2 flex flex-col gap-1 pl-6 text-muted">
+        {row.targetUid && <DetailLine label="uid" value={row.targetUid} />}
+        {hasBytes && (
+          <DetailLine
+            label="bytes"
+            value={`${row.bytesBefore ?? '—'} → ${row.bytesAfter ?? '—'}`}
+          />
+        )}
+        {row.ip && <DetailLine label="ip" value={row.ip} />}
+        {row.userAgent && <DetailLine label="device" value={row.userAgent} />}
+        {detailJson && <DetailLine label="detail" value={detailJson} />}
+      </div>
+    </details>
+  );
+};
+
+export default ActivityRow;
--- a/features/activity/params.ts
+++ b/features/activity/params.ts
@@ -1,30 +1,32 @@
-import type { ActivityType, AuditCursor } from '@/lib/audit';
+import {
+  RESULT_FILTERS,
+  type ActivityType,
+  type AuditCursor,
+  type ResultFilter,
+} from '@/lib/audit';
 
 export const ACTIVITY_PAGE_SIZE = 50;
 
 const TYPES: ActivityType[] = ['all', 'transactions', 'imports', 'security'];
-const RESULTS = ['all', 'success', 'failure'] as const;
-type ResultFilter = (typeof RESULTS)[number];
 
 export const parseType = (raw: string | undefined): ActivityType =>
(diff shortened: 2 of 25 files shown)


features/activity/params.ts
```diff
@@ -0,0 +1,41 @@
...
+type ResultFilter = (typeof RESULTS)[number];
+
+export const parseType = (raw: string | undefined): ActivityType =>
+  TYPES.includes(raw as ActivityType) ? (raw as ActivityType) : 'all';
+
+export const parseResult = (raw: string | undefined): ResultFilter =>
+  RESULTS.includes(raw as ResultFilter) ? (raw as ResultFilter) : 'all';
+
+export const encodeCursor = (row: { createdAt: Date; id: string }): string =>
+  `${row.createdAt.getTime()}_${row.id}`;
```
naeemba: **Keyset cursor can skip rows that share a millisecond.** `createdAt` is a Postgres `timestamp` with no explicit precision, so it stores **microsecond** precision (rows are inserted with `now()` server-side, e.g. a multi-event import or rapid tx writes will commonly land in the same wall-clock ms but differ in µs). The postgres.js driver, however, parses `timestamp` into a JS `Date`, which is **millisecond**-precision — so `encodeCursor(row)` does `row.createdAt.getTime()` on an already-truncated value, and `decodeCursor` rebuilds `new Date(ms)`.

When this ms-truncated cursor is fed back into `repository.listByUser`, the keyset predicate is `createdAt < before.createdAt OR (createdAt = before.createdAt AND id < before.id)`. For a boundary row actually stored as `…123456`:
- `eq(createdAt, before.createdAt)` compares stored `…123456` against the truncated param `…123000` → **never matches**, so the `id` tiebreak branch is dead.
- `lt(createdAt, …123000)` excludes `…123456` and any sibling rows in that same ms.

Net effect: rows sharing a millisecond with the page-boundary row are **silently dropped** across the page boundary (no error, just missing audit events). The composite cursor was added specifically to avoid skip/repeat on tied timestamps, but ms-truncation defeats it.

Fix options: declare the column `timestamp('createdAt', { precision: 3 })` so stored precision matches what the driver returns and what the cursor encodes (cleanest, makes the invariant explicit); or key the cursor on `id` alone — `id` is a ULID, monotonic and unique, and `desc(id)` already orders effectively by time, so an `id`-only `lt(auditLog.id, before.id)` keyset avoids the timestamp-precision problem entirely.

features/activity/ActivityRow.tsx
```diff
@@ -0,0 +1,59 @@
+import type { AuditLog } from '@/db/schema/auditLog';
+import { describeAuditEvent } from '@/lib/audit/describe';
+import getDefaultDateLocale from '@/utils/getDefaultDateLocale';
+
+const formatWhen = (d: Date): string =>
+  new Date(d).toLocaleString(getDefaultDateLocale(), {
+    dateStyle: 'medium',
+    timeStyle: 'short',
+  });
```
naeemba: **Duplicated `formatWhen` helper.** This exact 5-line `formatWhen` (locale `toLocaleString` with `dateStyle: 'medium', timeStyle: 'short'`) is copy-pasted verbatim into `features/activity/ActivityCard.tsx` (lines 8-12). Two copies will drift. This is a feature-specific date+time formatter not covered by `utils/formatDate.ts` (which only does date-only `Format` enum values), so extract it once — either a colocated `features/activity/formatWhen.util.ts` imported by both components, or add a date+time variant to `utils/formatDate.ts` and reuse it. Reference: `PriceStatus.tsx` formats a similar timestamp inline, so a shared date-time formatter would help there too.

features/activity/params.ts
```diff
@@ -0,0 +1,41 @@
+import type { ActivityType, AuditCursor } from '@/lib/audit';
+
+export const ACTIVITY_PAGE_SIZE = 50;
+
+const TYPES: ActivityType[] = ['all', 'transactions', 'imports', 'security'];
+const RESULTS = ['all', 'success', 'failure'] as const;
+type ResultFilter = (typeof RESULTS)[number];
```
naeemba: nit: `ResultFilter` (`'all' | 'success' | 'failure'`) is re-spelled inline here, in `ActivityList.tsx` (line 8), and as an inline param type in `AuditService.listForUser`. `params.ts` already derives it cleanly as `type ResultFilter = (typeof RESULTS)[number]`. Consider exporting that one and reusing it (or `z.infer`-deriving from a shared schema) so the result-filter union has a single source of truth.

features/activity/params.ts
naeemba: Fixed in e4d8317. Went with the id-only keyset (your recommended cleanest option): cursor and ordering now key on the ULID `id` alone (`lt(auditLog.id, before.id)` / `orderBy(desc(auditLog.id))`), so the µs-vs-ms timestamp truncation can no longer drop rows that share a millisecond. `AuditCursor` is now `{ id }`, `encodeCursor`/`decodeCursor` round-trip the ULID (with a ULID-shape regex guard replacing the old ms parse), and the composite index was re-cut as `(userId, id desc)` to match the new order. Note on monotonicity: the `ulid` package's plain `ulid()` is not strictly monotonic within a millisecond, so I rewrote the pagination test to assert the real invariant (stable desc(id) order, no overlap, full coverage) rather than chronological order within a same-ms batch.

features/activity/ActivityRow.tsx
naeemba: Fixed in c6d4a39. Extracted the shared formatter to `features/activity/formatWhen.util.ts`; both `ActivityRow` and `ActivityCard` now import it, so the two copies can no longer drift. (Left `PriceStatus.tsx` out of scope for this PR.)

features/activity/params.ts
naeemba: Fixed in e4d8317. Hoisted the union to `lib/audit` as the single source of truth — `RESULT_FILTERS` tuple plus `type ResultFilter = (typeof RESULT_FILTERS)[number]`, both exported from `@/lib/audit`. `params.ts` now imports `ResultFilter` (and uses `RESULT_FILTERS` for runtime validation), and the inline re-spellings in `ActivityList.tsx` and `AuditService.listForUser` were replaced with the shared type. Put it in lib rather than features so the backend service can share it too (features/ cannot be imported by lib/).

PR #35, opened 2026-06-26
feat(prices): manual price records (/prices)

naeemba:
## Summary

Adds a top-level **`/prices`** page where a user records dated exchange rates for commodities their price provider doesn't cover — e.g. **KIRT** (Kilo IRT), a custom unit no feed knows about. Manual rates are stored per-user and merged into the regenerated `price-db.ledger`, so ledger's `-X` conversions and historical valuation use them and they survive every provider refresh.

Closes the gap noted in the price-db banner ("Manual price overrides belong in your main journal") with a first-class, DB-backed flow instead.

## How it works

- **New per-user `manual_price` table** (the existing `commodity_price` table is global — `BTC=USD` is the same for everyone; a custom KIRT valuation is not, so it gets its own table scoped by `userId`).
- **Batch entry:** pick a shared **date** (+ optional **time**) and **quote currency**, then add one or more rows of **commodity** (autocomplete that also accepts brand-new symbols like KIRT) + **rate**.
- **Valuation over time:** every rate is a separate dated `P` directive — old dates are never overwritten, so ledger picks the rate in effect at each report date. Blank time defaults to end-of-day UTC, making a date-only manual rate authoritative for that day (and it beats a same-day fetched rate via "later timestamp wins").
- **Merged into `price-db.ledger`:** manual rows are always emitted — even for symbols never transacted — through the same `formatLedgerDateTime` render path as fetched rows.

## Changes

- `db/schema/manualPrice.ts` + migration `0009` — per-user table, unique on `(userId, symbol, quote, pricedAt)`.
- `lib/prices/manualRepository.ts` — CRUD (`upsertMany`, `listForUser`, `deleteForUser`).
- `lib/prices/manualSchema.ts` — Zod draft schema + `buildPricedAt` (end-of-day default).
- `lib/prices/service.ts` — `addManualPrices` / `listManualPrices` / `deleteManualPrice` / `getBaseCurrency` / `listCommoditiesForUser`; `regenerateUserPriceDb` now merges manual rows (manual-wins-on-tie).
- `features/prices/actions/` — `addManualPrices` / `deleteManualPrice` server actions (auth → rate-limit → validate → service → audit → revalidate); `price.add`/`price.delete` audit actions.
- `app/prices/page.tsx` + `features/prices/PricesView.tsx` — page shell + batch form & history/delete.
- `components/nav/config.ts` — Prices nav entry.

## Testing

- 619 tests pass (added repository, schema, and service tests — including real-DB assertions that a never-transacted KIRT lands in `price-db.ledger` and that a same-day manual rate sorts after the fetched one).
- `pnpm type-check` and `pnpm lint` clean.

## Follow-ups (non-blocking)

- Per-row Rate label association for rows 2+ (a11y polish).
- Dynamic-row keys use index (mid-list delete focus).
- `price.delete` audit records `success` even on a 0-row no-op.

Design spec and implementation plan are committed under `docs/superpowers/`.

### DIFF
--- a/features/prices/PricesView.tsx
+++ b/features/prices/PricesView.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import { Trash2 } from 'lucide-react';
-import { useActionState, useState } from 'react';
+import { useActionState, useState, useTransition } from 'react';
 import Combobox from '@/components/Combobox/Combobox';
 import { Button } from '@/components/ui/button';
 import { Input } from '@/components/ui/input';
@@ -12,7 +12,7 @@ import {
   deleteManualPriceAction,
   type PriceActionState,
 } from '@/features/prices/actions';
-import { formatLedgerDateTime } from '@/utils/formatDate';
+import { formatLedgerInstant } from '@/utils/formatDate';
 
 type Row = { symbol: string; price: string };
 
@@ -34,6 +34,20 @@ export const PricesView = ({ prices, commodities, baseCurrency }: Props) => {
   const [time, setTime] = useState('');
   const [quote, setQuote] = useState(baseCurrency);
   const [rows, setRows] = useState<Row[]>([{ symbol: '', price: '' }]);
+
+  const [isDeleting, startDelete] = useTransition();
+  const [deletingId, setDeletingId] = useState<number | null>(null);
+  const [deleteError, setDeleteError] = useState<string | null>(null);
+
+  const handleDelete = (id: number) => {
+    setDeleteError(null);
+    setDeletingId(id);
+    startDelete(async () => {
+      const result = await deleteManualPriceAction(id);
+      setDeletingId(null);
+      if (!result.ok) setDeleteError(result.message);
+    });
+  };
 
   const updateRow = (i: number, patch: Partial<Row>) =>
     setRows((rs) => rs.map((r, j) => (j === i ? { ...r, ...patch } : r)));
@@ -151,6 +165,9 @@ export const PricesView = ({ prices, commodities, baseCurrency }: Props) => {
 
       <section className="space-y-2">
         <h2 className="text-lg font-medium">History</h2>
+        {deleteError && (
+          <p className="text-destructive text-sm">{deleteError}</p>
+        )}
         {prices.length === 0 ? (
           <p className="text-muted-foreground text-sm">No manual prices yet.</p>
         ) : (
@@ -168,23 +185,22 @@ export const PricesView = ({ prices, commodities, baseCurrency }: Props) => {
               {prices.map((p) => (
                 <tr key={p.id} className="border-t">
                   <td className="py-1">
-                    {formatLedgerDateTime(new Date(p.pricedAt))}
+                    {formatLedgerInstant(new Date(p.pricedAt))}
                   </td>
                   <td>{p.symbol}</td>
                   <td className="text-right tabular-nums">{p.price}</td>
                   <td>{p.quote}</td>
                   <td className="text-right">
-                    <form action={deleteManualPriceAction}>
-                      <input type="hidden" name="id" value={p.id} />
-                      <Button
-                        type="submit"
-                        variant="ghost"
-                        size="icon"
-                        aria-label={`Delete ${p.symbol} rate`}
-                      >
-                        <Trash2 className="h-4 w-4" />
-                      </Button>
-                    </form>
+                    <Button
+                      type="button"
+                      variant="ghost"
+                      size="icon"
+                      aria-label={`Delete ${p.symbol} rate`}
+                      disabled={isDeleting && deletingId === p.id}
+                      onClick={() => handleDelete(p.id)}
+                    >
+                      <Trash2 className="h-4 w-4" />
+                    </Button>
                   </td>
                 </tr>
               ))}
--- a/features/prices/PricesView.tsx
+++ b/features/prices/PricesView.tsx
@@ -1,0 +1,197 @@
+'use client';
+
+import { Trash2 } from 'lucide-react';
+import { useActionState, useState } from 'react';
+import Combobox from '@/components/Combobox/Combobox';
+import { Button } from '@/components/ui/button';
+import { Input } from '@/components/ui/input';
+import { Label } from '@/components/ui/label';
+import type { ManualPrice } from '@/db/schema';
+import {
+  addManualPricesAction,
+  deleteManualPriceAction,
+  type PriceActionState,
+} from '@/features/prices/actions';
+import { formatLedgerDateTime } from '@/utils/formatDate';
+
+type Row = { symbol: string; price: string };
+
+type Props = {
+  prices: ManualPrice[];
+  commodities: string[];
+  baseCurrency: string;
+};
+
+const todayUtc = () => new Date().toISOString().slice(0, 10);
+
+export const PricesView = ({ prices, commodities, baseCurrency }: Props) => {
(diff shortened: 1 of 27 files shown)


features/prices/actions/deleteManualPrice.ts
```diff
@@ -0,0 +1,26 @@
...
+import { auditService, auditRequestMeta } from '@/lib/audit';
+import { requireUser } from '@/lib/auth/require-user';
+import { priceService } from '@/lib/prices';
+import { rateLimit, WRITE } from '@/lib/rate-limit';
+import { revalidatePath } from 'next/cache';
+
+export async function deleteManualPriceAction(
+  formData: FormData
+): Promise<void> {
+  const user = await requireUser();
```
naeemba, line 16: Every sibling delete action in this codebase (`deleteTemplate.ts`, `deleteSavedView.ts`, `deleteTransaction.ts`) returns a result envelope `{ ok: true } | { ok: false; message }` and is invoked imperatively so the client can surface failures. This action instead is a raw `(FormData) => Promise<void>` wired straight into `<form action>`, which silently swallows the rate-limit block (`if (!rateLimit(...).allowed) return;`), the invalid-id guard, and any thrown delete error — the user sees nothing and the row stays. It also unconditionally records `result: 'success'` even when nothing was deleted (e.g. wrong owner → `deleteForUser` no-op, or rate-limited return before the audit call is skipped but a successful no-op still logs success). Consider matching the sibling pattern: return a result the `PricesView` form can react to, and derive the audit `result` from whether a row was actually removed. (PR notes the success-on-no-op as a follow-up; flagging the broader silent-failure divergence here.)

lib/prices/service.ts
```diff
@@ -63,18 +68,83 @@ export class PriceService {
...
+    );
+    const body = renderPriceDb(merged);
     const target = path.join(layout.dir, PRICE_DB_NAME);
     await this.deps.journalRepo.writeFileAtomic(target, body);
     try {
       revalidateTag(getJournalCacheTag(userId), 'max');
     } catch {
-      // revalidateTag throws outside a Next.js request context (e.g. cron, tests).
-      // This is acceptable — the cache will be invalidated on the next request.
+      // revalidateTag throws outside a Next.js request context (cron, tests).
```
naeemba, line 97: nit: this inline return type `{ ok: true } | { ok: false; formError: string }` re-spells the same success/error concept that `PriceActionState` (`features/prices/actions/types.ts`) already models, and `addManualPricesAction` immediately maps between the two. Consider exporting a single shared result type (e.g. reuse/extend `PriceActionState`, or a discriminated `{ ok: true } | { ok: false; formError: string }` in one place) and referencing it from both the service and the action so the shape can't drift.

lib/prices/service.ts
```diff
@@ -63,18 +68,83 @@ export class PriceService {
...
+    await this.deps.manualRepo.deleteForUser(userId, id);
+    await this.regenerateUserPriceDb(userId);
+  }
+
+  async getBaseCurrency(userId: string): Promise<string> {
+    return this.resolveBaseCurrency(userId);
+  }
+
+  async listCommoditiesForUser(userId: string): Promise<string[]> {
+    return this.listNormalizedSymbolsForUser(userId);
```
naeemba: nit: `getBaseCurrency` and `listCommoditiesForUser` are thin pass-throughs to `resolveBaseCurrency` / `listNormalizedSymbolsForUser`, which this PR also flips from `private` to public to make them callable. That leaves two public names for each operation. Either call the now-public originals directly from the page and drop these wrappers, or keep the wrappers and leave the originals private — but not both, to avoid an ambiguous public surface.

features/prices/PricesView.tsx
```diff
@@ -0,0 +1,197 @@
...
+                <th className="text-right">Rate</th>
+                <th>Quote</th>
+                <th />
+              </tr>
+            </thead>
+            <tbody>
+              {prices.map((p) => (
+                <tr key={p.id} className="border-t">
+                  <td className="py-1">
+                    {formatLedgerDateTime(new Date(p.pricedAt))}
```
naeemba: nit: the history `When` column always renders the full `formatLedgerDateTime` timestamp, so a date-only entry shows the synthetic `23:59:59` end-of-day sentinel (the implementation detail used purely for ledger ordering). The design spec called for showing the time only "when not the end-of-day default" — consider suppressing `23:59:59` so a date-only rate reads as a plain date to the user.

features/prices/actions/deleteManualPrice.ts
naeemba: Fixed in e6cd4b1. `deleteManualPriceAction` now returns `{ ok: true } | { ok: false; message }` and `PricesView` invokes it imperatively via `useTransition`, surfacing rate-limit blocks, the invalid-id guard and any thrown delete error inline. `deleteForUser`/`deleteManualPrice` return whether a row was actually removed, and the audit `result` is derived from that (a wrong-owner no-op now records `failure`).

lib/prices/service.ts
naeemba: Fixed in e6cd4b1. Exported a shared `ManualPriceResult` (`{ ok: true } | { ok: false; formError: string }`) from `lib/prices/service` and referenced it from `addManualPrices` so the success/error shape is defined in one place.

lib/prices/service.ts
naeemba: Fixed in e6cd4b1. Dropped the `getBaseCurrency`/`listCommoditiesForUser` wrappers and call the now-public `resolveBaseCurrency`/`listNormalizedSymbolsForUser` directly from `app/prices/page.tsx`, so there's a single public name per operation.

features/prices/PricesView.tsx
naeemba: Fixed in 211895c. Added `formatLedgerInstant`, which suppresses the `23:59:59` end-of-day sentinel and renders a plain `YYYY/MM/DD`, and used it for the history `When` column so a date-only rate reads as a plain date.

lib/audit/schema.ts
```diff
@@ -5,6 +5,8 @@ export const AUDIT_ACTIONS = [
   'tx.edit',
   'tx.delete',
   'journal.import',
+  'price.add',
+  'price.delete',
```
naeemba, line 9: `price.add`/`price.delete` are registered here and recorded from both actions, but no matching entry was added to the `COPY` map in `lib/audit/describe.ts`. Every other mutating action (`tx.add`, `crypto.*`, `journal.import`) has a `[successLabel, failureLabel]` pair there; without one, `describeAuditEvent` falls through to the `row.action.replace(/[._]/g, ' ')` fallback, so the activity log renders the raw `price add` / `price delete` instead of plain-English copy like the rest of the log.

The `every action renders a non-empty label` test in `describe.test.ts` still passes (the fallback is non-empty), which is why this slipped past the earlier rounds — the gap is the missing human-readable label, not an empty one. Add entries to `describe.ts`, e.g. `'price.add': ['Recorded a price', 'Failed to record a price']` and `'price.delete': ['Deleted a price', 'Failed to delete a price']`, to match the sibling pattern.

naeemba: Fixed in 6e05483 — added `'price.add': ['Recorded a price', 'Failed to record a price']` and `'price.delete': ['Deleted a price', 'Failed to delete a price']` to the COPY map in `lib/audit/describe.ts`, so the activity log now renders plain-English copy instead of the raw action-name fallback. Verified `describe.test.ts` still passes.

PR #36, opened 2026-06-27
feat(mobile): M0 foundational responsive primitives + globals

naeemba:
## Stacked PR (1/6)

First in a 6-PR mobile-responsiveness stack. **Merge in order.** Subsequent PRs branch off this one.

## Summary — Phase M0 (foundational fixes)

Cheap, global changes to shared primitives + `globals.css` that improve every screen at once. No desktop regressions — all bumps are mobile-first with `md:`/`sm:` restoring the original desktop density.

### Changes
- **Reusable responsive table wrapper** (`components/ui/table.tsx`, new): adds `<TableScroll>` providing `overflow-x-auto` + iOS momentum scroll (`[-webkit-overflow-scrolling:touch]`) + mobile edge-bleed (`-mx-3 px-3 sm:mx-0 sm:px-0`) so wide tables scroll inside a contained area and text isn't flush to the screen edge. The file also exports the standard shadcn `Table` primitives so M1 PRs can consume a clean API. Existing raw `<table>` call sites are untouched.
- **`100vh` → `100dvh`** on `body` (`app/globals.css`) so mobile browser chrome / the on-screen keyboard don't push the footer + submit buttons off-screen.
- **Mobile inner gutter** baked into `TableScroll` (item 3) so `p-0`/`gap-0` table cards no longer leave content flush against the border.
- **Touch targets ≥ 44px** in primitives:
  - `input.tsx`: `h-8` → `h-10` on mobile, `md:h-8` restores desktop; keeps `text-base` (with `md:text-sm`) to avoid iOS zoom.
  - `button.tsx`: `icon` / `icon-xs` / `icon-sm` / `icon-lg` get a 44px mobile hit-area (`size-11`), `md:` restores original sizes.
  - `sidebar.tsx`: menu-button default/sm heights → 44/40px on mobile, `md:` restores `h-8`/`h-7`.
- **Horizontal-overflow guard**: `overflow-x: hidden` + `max-width: 100%` on `html, body` so a single wide offender can't make the whole page pannable (intentional horizontal scroll lives inside `TableScroll`).

### Acceptance criteria met
- 375px viewport: no page scrolls horizontally except intentionally inside a table scroll container.
- Primary inputs/buttons/sidebar items are tappable (≥44px) on mobile; desktop density preserved via `md:`.
- Footer/submit stays reachable with the keyboard open (`100dvh`).

## Verification (all green)
- `pnpm lint` — clean
- `pnpm type-check` — clean
- `pnpm test` — **624 passed (124 files)**. (Expected pino error logs print for failure-path tests; the summary confirms all passed.)

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -149,7 +149,11 @@ html,
 body {
   background: var(--bg);
   color: var(--fg);
-  min-height: 100vh;
+  min-height: 100dvh;
+  /* Horizontal-overflow guard: a single wide offender can't make the whole
+     page pannable. Intentional horizontal scroll lives inside TableScroll. */
+  max-width: 100%;
+  overflow-x: hidden;
 }
 
 body {
--- a/app/globals.css
+++ b/app/globals.css
@@ -149,9 +149,20 @@ html,
 body {
   background: var(--bg);
   color: var(--fg);
+  /* Progressive enhancement: keep the 100vh fallback for engines without
+     dynamic-viewport-unit support, then override with 100dvh where available. */
+  min-height: 100vh;
   min-height: 100dvh;
-  /* Horizontal-overflow guard: a single wide offender can't make the whole
-     page pannable. Intentional horizontal scroll lives inside TableScroll. */
+}
+
+/* Horizontal-overflow guard: a single wide offender can't make the whole
+   page pannable. Intentional horizontal scroll lives inside TableScroll.
+   Scoped to the inner content box — a *sibling* of the sticky AppHeader, not
+   an ancestor — so neither the viewport nor any ancestor of the header becomes
+   an overflow scroll container. (Setting overflow on the scroll root, or on an
+   ancestor of a `position: sticky` element, demotes the header to a no-op in
+   some engines, notably Safari.) */
+.app-content-box {
   max-width: 100%;
   overflow-x: hidden;
 }
--- a/components/ui/table.tsx
+++ b/components/ui/table.tsx
@@ -1,0 +1,150 @@
+import * as React from 'react';
+import { cn } from '@/lib/utils';
+
+/**
+ * TableScroll — reusable responsive scroll container for wide data tables.
+ *
+ * On mobile a wide `<table>` overflows the viewport and clips. Wrapping it in
+ * `<TableScroll>` makes only the table scroll horizontally (with iOS momentum)
+ * while the rest of the page stays put. It also:
+ *  - bleeds to the card edges on mobile (`-mx-3`) so the scroll track uses the
+ *    full width, then restores normal flow at `sm:` (`sm:mx-0`);
+ *  - re-applies a small inner gutter on mobile (`px-3`) so text isn't flush
+ *    against the screen edge, removed again at `sm:` so desktop is unchanged.
+ *
+ * Designed to wrap raw `<table>` elements (the app's current pattern) as well
+ * as the shadcn `<Table>` primitive below. M1 PRs consume this directly.
+ */
+function TableScroll({
+  className,
+  bleed = true,
+  ...props
+}: React.ComponentProps<'div'> & {
+  /**
+   * When true (default) the container bleeds to the card edges on mobile with a
+   * matching inner gutter so content breathes. Set false to keep the scroll
+   * container within the normal content box (no negative margins).
+   */
+  bleed?: boolean;
+}) {
+  return (
+    <div
+      data-slot="table-scroll"
+      className={cn(
+        'w-full overflow-x-auto [-webkit-overflow-scrolling:touch]',
+        bleed && '-mx-3 px-3 sm:mx-0 sm:px-0',
+        className
+      )}
+      {...props}
+    />
+  );
+}
+
+function Table({ className, ...props }: React.ComponentProps<'table'>) {
+  return (
+    <TableScroll>
+      <table
+        data-slot="table"
+        className={cn('w-full caption-bottom text-sm', className)}
+        {...props}
+      />
+    </TableScroll>
+  );
+}
+
+function TableHeader({ className, ...props }: React.ComponentProps<'thead'>) {
+  return (
+    <thead
+      data-slot="table-header"
+      className={cn('[&_tr]:border-b', className)}
+      {...props}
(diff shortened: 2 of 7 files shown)


app/globals.css
```diff
@@ -149,7 +149,11 @@ html,
 body {
   background: var(--bg);
   color: var(--fg);
-  min-height: 100vh;
+  min-height: 100dvh;
+  /* Horizontal-overflow guard: a single wide offender can't make the whole
+     page pannable. Intentional horizontal scroll lives inside TableScroll. */
+  max-width: 100%;
+  overflow-x: hidden;
```
naeemba, line 167: `overflow-x: hidden` here applies to **both** `html` and `body`. Because the app scrolls on the viewport and `AppHeader` is `sticky top-0` (`components/Header/AppHeader.tsx:49`) — a descendant of `body` with no intermediate `overflow-y-auto` scroll container — setting `overflow` on the scroll root can demote the sticky header to a no-op in some engines (Safari is historically the offender). The M0 acceptance criteria only check horizontal-scroll suppression, not that the header still sticks. Please verify the dashboard header still pins on scroll at 375px (iOS Safari especially) after this change. If it regresses, scope the guard to a wrapper element (e.g. the `sidebar-wrapper`/inset content box) rather than `html, body`, so the sticky scroll root stays clean.

components/ui/table.tsx
```diff
@@ -0,0 +1,150 @@
...
+    <thead
+      data-slot="table-header"
+      className={cn('[&_tr]:border-b', className)}
+      {...props}
+    />
+  );
+}
+
+function TableBody({ className, ...props }: React.ComponentProps<'tbody'>) {
+  return (
```
naeemba, line 66: The new shadcn `Table`/`TableHead`/`TableCell` primitives carry their own Tailwind padding/font/border utilities (`h-10 px-2`, `p-2`, `[&_tr]:border-b`, `text-muted-foreground`, etc.), but `app/globals.css` already styles every `<table>`/`thead th`/`tbody td` globally (padding `0.75rem 1rem`, uppercase 0.7rem headers, border-bottom, hover). When M1 starts rendering `<Table>` from this file, the two systems will both apply and fight (element CSS vs. utility classes — specificity ties broken by source order, not intent), producing doubled/mismatched padding and borders versus the current raw-`<table>` look. Worth deciding now whether this primitive should replace the global `table {}` block or whether these primitives should be stripped of the styling that the global sheet already owns — otherwise each M1 adoption silently restyles its table. The primitive is currently unused, so no runtime impact in this PR; flagging so M1 has a clear contract.

app/globals.css
```diff
@@ -149,7 +149,11 @@ html,
 body {
   background: var(--bg);
   color: var(--fg);
-  min-height: 100vh;
+  min-height: 100dvh;
```
naeemba, line 155: nit: replacing `100vh` outright with `100dvh` removes the fallback for engines that do not support dynamic viewport units (pre-2022 Safari/Chrome). The standard progressive-enhancement form keeps both — declare `min-height: 100vh;` then `min-height: 100dvh;` on the next line so unsupported browsers keep a sane min-height and supporting ones override. Low risk given the target audience, but it is the cheaper-safer form.

app/globals.css
naeemba: Fixed in 1c7bf62. You're right about the Safari sticky-root hazard. Rather than guard `html, body`, I moved the horizontal-overflow guard onto the inner app content box (`.app-content-box` on the `mx-auto max-w-7xl` div in AppShell). That div is a *sibling* of the sticky `AppHeader` (both children of `SidebarInset`), not an ancestor of it and not the viewport scroll root — so `overflow-x: hidden` there clips wide offenders without turning the body (the sticky scroll root) or any ancestor of the header into an overflow scroll container. I avoided putting it on `[data-slot=sidebar-inset]` precisely because that *is* an ancestor of the header, and `overflow-x: hidden` with `overflow-y: visible` computes `overflow-y` to `auto`, which would have re-introduced the exact regression you flagged.

app/globals.css
naeemba: Fixed in 1c7bf62. Restored the progressive-enhancement form: `min-height: 100vh;` followed by `min-height: 100dvh;` so engines without dynamic-viewport-unit support keep a sane min-height and supporting ones override.

components/ui/table.tsx
naeemba: Agreed this needs a contract, and thanks for catching it before M1 adoption. Decision for M1: the shadcn `Table` primitive becomes the single source of truth and the global `table {}` / `thead th` / `tbody td` block in `app/globals.css` gets retired (folded into the primitive's utilities) as part of the first M1 table migration — so the two systems never co-apply. Keeping that out of this M0 PR since the primitive is unused here (no runtime impact, as you note) and the global block still backs every existing raw `<table>`; ripping it out now would restyle all current tables with nothing rendering the primitive yet. Capturing this as the M1 contract rather than changing code in this PR.

app/globals.css
naeemba: Confirmed fixed in 1c7bf62. The guard now lives on `.app-content-box` (globals.css:165), which is a sibling of the sticky `AppHeader` — both are direct children of `SidebarInset` (a `flex flex-col main` with no `overflow` set, sidebar.tsx:306). The viewport/body remains the scroll root and no ancestor of the header becomes an overflow container, so `sticky top-0` is preserved. Resolving.

app/globals.css
naeemba: Confirmed fixed in 1c7bf62. Progressive-enhancement form restored: `min-height: 100vh;` then `min-height: 100dvh;` (globals.css:154-155). Resolving.

components/ui/table.tsx
naeemba: Acknowledged — agreed this is an M1 contract decision, not an M0 defect. The primitive is unused in this PR (no runtime conflict with the global `table {}` block yet), and the documented plan to retire the global table block when the first M1 migration adopts `<Table>` is the right call. Resolving this thread as the captured contract; the global-vs-primitive de-dup will be enforced at M1 adoption time.

PR #37, opened 2026-06-27
feat(mobile): m1 read-first views responsive

naeemba:
## Phase M1 — read-first data views responsive

Makes the highest-traffic read views usable on a 375px phone by reusing M0's
`<TableScroll>` wrapper and adding a mobile card-reflow to the two widest views.
Desktop appearance is unchanged (all mobile changes are gated behind `md:`/`sm:`).

### Changes

**Card-reflow (stacked rows < `md`, original table at `md:`+):**
- **Transactions** (`features/transactions/TransactionTable.tsx`) — the 7-col table
  is hidden below `md` and replaced by a stacked card list: payee + status badge +
  row actions on the top row, date underneath, accounts summary + per-currency
  amounts on the bottom row. The original table is preserved verbatim, wrapped in
  `<TableScroll className="hidden md:block">`. Shared helpers (`formatTxDate`,
  `accountsSummary`) keep the two renderers in sync.
- **Account register** (`app/accounts/[account]/page.tsx`) — row parsing hoisted to a
  `rows` array reused by both renderers. Below `md`, each register entry renders as a
  card (payee + date header, Amount/Total as a definition list) so the multi-currency
  Total column no longer overflows. Desktop table moved inside `<TableScroll bleed={false}>`
  within its bordered card (`hidden ... md:block`).

**Scroll-wrap + de-overflow:**
- **Dashboard "recent"** (`features/dashboard/Dashboard.tsx`) — raw table wrapped in
  `<TableScroll bleed={false}>` (the card already has `overflow-hidden`, so the no-bleed
  variant keeps the scroll inside it). Amount column kept `whitespace-nowrap`; payee/account
  wrap.
- **Balance** (`app/balance/page.tsx`) — same `<TableScroll bleed={false}>` wrap; account
  column wraps, balance column kept `whitespace-nowrap`.

**Grid/layout:**
- **Dashboard journal-health grid** — was `grid-cols-2` on mobile (cramped for 6 stats);
  now `grid-cols-1` below `sm`, unchanged at `sm:`/`lg:`.
- **Account-name wrapping** — long text columns allowed to wrap across the touched views;
  `whitespace-nowrap` retained only on dates and amounts.

### Verification

- `pnpm lint` — clean
- `pnpm type-check` — clean
- `pnpm test` — **124 files / 624 tests passed** (matches baseline). Pino error logs from
  failure-path tests are expected.

### Notes

- `<TableScroll bleed={false}>` is used wherever the table lives inside an
  `overflow-hidden` card, so the horizontal scroll stays contained instead of being clipped.
- No tests touched the table markup (existing tests cover utils/parsers), so none needed updating.

Stacked PR (2/6) — stacked on #36 (M0). Merge after #36.

### DIFF
--- a/features/transactions/TransactionTable.tsx
+++ b/features/transactions/TransactionTable.tsx
@@ -1,6 +1,7 @@
 'use client';
 
 import RowActions from './RowActions';
+import { TableScroll } from '@/components/ui/table';
 import type { Transaction } from '@/lib/journal/parser';
 import formatAmount from '@/utils/formatAmount';
 import Link from 'next/link';
@@ -12,6 +13,19 @@ const statusBadge = (status: Transaction['status']) => {
   if (status === 'pending') return <span className="text-warning">!</span>;
   return null;
 };
+
+const formatTxDate = (date: string) =>
+  new Date(date).toLocaleDateString(undefined, {
+    year: 'numeric',
+    month: '2-digit',
+    day: '2-digit',
+  });
+
+const accountsSummary = (t: Transaction) =>
+  `${t.postings
+    .slice(0, 2)
+    .map((p) => p.account)
+    .join(' → ')}${t.postings.length > 2 ? ' …' : ''}`;
 
 const magnitudeByCurrency = (t: Transaction): Array<[string, number]> => {
   const sums = new Map<string, number>();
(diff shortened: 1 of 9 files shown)


features/transactions/TransactionTable.tsx
```diff
@@ -13,6 +14,19 @@ const statusBadge = (status: Transaction['status']) => {
   return null;
 };
```
naeemba, line 17: nit: `formatTxDate` re-implements `formatDate(date, Format.DATE)` from `@/utils/formatDate` (identical `{year:"numeric", month:"2-digit", day:"2-digit"}` options). The other views touched in this PR (account register, dashboard "recent") render dates via `formatDate(..., Format.DATE)`, so the transactions view is the odd one out — it formats with the raw browser locale (`toLocaleDateString(undefined, ...)`) rather than the app-configured `getDefaultDateLocale()`/`env.DATE_LOCALE`. This was pre-existing inline behavior, but the extraction was a natural moment to align it. Caveat: `formatDate` reads `env` (server-side) and this is a `'use client'` component, so if you reuse it confirm the locale resolves on the client; otherwise leaving it inline is defensible — just flagging the inconsistency.

PR #39, opened 2026-06-27
feat(mobile): m1 remainder tables responsive

naeemba:
## Summary

Completes the **M1 remainder** of the mobile-responsiveness roadmap — wraps the last data tables in the shared `<TableScroll>` container so they scroll smoothly inside a contained area on phones instead of overflowing the viewport, and verifies recharts is responsive.

### Tables wrapped (`<TableScroll>` + de-overflow)
- **Prices** history table — `features/prices/PricesView.tsx`
- **Portfolio** — `features/portfolio/Portfolio.tsx`
- **Cash flow / monthly comparison** — `features/monthlyComparison/MonthlyComparison.tsx`
- **Debts** — `app/debts/page.tsx`
- **Monthly register** — `app/registers/monthly/[account]/page.tsx`
- **Payees** — `features/payees/Payees.tsx`

### Pattern (consistent with M1 read-first work)
- `bleed={false}` on `<TableScroll>` where the table lives inside an `overflow-hidden` card (portfolio, cashflow, debts, monthly register, payees); default bleed for the prices history table (not card-wrapped).
- Dropped `whitespace-nowrap` on long text columns (account / commodity / payee names) so they wrap; **kept** `whitespace-nowrap` on date and amount columns so numbers/dates never clip.
- Mobile-first; desktop unchanged.

### recharts responsiveness
Verified, no code change needed: every chart routes through `Chart` → `ChartContainer` → recharts `<ResponsiveContainer>` with a `w-full` box and `height` set via CSS only. No fixed pixel widths anywhere in `features/`. (The `LineChart`/`PieChart` references in `Landing.tsx` are lucide icons, not recharts.)

### Roadmap
All M1 remainder checkboxes ticked in `MOBILE-ROADMAP.md` (prices, portfolio, cashflow, debts, monthly register, payees, recharts). After this PR, **all M1 items are done**.

## Verification
- `pnpm lint` — pass
- `pnpm type-check` — pass
- `pnpm test` — **624 passed / 124 files** (baseline held; pino audit-log error logs are expected test-env noise)

Stacked PR (4/6) — stacked on #38 (M2). Merge after #38.

PR #40, opened 2026-06-27
feat(mobile): m3 nav drawer auto-close and responsive chrome

naeemba:
## M3 — Navigation & chrome

Mobile-responsiveness roadmap, Phase M3. Makes the sidebar nav usable one-handed on phones and trims heavy mobile chrome.

### Changes
- **Sidebar drawer auto-close on navigation** (the important one). The shadcn `SidebarMenuButton` only fires `onClick` and never calls `setOpenMobile(false)`, so on mobile the drawer stayed open over the content after tapping a nav link. Fixed with a pathname `useEffect` in `components/Sidebar/AppSidebar.tsx`:
  - Chose approach (b) over editing the shared `SidebarMenuButton` primitive — it's the least intrusive option, leaves the primitive (used for non-nav purposes too) untouched, and closes the drawer on **every** navigation entry point: nav links, command palette, breadcrumbs, and browser back/forward. The effect keys off `pathname`; `setOpenMobile` is a stable memoized setter, so it only runs on real route changes and is a no-op on mount (drawer already closed).
- **Responsive content bottom padding** — `pb-20` → `pb-12 sm:pb-20` in `components/AppShell/AppShell.tsx`, lighter on short mobile viewports.
- **Mobile sidebar width** — kept at `18rem`. Now that the overlay auto-closes the width is acceptable, and trimming to 16rem risked truncating longer labels ("Add transaction", "Periodic Balance"). Documented in the roadmap.
- **Header height / safe area** — deferred to M5 (the `viewport-fit=cover` / `env(safe-area-inset-*)` work lands there), per the roadmap. No change here.

### Verification
- `pnpm lint` — clean
- `pnpm type-check` — clean
- `pnpm test` — **624 passed / 124 files** (baseline unchanged; pino audit-log error logs are expected test noise)

No DOM testing library (jsdom / @testing-library) is installed — existing component tests are server static-markup only, so there's no seam to test a client `useEffect` reacting to pathname changes without adding an out-of-scope dependency. The change is a small, well-understood pattern; no brittle test fabricated.

Stacked PR (5/6) — stacked on #39 (M1 remainder). Merge after #39.

### DIFF
--- a/MOBILE-ROADMAP.md
+++ b/MOBILE-ROADMAP.md
@@ -133,17 +133,23 @@ including adding/removing posting rows, with the submit button reachable.
 
 ## M3 — Navigation & chrome
 
-- [ ] 🟠 **Sidebar drawer doesn't auto-close on navigation.** `SidebarMenuButton`
+- [x] 🟠 **Sidebar drawer doesn't auto-close on navigation.** `SidebarMenuButton`
   fires `onClick` but never calls `setOpenMobile(false)`, so the drawer stays
   open over the content after tapping a nav link. Close it on mobile navigation
   (either in `SidebarMenuButton` or via a pathname effect in `AppSidebar`).
   `components/ui/sidebar.tsx:255-288`, `components/Sidebar/AppSidebar.tsx`
-- [ ] 🟡 **Mobile sidebar width** `SIDEBAR_WIDTH_MOBILE = '18rem'` (288px ≈ 77% of a
+  → Done via a pathname `useEffect` in `AppSidebar` (least intrusive; leaves the
+  shared `SidebarMenuButton` primitive untouched; covers every navigation entry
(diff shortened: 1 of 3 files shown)


naeemba (commented):
Reviewed M3 nav drawer auto-close + responsive chrome. The pathname `useEffect` is the right call: `setOpenMobile` is a `useState` setter (stable reference), so the deps are correct and the effect only fires on real route changes; the initial mount run is a no-op since the drawer starts closed. Keying off `pathname` covers every navigation entry point without touching the shared `SidebarMenuButton` primitive, and reuses the `usePathname()` already consumed here. The `pb-12 sm:pb-20` responsive padding sits in the correct presentation layer. No layering, duplication, or correctness issues found.

PR #43, opened 2026-06-27
feat(dev): one-command local stack via docker-compose + pnpm app:dev

naeemba:
## Summary

Adds a docker-compose stack and a `pnpm app:dev` command that brings up everything needed for local development in one shot: Postgres, a Garage (S3-compatible) object store, and the Next dev server.

The app itself runs on the host (fast HMR, reuses the host `ledger` binary); docker-compose only provides the backing services.

## What's included

| File | Purpose |
|------|---------|
| `docker-compose.yml` | **Postgres 17** (host port **5433**) + **Garage v2.3.0** (S3 `:3900`, admin `:3903`), healthchecked, named volumes |
| `docker/garage.dev.toml` | Single-node Garage config with fixed **dev-only** secrets |
| `scripts/app-dev.sh` | Bootstraps `.env` if absent → brings services up `--wait`-healthy → provisions Garage (layout + `ledger` bucket + access key, idempotent) → runs `pnpm dev` |
| `package.json` | `app:dev`, `app:down`, `app:reset` |

### Console email transport
Magic-link sign-in needs Postal vars, which don't exist locally. Added `EMAIL_TRANSPORT=console` (default in `.env.example`) which **prints the magic link/code to the dev server console** instead of sending mail. `POSTAL_*` are now required only when `EMAIL_TRANSPORT=postal` (mirrors the existing `STORAGE_BACKEND=s3` pattern). All three email call sites (auth, encryption reset, account deletion) now use the selected transport.

## Notes / decisions

- **Port 5433**, not 5432 — avoids clashing with other local Postgres containers (e.g. `raxel_postgres`).
- Garage's `GARAGE_DEFAULT_*` auto-provisioning doesn't fire on a fresh node (no layout yet), so the script provisions the bucket + key explicitly. The official image is distroless, so bootstrap steps are direct `/garage` binary calls.
- `STORAGE_BACKEND` defaults to `s3` (real Garage). Set `memory` for a quick boot with no Garage.

## Verification

- `pnpm type-check`, `pnpm lint` — clean
- **627 tests pass** (125 files)
- From a wiped state: both services reach **healthy**; Garage bootstrap provisions on first run and **no-ops on re-run** (idempotent)
- **Real S3 put/get round-trip** succeeds with the app's exact credentials/client config
- Auth + drizzle **migrations apply** against the running Postgres

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,13 +1,17 @@
 # Required
-DATABASE_URL=postgres://postgres:postgres@localhost:5432/ledger_dev
+DATABASE_URL=postgres://postgres:postgres@localhost:5433/ledger_dev
 BETTER_AUTH_SECRET=replace-with-32-plus-chars-of-random-data
 BETTER_AUTH_URL=http://localhost:3000
 
-# Email (magic link) — delivered via the self-hosted Postal server.
-# Both Postal vars are required: a missing/empty value fails fast at startup.
+# Email (magic link).
+#   EMAIL_TRANSPORT=console → print the magic link / code to the dev server
+#     console; nothing is sent. The default for `pnpm app:dev`.
+#   EMAIL_TRANSPORT=postal  → deliver via the self-hosted Postal server; both
+#     POSTAL_* vars below become required (a missing value fails fast at startup).
+EMAIL_TRANSPORT=console
 EMAIL_FROM=auth@example.com
-POSTAL_API_URL=https://postal.raxel.studio
-POSTAL_API_KEY=
+# POSTAL_API_URL=https://postal.raxel.studio
+# POSTAL_API_KEY=
 
 # Optional: Google sign-in. Set NEXT_PUBLIC_ENABLE_GOOGLE=1 in lockstep.
 # GOOGLE_CLIENT_ID=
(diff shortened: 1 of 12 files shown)


.env.example
```diff
@@ -22,12 +26,14 @@ DATA_DIR=./data
...
-# Set STORAGE_BACKEND=s3 in production and fill the S3_* vars below.
-STORAGE_BACKEND=memory
-# S3_ENDPOINT=http://garage:3900
-# S3_REGION=garage
-# S3_BUCKET=ledger
-# S3_ACCESS_KEY_ID=
-# S3_SECRET_ACCESS_KEY=
-# S3_FORCE_PATH_STYLE=true
+# `pnpm app:dev` runs a Garage container, so the defaults below point at it. The
+# S3_ACCESS_KEY_ID / S3_SECRET_ACCESS_KEY must match GARAGE_DEFAULT_* in
```
naeemba: This comment points at `GARAGE_DEFAULT_*` env vars in `docker-compose.yml`, but no such vars exist there — the PR's own notes confirm Garage's `GARAGE_DEFAULT_*` auto-provisioning doesn't fire on a fresh node, which is why the keys are provisioned by the script instead. The actual source of truth these values must match is the hardcoded `GARAGE_ACCESS_KEY` / `GARAGE_SECRET_KEY` in `scripts/app-dev.sh`. As written, anyone rotating these dev keys would look in `docker-compose.yml`, find nothing, and miss the two places that actually need to stay in sync (this file + `app-dev.sh`). Please point the comment at `scripts/app-dev.sh`.

scripts/app-dev.sh
```diff
@@ -0,0 +1,70 @@
...
+fi
+
+# --- .env bootstrap -----------------------------------------------------------
+# Only create one if it's missing; never clobber an existing secrets file.
+if [ ! -f .env ]; then
+  cp .env.example .env
+  if command -v openssl >/dev/null 2>&1; then
+    secret=$(openssl rand -base64 32)
+    # Portable in-place sed (works on both BSD/macOS and GNU).
+    sed -i.bak "s|^BETTER_AUTH_SECRET=.*|BETTER_AUTH_SECRET=${secret}|" .env && rm -f .env.bak
```
naeemba, line 26: nit: `.env.bak` is created by `sed -i.bak` and only removed on success via `&&`. Because the script runs under `set -euo pipefail`, if `sed` fails the script aborts before `rm`, leaving `.env.bak` — a copy of `.env` containing the freshly generated `BETTER_AUTH_SECRET` — sitting in the repo root. `.env.bak` is not covered by `.gitignore` (which only ignores `.env` and `.env*.local`), so it could be staged and committed by accident. Consider adding `.env.bak` to `.gitignore`, or using a temp path outside the repo for the backup.

.env.example
naeemba: Good catch. Repointed the comment at the GARAGE_ACCESS_KEY / GARAGE_SECRET_KEY provisioned in scripts/app-dev.sh, which is the real source of truth these values must match. Fixed in 87c0d75.

scripts/app-dev.sh
naeemba: Agreed. Added .env.bak to .gitignore so a sed failure under set -euo pipefail can't leave a committable backup of .env (with the generated BETTER_AUTH_SECRET) in the repo root. Fixed in 5c814e5.

PR #44, opened 2026-06-27
feat(transactions): add "Save & add another" button

naeemba:
## Summary

Adds a secondary submit button to the new-transaction form: **Save & add another**. It saves the transaction and resets the form in place for the next entry, instead of redirecting to the dashboard — making it faster to enter several transactions in a row.

The primary **Add transaction** button is unchanged (save → toast → redirect to `/`). The new button is create-mode only; edit mode is untouched.

## How it works

- A `saveAndAddAnother` ref is set by the new button's `onClick` and cleared by the primary button's, so the success effect knows whether to reset-in-place or redirect. The flag is cleared after each save, so an Enter-key submit defaults to redirect.
- Both paths still fire the "Transaction saved" toast and call `router.refresh()` so the account/payee autocomplete lists pick up the new entry.
- A memoized `resetForm` helper restores the initial blank draft (today's date, empty payee/note, two empty postings).

## Testing

- `tsc --noEmit` — passes
- `eslint` — clean

### DIFF
--- a/features/transactions/TransactionForm.tsx
+++ b/features/transactions/TransactionForm.tsx
@@ -57,6 +57,13 @@ const todayISO = (): string => {
 
 const initialState: TransactionActionState = { ok: false };
 
+// Two blank postings — the minimum a transaction needs. Defined once so the
+// initial draft and `resetForm` stay in sync.
+const emptyPostings = (currency: string): Posting[] => [
+  { account: '', amount: '', currency },
+  { account: '', amount: '', currency },
+];
+
 const fieldError = (state: TransactionActionState | null, key: string) =>
   state?.fieldErrors?.[key];
 
@@ -86,10 +93,7 @@ const TransactionForm = ({
       account: p.account,
       amount: p.amount,
       currency: p.currency,
-    })) ?? [
-      { account: '', amount: '', currency: defaultCurrency },
-      { account: '', amount: '', currency: defaultCurrency },
-    ]
+    })) ?? emptyPostings(defaultCurrency)
   );
 
   const formRef = useRef<HTMLFormElement>(null);
@@ -102,10 +106,7 @@ const TransactionForm = ({
     setPayee('');
     setStatus('none');
     setNote('');
-    setPostings([
-      { account: '', amount: '', currency: defaultCurrency },
-      { account: '', amount: '', currency: defaultCurrency },
-    ]);
+    setPostings(emptyPostings(defaultCurrency));
   }, [defaultCurrency]);
 
   useEffect(() => {
--- a/features/transactions/TransactionForm.tsx
+++ b/features/transactions/TransactionForm.tsx
@@ -1,6 +1,12 @@
 'use client';
 
-import { useActionState, useEffect, useRef, useState } from 'react';
+import {
+  useActionState,
+  useCallback,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
 import { toast } from 'sonner';
 import type { TransactionActionState } from './actions';
 import Combobox from '@/components/Combobox';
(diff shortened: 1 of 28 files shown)


features/transactions/TransactionForm.tsx
```diff
@@ -87,16 +93,39 @@ const TransactionForm = ({
...
+    if (!state?.ok) return;
+    if (mode === 'edit') {
+      toast.success('Transaction updated');
+      router.push('/transactions');
       router.refresh();
+      return;
     }
-  }, [state, router, mode]);
+    toast.success('Transaction saved');
+    if (saveAndAddAnother.current) {
```
naeemba: The `saveAndAddAnother` ref is only cleared inside the `state.ok` branch of this effect. If a "Save & add another" submit *fails* (action returns `ok: false`), the effect early-returns at line 112 and the ref stays `true`. The user then corrects the input and submits again — if they do so by pressing Enter (which fires no button `onClick`), the ref is still `true`, so a successful save resets in place instead of redirecting to `/`.

That contradicts the documented contract ("an Enter-key submit defaults to redirect"). Clicking the primary "Add transaction" button still works because its `onClick` resets the ref, but the Enter path does not.

Consider clearing the flag on any settled result rather than only on success — e.g. reset `saveAndAddAnother.current = false` before the `!state.ok` early-return, capturing its value into a local first:

```ts
const addAnother = saveAndAddAnother.current;
saveAndAddAnother.current = false;
if (!state?.ok) return;
...
if (mode !== 'edit' && addAnother) { resetForm(); router.refresh(); return; }
```

features/transactions/TransactionForm.tsx
```diff
@@ -87,16 +93,39 @@ const TransactionForm = ({
...
+
+  const resetForm = useCallback(() => {
+    setDate(todayISO());
+    setPayee('');
+    setStatus('none');
+    setNote('');
+    setPostings([
+      { account: '', amount: '', currency: defaultCurrency },
+      { account: '', amount: '', currency: defaultCurrency },
+    ]);
```
naeemba: nit: the two-blank-postings initializer is now duplicated between the `useState` default (lines 90-92) and `resetForm` here. Consider hoisting a small `emptyPostings()` helper (or a `blankPosting()` returning `{ account: "'"'"'", amount: "'"'"'", currency: defaultCurrency }`) so the blank-draft shape is defined once.

features/transactions/TransactionForm.tsx
naeemba: Good catch — fixed in ce07104. The flag is now captured into a local and the ref is cleared on every settled result before the `!state.ok` early-return, so a failed 'Save & add another' no longer leaves the ref set for the next Enter-key submit.

features/transactions/TransactionForm.tsx
naeemba: Done in 4fc5692 — hoisted an `emptyPostings(currency)` helper used by both the postings useState default and resetForm, so the blank-draft shape is defined once.

naeemba:
Missed muted-text migration: `utils/formatAmount.tsx:24` still renders its blank-value placeholder with `text-muted`:

```tsx
return <span className="text-muted">—</span>;
```

This PR migrated every other muted *text* usage to `text-muted-foreground` and repurposed `--muted` into a dark elevated surface token (dark mode `oklch(27% 0.015 250)`, light mode `oklch(0.97 0 0)`). With that change, this em-dash now resolves to a surface color rather than a foreground color, so it is near-invisible: ~27% gray on a `--card` (22%) / `--bg` (17%) background in dark mode, and near-white on white in light mode.

`formatAmount` renders in ~12 views (Dashboard, accounts, balance, payees, reconcile, portfolio, etc.), so the broken placeholder is widely visible. The file is outside this PR's diff, but it is the one remaining instance of exactly the bug this PR set out to fix — switch it to `text-muted-foreground` to complete the migration.

PR #45, opened 2026-06-28
feat(crypto): single-tap passkey login + journal unlock

naeemba:
## Summary

A passkey user previously tapped their authenticator twice — once to sign in, once to unlock the encrypted journal — because login (a WebAuthn signature) and decryption (a WebAuthn PRF output) were obtained in two separate ceremonies. This collapses them into a single tap.

The login ceremony now requests the PRF extension and feeds its output into the existing `derivePrfKek → unwrapDek → postDek` pipeline before redirecting, so an enrolled user lands on the dashboard already unlocked.

## How

- **Fixed PRF salt.** The per-passkey random `prfSalt` required a post-auth server round-trip, which made a single ceremony impossible (the salt must be known *before* the assertion). It's replaced by a fixed versioned constant `PRF_SALT = "ledger-prf-v1"`. This is safe: the PRF output's entropy comes from the authenticator's hardware secret bound to the rpId — the salt is a non-secret application label, and domain separation is already handled by HKDF's `info`.
- **Dropped `prfSalt`** from the schema, types, repository, enroll action, and `/api/crypto/material` response. Migration `0010` wipes stale wraps (made with random salts, now undecryptable) before dropping the column.
- **Single-tap wiring** in `AuthForm.onPasskey`: `signIn.passkey({ extensions: { prf: { eval: { first: PRF_SALT } } }, returnWebAuthnResponse: true })`, then a best-effort `tryUnlockFromWebAuthn` that **never blocks login** — if PRF is absent (unsupported device) or the passkey isn't enrolled for unlock, the user still lands logged in and falls back to the passphrase unlock screen.

## Properties preserved

- **Zero-knowledge:** the server still stores and returns only opaque blobs; the DEK and PRF output never newly cross the wire.
- **Modal prompt** (not conditional-UI autofill), since PRF return is unreliable under conditional mediation.
- Passphrase, recovery-code, and standalone-passkey unlock paths unchanged.

## Notes

- No real users exist on the server yet, so existing wraps are simply wiped rather than migrated; users re-enroll passkey unlock.
- Confirmed `@better-auth/passkey` 1.6.10 forwards `extensions` and returns `clientExtensionResults` — no library fork needed.

## Testing

- `pnpm type-check`, `pnpm lint` clean; `pnpm vitest run` → 633/633 across 125 files.
- New unit tests assert the fixed salt is sent, the correct wrap is selected, unlock no-ops when PRF is absent, and errors are swallowed without blocking login.

Spec: `docs/superpowers/specs/2026-06-28-passkey-single-tap-unlock-design.md`
Plan: `docs/superpowers/plans/2026-06-28-passkey-single-tap-unlock.md`

### DIFF
--- a/features/auth/AuthForm.tsx
+++ b/features/auth/AuthForm.tsx
@@ -12,6 +12,11 @@ import {
   type Method,
 } from './authState';
 import { resolveCallbackUrl } from './callbackUrl';
+import { PRF_SALT } from '@/features/crypto/lib/clientCrypto';
+import {
+  tryUnlockFromWebAuthn,
+  type WebAuthnResult,
+} from '@/features/crypto/lib/passkeyFlow';
 import { authClient } from '@/lib/auth-client';
 import { useWebAuthnSupported } from '@naeemba/next-starter/client';
 import Link from 'next/link';
@@ -107,11 +112,25 @@ export function AuthForm({ mode }: AuthFormProps) {
   }
 
   function onPasskey() {
-    const passkey = authClient.signIn.passkey;
+    let webauthn: WebAuthnResult | undefined;
     return runAttempt(
       'passkey',
-      () => passkey(),
-      () => {
+      async () => {
+        const res = await authClient.signIn.passkey({
+          extensions: {
+            prf: { eval: { first: PRF_SALT as unknown as BufferSource } },
+          },
+          returnWebAuthnResponse: true,
+        } as Parameters<typeof authClient.signIn.passkey>[0]);
+        if (res && 'webauthn' in res && res.webauthn) {
+          webauthn = res.webauthn as unknown as WebAuthnResult;
+        }
+        return res;
+      },
+      async () => {
+        // Best-effort: unlock the journal from the same ceremony's PRF output.
+        // Never throws; falls through to passphrase unlock when unavailable.
+        await tryUnlockFromWebAuthn(webauthn);
         const callbackURL = resolveCallbackUrl('callbackUrl', CALLBACK_URL);
         window.location.assign(callbackURL);
       }
--- a/features/auth/AuthForm.tsx
+++ b/features/auth/AuthForm.tsx
@@ -12,11 +12,7 @@ import {
   type Method,
 } from './authState';
 import { resolveCallbackUrl } from './callbackUrl';
-import { PRF_SALT } from '@/features/crypto/lib/clientCrypto';
-import {
-  tryUnlockFromWebAuthn,
-  type WebAuthnResult,
-} from '@/features/crypto/lib/passkeyFlow';
+import { signInWithPasskey } from '@/features/crypto/lib/passkeyFlow';
 import { authClient } from '@/lib/auth-client';
 import { useWebAuthnSupported } from '@naeemba/next-starter/client';
 import Link from 'next/link';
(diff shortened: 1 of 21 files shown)


features/auth/AuthForm.tsx
```diff
@@ -107,11 +112,25 @@ export function AuthForm({ mode }: AuthFormProps) {
...
+            prf: { eval: { first: PRF_SALT as unknown as BufferSource } },
+          },
+          returnWebAuthnResponse: true,
+        } as Parameters<typeof authClient.signIn.passkey>[0]);
+        if (res && 'webauthn' in res && res.webauthn) {
+          webauthn = res.webauthn as unknown as WebAuthnResult;
+        }
+        return res;
+      },
+      async () => {
```
naeemba: The new single-tap wiring here is the highest-risk part of the change and has no direct test. `passkeyFlow.test.ts` exercises `tryUnlockFromWebAuthn` in isolation, but the AuthForm-specific logic is untested: the `res.webauthn` extraction, the `'webauthn' in res && res.webauthn` guard, that `tryUnlockFromWebAuthn(webauthn)` runs before the redirect, and that login still redirects (and does not surface an error) when unlock is unavailable. `AuthScreen.test.tsx` does not cover the passkey path at all. Worth a focused test that mocks `authClient.signIn.passkey` (both with PRF results and with `error` set) and asserts redirect-vs-error behavior, mirroring how the plan describes the AuthForm test.

features/crypto/lib/clientCrypto.ts
```diff
@@ -3,6 +3,14 @@ import { argon2id } from 'hash-wasm';
...
 const RECOVERY_INFO = new TextEncoder().encode('ledger-recovery-v1');
 const PASSKEY_INFO = new TextEncoder().encode('ledger-passkey-v1');
+/**
+ * Fixed PRF eval input for passkey-derived KEKs. Not a secret and not entropy:
+ * the PRF output is already uniquely bound to (authenticator secret, rpId) by the
+ * hardware, and domain separation is handled by derivePrfKek's HKDF info. A fixed
+ * value lets the login ceremony request PRF without a pre-auth salt fetch. The
+ * version suffix leaves a rotation handle.
+ */
+export const PRF_SALT: Uint8Array = new TextEncoder().encode('ledger-prf-v1');
```
naeemba: nit: `PRF_SALT` is an exported, mutable `Uint8Array` shared across every call site (webauthn assertions, AuthForm extension options, tests). Nothing mutates it today, but a shared crypto input buffer is safer frozen — e.g. derive it once and treat it as read-only, or hand out a copy at each use. Optional.

features/auth/AuthForm.tsx
naeemba: Added focused coverage. I extracted the AuthForm passkey wiring into signInWithPasskey (passkeyFlow.ts) so the previously-untested logic is now a pure, unit-testable unit, and AuthForm is a thin caller. The new tests in passkeyFlow.test.ts mock authClient.signIn.passkey and assert: the PRF salt request, unlock-on-assertion (postDek called), error pass-through with no unlock attempt, redirect-still-proceeds when no PRF/webauthn is returned (error null), and that a swallowed unlock failure still reports success. Done in 18f630c.

features/crypto/lib/clientCrypto.ts
naeemba: Done in f64205a. Replaced the exported mutable PRF_SALT buffer with a prfSalt() accessor that returns a fresh Uint8Array on each call, so no call site can mutate the shared crypto input. Updated the webauthn assertions, AuthForm extension options, and tests accordingly.

PR #46, opened 2026-06-28
fix(combobox): modal dialog on mobile so free-text entry survives

naeemba:
## Problem

On mobile, creating a new account/payee from the add-transaction autocomplete was impossible: after typing, scrolling/tapping toward the **Use "…"** option closed the dropdown and discarded the typed text — so a transaction with a brand-new item couldn't be added.

## Root cause

Two compounding causes:

1. **Base UI's `Popover` hardcodes touch dismissal to `'sloppy'`** (`@base-ui/react` `PopoverRoot.js`), with no public prop to tune it. A touch-drag, or the synthetic tap that lands "outside" once the **soft keyboard reflows the layout**, fires an `outsidePress` close before the tap registers as a selection.
2. **Every close wiped the search** — `handleOpenChange(false)` always ran `setSearch('')`, so any stray dismissal also threw away the typed text.

## Fix

Render the search-and-list inside a **modal `Dialog` on touch/small screens** (`useIsMobile()`, 768px breakpoint) while keeping the **`Popover` unchanged on desktop**. The trigger button and `Command` list are shared between both branches, so commit / free-text / filtering behaviour is identical — only the wrapper differs.

A modal dialog has a backdrop and an anchored, large list as the tap target, so keyboard reflow can't move content out from under the finger and touches inside never count as "outside." The **Use "…"** create option stays reachable and tappable.

This applies everywhere `Combobox` is used (add-transaction accounts/payees, filters, templates, prices, base-currency picker).

## Verification

- `pnpm type-check` — clean
- `eslint` — clean
- Component test suite (8 files / 27 tests) — green

No unit test added: the branch is gated on `matchMedia` + open state, and the component test harness runs in a `node`/SSR-string environment with no `matchMedia`, jsdom, or touch, so it cannot exercise the mobile path. Needs on-device / mobile-emulation verification.

### DIFF
--- a/components/Combobox/Combobox.tsx
+++ b/components/Combobox/Combobox.tsx
@@ -12,10 +12,18 @@ import {
   CommandList,
 } from '@/components/ui/command';
 import {
+  Dialog,
+  DialogContent,
+  DialogDescription,
+  DialogTitle,
+  DialogTrigger,
+} from '@/components/ui/dialog';
+import {
   Popover,
   PopoverContent,
   PopoverTrigger,
 } from '@/components/ui/popover';
+import { useIsMobile } from '@/hooks/use-mobile';
 import { cn } from '@/lib/utils';
 
 type Props = {
@@ -37,6 +45,7 @@ const Combobox = ({
   allowFreeText = true,
   triggerClassName,
 }: Props) => {
+  const isMobile = useIsMobile();
   const [open, setOpen] = React.useState(false);
   const [search, setSearch] = React.useState('');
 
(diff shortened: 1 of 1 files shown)


components/Combobox/Combobox.tsx
```diff
@@ -53,51 +62,77 @@ const Combobox = ({
...
+        >
+          <DialogTitle className="sr-only">{placeholder}</DialogTitle>
+          <DialogDescription className="sr-only">
+            {placeholder}
+          </DialogDescription>
+          {list}
+        </DialogContent>
+      </Dialog>
+    );
+  }
```
naeemba, line 123: This mobile branch hand-rolls a `Dialog` + `DialogContent` (`showCloseButton={false}`, `top-1/N translate-y-0 overflow-hidden p-0`) wrapping sr-only `DialogTitle`/`DialogDescription` around a `Command`. That is exactly what the existing `CommandDialog` in `components/ui/command.tsx` already does — see its canonical use in `components/CommandPalette/CommandPalette.tsx`, which passes `open`/`onOpenChange`/`title`/`description` and renders `CommandInput`/`CommandList` as children. Reusing `CommandDialog` here would drop the inline `Dialog`/`DialogContent`/`DialogTitle`/`DialogDescription` imports and keep the mobile overlay visually and behaviourally consistent with the command palette. Since `Combobox` already controls `open` via `setOpen`, the trigger button can stay a sibling with `onClick={() => setOpen(true)}` instead of `DialogTrigger`, and the shared `list` would become just the `CommandInput` + `CommandList` (no inner `Command`, since `CommandDialog` supplies one). Note the styling also silently diverges (`top-1/4` + `gap-0` here vs `top-1/3` in `CommandDialog`).

components/Combobox/Combobox.tsx
```diff
@@ -53,51 +62,77 @@ const Combobox = ({
...
+      <Dialog open={open} onOpenChange={handleOpenChange}>
+        <DialogTrigger render={trigger} />
+        <DialogContent
+          showCloseButton={false}
+          className="top-1/4 translate-y-0 gap-0 overflow-hidden p-0"
+        >
+          <DialogTitle className="sr-only">{placeholder}</DialogTitle>
+          <DialogDescription className="sr-only">
+            {placeholder}
+          </DialogDescription>
```
naeemba: nit: `DialogTitle` and `DialogDescription` are both set to the same `{placeholder}` string, so a screen reader announces the placeholder twice (e.g. "Select… Select…"). Consider a distinct description (or reuse `CommandDialog`, whose defaults already provide sensible sr-only title/description text).

components/Combobox/Combobox.tsx
naeemba: Done in 8ec10ef. Replaced the hand-rolled Dialog/DialogContent/DialogTitle/DialogDescription wrapper with the shared CommandDialog: it controls open/onOpenChange/title/description and supplies its own Command, so the inner Command is gone from the shared list (the desktop popover now wraps it instead) and the styling/behaviour matches the command palette. The trigger stays a sibling driven by setOpen rather than DialogTrigger.

components/Combobox/Combobox.tsx
naeemba: Fixed in 8ec10ef. Switching to CommandDialog resolves this — its sr-only title and description are now distinct ("Select…" for the title and "Search the list or type a new value" for the description) instead of announcing the placeholder twice.

PR #47, opened 2026-06-28
fix(amounts): consistent comma thousands separators

naeemba:
## Problem

Amount formatting was inconsistent between signs: positive amounts rendered raw with no thousands separators, while negative amounts were force-formatted to three decimals with commas — producing mismatches like `Kirt 1997.5` vs `Kirt (1,997.500)`. Amount input fields had no grouping at all, making large numbers easy to mistype.

## Changes

- **`utils/formatAmount.tsx`** — positive and negative amounts now both get comma thousands separators and both preserve ledger's original decimal precision (no more forced `.000` on negatives).
- **`features/transactions/AmountInput.tsx`** (new) — a controlled amount field for the transaction form's posting rows. It **displays** comma-grouped digits but **emits the raw, un-grouped string** to the form, so `Number(...)` balance maths and the submitted draft are untouched. The caret is restored after each reformat, so inserting a digit that shifts a comma doesn't jump the cursor to the end.

The cleaning, grouping, and caret-mapping logic is extracted into pure exported functions, covered by unit tests (the project's vitest runs in a `node` environment with no DOM).

## Testing

- `pnpm exec vitest run features/transactions utils` — 68 tests pass (16 new in `AmountInput.test.ts`, plus updated `formatAmount.test.tsx`).
- `tsc --noEmit` and eslint clean (enforced by pre-commit hooks).

> [!NOTE]
> The live DOM caret restoration (`setSelectionRange` via a ref forwarded through the base-ui `Input`) is not exercised by the node tests — the caret math is tested, but the ref plumbing isn't. Worst case if that ref didn't forward: typing still yields correct values, the caret would just jump to the end. Worth a quick manual check in the new-transaction form.

### DIFF
--- a/features/transactions/AmountInput.tsx
+++ b/features/transactions/AmountInput.tsx
@@ -1,65 +1,13 @@
 'use client';
 
 import * as React from 'react';
+import {
+  caretAfterFormat,
+  cleanAmountInput,
+  countSignificant,
+  groupAmountInput,
+} from './AmountInput.util';
 import { Input } from '@/components/ui/input';
-
-// Strips a free-text amount field back to a raw, parseable number string:
-// a single optional leading minus, the digits, and at most one decimal point.
-// This is the value that gets stored and submitted, so it must never contain
-// the grouping commas the user sees while typing.
-export const cleanAmountInput = (input: string): string => {
-  const digitsAndDots = input.replace(/[^0-9.]/g, '');
-  const firstDot = digitsAndDots.indexOf('.');
-  const single =
-    firstDot === -1
-      ? digitsAndDots
-      : digitsAndDots.slice(0, firstDot + 1) +
-        digitsAndDots.slice(firstDot + 1).replace(/\./g, '');
-  // A minus is only meaningful at the very start of the field.
-  const negative = input.trimStart().startsWith('-');
-  return (negative ? '-' : '') + single;
-};
-
-// Renders a raw number string with comma thousands separators for display in
-// the input. Preserves a trailing dot and the exact decimals the user typed so
-// editing stays lossless; tolerates already-grouped input defensively.
-export const groupAmountInput = (raw: string): string => {
-  const clean = cleanAmountInput(raw);
-  if (clean === '') return '';
-  const negative = clean.startsWith('-');
-  const body = negative ? clean.slice(1) : clean;
-  const dot = body.indexOf('.');
-  const intPart = dot === -1 ? body : body.slice(0, dot);
-  const decPart = dot === -1 ? undefined : body.slice(dot + 1);
-  const groupedInt = intPart.replace(/\B(?=(\d{3})+(?!\d))/g, ',');
-  const out = decPart === undefined ? groupedInt : `${groupedInt}.${decPart}`;
-  return (negative ? '-' : '') + out;
-};
-
-// Maps a caret position expressed as "after N significant characters" onto the
-// grouped string, where commas are the only non-significant characters. Keeps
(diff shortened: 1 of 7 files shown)


features/transactions/AmountInput.tsx
```diff
@@ -0,0 +1,111 @@
...
+// editing stays lossless; tolerates already-grouped input defensively.
+export const groupAmountInput = (raw: string): string => {
+  const clean = cleanAmountInput(raw);
+  if (clean === '') return '';
+  const negative = clean.startsWith('-');
+  const body = negative ? clean.slice(1) : clean;
+  const dot = body.indexOf('.');
+  const intPart = dot === -1 ? body : body.slice(0, dot);
+  const decPart = dot === -1 ? undefined : body.slice(dot + 1);
+  const groupedInt = intPart.replace(/\B(?=(\d{3})+(?!\d))/g, ',');
```
naeemba: The comma-grouping core is now duplicated. This `\B(?=(\d{3})+(?!\d))` regex (and the int/dec split around it) is the exact same logic as `groupThousands` introduced in `utils/formatAmount.tsx` in this same PR. `groupAmountInput` is essentially `groupThousands` plus sign/empty/trailing-dot handling.

Export `groupThousands` from `utils/formatAmount.tsx` (or lift it into a shared util) and have `groupAmountInput` call it on the integer body, so the grouping rule lives in one place. As-is, the two can drift (e.g. if grouping ever changes to a locale-aware separator).

features/transactions/AmountInput.tsx
```diff
@@ -0,0 +1,111 @@
+'use client';
+
+import * as React from 'react';
+import { Input } from '@/components/ui/input';
+
+// Strips a free-text amount field back to a raw, parseable number string:
+// a single optional leading minus, the digits, and at most one decimal point.
+// This is the value that gets stored and submitted, so it must never contain
+// the grouping commas the user sees while typing.
```
naeemba: nit: The pure, DOM-free helpers (`cleanAmountInput`, `groupAmountInput`, `caretAfterFormat`, plus `countSignificant`/`SIGNIFICANT`) are defined inline in the component `.tsx` and then imported by `AmountInput.test.ts` purely for the unit tests. The established pattern in this repo is to colocate feature-specific pure logic in a `*.util(s).ts` file tested separately — see `features/dashboard/Dashboard.utils.ts` + `Dashboard.utils.test.ts`, `features/reconcile/Reconcile.utils.ts`, `features/accounts/Accounts.utils.ts`. Moving these three exports into `features/transactions/AmountInput.util.ts` would match that convention, keep the component file presentation-only, and let the test import from the util rather than reaching into the component module.

features/transactions/AmountInput.tsx
naeemba: Done in 34dbaa3. Exported `groupThousands` from `utils/formatAmount.tsx` and `groupAmountInput` now calls it for the integer body, so the comma-grouping rule lives in one place and the two can no longer drift.

features/transactions/AmountInput.tsx
naeemba: Done in 34dbaa3. Moved `cleanAmountInput`, `groupAmountInput`, `caretAfterFormat` and `countSignificant` into `features/transactions/AmountInput.util.ts`, matching the `Dashboard.utils.ts` / `Reconcile.utils.ts` / `Accounts.utils.ts` convention. The component is now presentation-only and the test (renamed to `AmountInput.util.test.ts`) imports from the util.

PR #50, opened 2026-06-28
feat(transactions): raw-text entry lens (layered entry phase 2)

naeemba:
## Phase 2 — Raw Lens

Adds a **Raw** tab to the layered transaction-entry shell where the transaction is read and edited as ledger text — the power-user fast path and in-app hand-editing. Second phase of the layered-entry roadmap (Phase 1 = #49).

Plan: `docs/superpowers/plans/2026-06-29-layered-transaction-entry-phase2-raw-lens.md`

### What's in it
- **`parsedBlockToDraft`** — pure mapping from the parser's `ParsedBlock` onto the canonical `DraftState`. Preserves the prior `uid` when the edited text omits the `; :uid:` line, so hand-editing raw text never drops the identity the edit-mode concurrency guard relies on.
- **`RawLens`** — controlled `<textarea>` seeded from `formatTransaction(draft)` on mount; each keystroke re-parses via the existing `parseBlock`, dispatching `replaceAll` only on a successful parse and showing an inline error otherwise (invalid text never clobbers the shared draft).
- **Shell wiring** — registers the always-enabled Raw tab and gates submit while the raw text is unparseable.
- **Carried cleanup** — homes the `SubmitAction` type in `features/transactions/actions` (was re-declared locally in `TransactionEntry.tsx`).

### Design note
Lenses are conditionally mounted, so sync is one-directional within RawLens's lifetime — seed-on-mount, push-up-on-valid-parse — which sidesteps cursor-jump / partial-parse thrash without a draft→text effect.

### Testing
- `parsedBlockToDraft` fully unit-tested incl. `formatTransaction → parseBlock → parsedBlockToDraft` round-trips and uid round-trip
- Static smoke tests for `RawLens` and the new Raw tab
- `pnpm type-check`, `pnpm lint`, and full suite (694 tests) green

Interactive checks (live Form↔Raw round-trip while typing, parse-error submit-gating, uid-preserving edit save) are the manual step — they need a running app with DB + auth + a journal.


features/transactions/entry/RawLens.tsx
```diff
@@ -0,0 +1,62 @@
...
+  };
+
+  return (
+    <div className="flex flex-col gap-2">
+      <textarea
+        value={text}
+        onChange={(e) => onChange(e.target.value)}
+        spellCheck={false}
+        rows={10}
+        className="w-full resize-y rounded-md border border-input bg-background px-3 py-2 font-mono text-sm leading-relaxed focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring"
```
naeemba: nit: this hand-rolls a bare `<textarea>` with its own border/focus/rounded classes, duplicating styling the shared `@/components/ui/textarea` component already provides. `FormLens` uses the shared `<Textarea>`. Since that component spreads props and merges `className`, you could reuse it here (passing `font-mono resize-y leading-relaxed` via `className` plus `rows`/`spellCheck`/`aria-label`) to stay consistent with the sibling lens and avoid re-spelling the focus-ring styles.

features/transactions/entry/RawLens.tsx
```diff
@@ -0,0 +1,62 @@
...
+  // shell-level parse error left over from a previous Raw editing session.
+  useEffect(() => {
+    onError?.(null);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  const onChange = (value: string) => {
+    setText(value);
+    const block = parseBlock(value);
+    if (!block || block.postings.length === 0) {
```
naeemba, line 40: nit: `parseBlock` silently drops any non-empty posting line it cannot parse (a line that isn't a uid, comment, or valid posting is skipped without error). So while hand-editing here, a typo on one posting line still yields `block.postings.length >= 1` for the other line(s) — RawLens treats the edit as valid, shows no parse error, leaves submit enabled, and dispatches a draft with the malformed posting silently dropped (only visible after switching to Form). This is inherent to the existing parser contract rather than introduced here, but for the hand-edit fast path it's a quiet data-loss footgun worth a follow-up (e.g. flag a line that's indented/non-comment yet unparseable).

PR #51, opened 2026-06-29
feat(transactions): layered entry Phase 3 — type engine (model extension + five adapters)

naeemba:
## Phase 3 — Type Engine (model extension + five adapters)

Third phase of the layered transaction-entry feature ([roadmap](docs/superpowers/plans/2026-06-29-layered-transaction-entry-roadmap.md)). Pure-logic phase — **no UI**; Phase 4 will render this engine as the "Types" tab.

Phase 3 as scoped in the roadmap assumed the Exchange (`@@` total-cost) and Fix-balance (`=` assertion) adapters could compile into the draft model — but the posting model was uniformly `{ account, amount, currency }` and could represent neither. Per decision, this PR extends the model **and** ships all five adapters together.

### 1. Posting-model extension (additive)
Optional `cost` (`@@` total-cost) and `assertion` (`=` balance-assertion) threaded through every layer so all three lenses round-trip losslessly:
- Parser `parsePostingLine` — reads `… AMT CUR @@ AMT CUR` and bare `… = AMT CUR`
- Draft model + `serializeDraftJson` + `parsedBlockToDraft` — carry-through (spread-when-present)
- Zod `transactionDraftSchema` + `formatPosting` — validate/emit both forms
- `computeBalance` — assertion-only postings excluded from blank-count + sum; cost postings balance on the cost currency

Plain postings parse / validate / format / balance **byte-identically** — the extension is purely additive. The server save path is untouched (it only calls `transactionDraftSchema` + `formatTransaction`).

### 2. Type engine (`features/transactions/entry/types/`)
- `accountRole.ts` — `classifyAccount` over the standard five roots, graceful `unknown` fallback
- `adapter.ts` — `TransactionTypeAdapter<F>` interface (`compile`/`detect`) + header/amount helpers
- Five adapters: **Expense, Income, Transfer, Exchange, Fix balance**, each with a `detect(compile(fields)) ≈ fields` round-trip test
- `registry.ts` — ordered adapters + `detectType` dispatcher (first match wins; `null` → fall back to Form)

### Testing
- Full suite **748 tests pass**; type-check + lint clean
- **End-to-end against the real `ledger` CLI:** compiled Expense + Exchange + Fix-balance through the actual adapters → `serializeDraftJson` → `transactionDraftSchema` → `formatTransaction`; `ledger bal` ran clean, the fix-balance assertion held (account resolved to target, `Equity:Adjustments` absorbed the difference), and the exchange `@@` balanced on the cost currency

### Deferred
Interactive Raw-lens UI save of an Exchange / Fix-balance (needs a running app + auth, as in Phases 1–2). A few cosmetic Minors (benign NaN-sign guards, an unreachable guard in `exchange.detect`) are noted to sweep opportunistically in Phase 4.

### DIFF
--- a/features/transactions/entry/balance.ts
+++ b/features/transactions/entry/balance.ts
@@ -19,8 +19,11 @@ export const computeBalance = (postings: DraftPosting[]): Balance => {
   for (const p of active) {
     if (p.cost) {
       const cost = Number(p.cost.amount);
-      if (!Number.isFinite(cost)) return { kind: 'invalid' };
-      const sign = Number(p.amount) < 0 ? -1 : 1;
+      const amount = Number(p.amount);
+      if (!Number.isFinite(cost) || !Number.isFinite(amount)) {
+        return { kind: 'invalid' };
+      }
+      const sign = amount < 0 ? -1 : 1;
       byCurrency.set(
         p.cost.currency,
         (byCurrency.get(p.cost.currency) ?? 0) + sign * cost
--- a/features/transactions/entry/balance.ts
+++ b/features/transactions/entry/balance.ts
@@ -8,14 +8,28 @@ export type Balance =
   | { kind: 'unbalanced'; issues: [string, number][] };
 
 export const computeBalance = (postings: DraftPosting[]): Balance => {
-  const blanks = postings.filter((p) => p.amount.trim() === '').length;
+  // Assertion-only postings check a balance; they don't participate in balancing.
+  const active = postings.filter(
+    (p) => !(p.assertion && p.amount.trim() === '')
+  );
+  const blanks = active.filter((p) => p.amount.trim() === '').length;
   if (blanks > 1) return { kind: 'too-many-blanks' };
   if (blanks === 1) return { kind: 'auto-balance' };
   const byCurrency = new Map<string, number>();
-  for (const p of postings) {
-    const value = Number(p.amount);
-    if (!Number.isFinite(value)) return { kind: 'invalid' };
-    byCurrency.set(p.currency, (byCurrency.get(p.currency) ?? 0) + value);
+  for (const p of active) {
+    if (p.cost) {
+      const cost = Number(p.cost.amount);
+      if (!Number.isFinite(cost)) return { kind: 'invalid' };
+      const sign = Number(p.amount) < 0 ? -1 : 1;
+      byCurrency.set(
+        p.cost.currency,
+        (byCurrency.get(p.cost.currency) ?? 0) + sign * cost
+      );
+    } else {
+      const value = Number(p.amount);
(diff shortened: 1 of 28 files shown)


features/transactions/entry/types/exchange.ts
```diff
@@ -0,0 +1,81 @@
...
+          currency: f.gaveCurrency,
+        },
+      ]
+    ),
+  detect: (draft): ExchangeFields | null => {
+    if (draft.postings.length !== 2) return null;
+    if (draft.postings.some((p) => p.assertion)) return null;
+    const got = draft.postings.find((p) => p.cost);
+    const gave = draft.postings.find((p) => !p.cost);
+    if (!got || !gave || got === gave || !got.cost) return null;
```
naeemba: nit: `!got.cost` is unreachable here. `got` is selected via `find((p) => p.cost)`, so `got.cost` is always truthy on this branch. The guard is harmless but dead; it only exists to satisfy the narrowing for line 68/73. Consider replacing it with an explicit non-null assertion or restructuring so the intent reads clearly.

features/transactions/entry/draftReducer.ts
```diff
@@ -1,7 +1,12 @@
+export type DraftCost = { amount: string; currency: string };
+export type DraftAssertion = { amount: string; currency: string };
```
naeemba: nit: `DraftCost` and `DraftAssertion` are byte-identical (`{ amount: string; currency: string }`), and the same shape is re-spelled inline in `ParsedPosting` (lib/journal/parser.ts) and as `annotationSchema` in lib/transactions/schema.ts. Consider a single shared `Annotation` type (or deriving the parser/draft types from `z.infer` of the annotation schema) so the shape lives in one place.

features/transactions/entry/balance.ts
```diff
@@ -8,14 +8,28 @@ export type Balance =
...
   const byCurrency = new Map<string, number>();
-  for (const p of postings) {
-    const value = Number(p.amount);
-    if (!Number.isFinite(value)) return { kind: 'invalid' };
-    byCurrency.set(p.currency, (byCurrency.get(p.currency) ?? 0) + value);
+  for (const p of active) {
+    if (p.cost) {
+      const cost = Number(p.cost.amount);
+      if (!Number.isFinite(cost)) return { kind: 'invalid' };
+      const sign = Number(p.amount) < 0 ? -1 : 1;
```
naeemba: nit: the cost-sign derivation `Number(p.amount) < 0 ? -1 : 1` treats an empty/NaN amount as positive (`Number('') === 0`). Compile never emits a cost posting with a blank amount so this is benign today, but a defensive `Number.isFinite` guard (matching the one a few lines up) would make the intent explicit. Same pattern is duplicated in lib/transactions/schema.ts superRefine.

PR #52, opened 2026-06-29
feat(transactions): Phase 4 — the Types tab (five type forms)

naeemba:
## Phase 4 — Layered Transaction Entry: the "Types" tab

The user-facing front end for Phase 3's pure type engine. Adds a **Types** tab to the transaction-entry shell — a row of type chips plus a tailored mini-form per type — all editing the one shared `DraftState` so the existing Form and Raw tabs always reflect Types edits.

Spec: `docs/superpowers/specs/2026-06-29-layered-transaction-entry-phase4-type-forms-design.md`
Plan: `docs/superpowers/plans/2026-06-29-layered-transaction-entry-phase4-type-forms.md`

### What's included
- **Shared primitives** (`typeForms/fields.tsx`): `Field`, `SectionLabel`, role-filtered `AccountField` (extracted from `FormLens`); `HeaderFieldsEditor`; `isEmptyDraft` predicate.
- **Five mini-forms** — Expense, Income, Transfer, Exchange, Fix-balance. Each owns its typed adapter and dispatches `replaceAll(compile(fields, ctx))` on every edit; header (date/payee/status/note/uid) is preserved across type-switches via `{ ...emptyFields(ctx), ...headerOf(draft) }`.
- **`TypeLens`** — chip selector with detect-driven state: empty draft → "Pick a type"; recognized shape → that form (seeded via `detect`); unrecognized → greyed chips + "use Form or Raw" notice.
- **Shell wiring** — tabs ordered `Types · Form · Raw`; Types is the default in create mode, and in edit mode when `detectType(draft)` matches (else Form). `canSubmit`/save flows unchanged.
- **Fix-balance live preview** — `getAccountBalance` server action backing a debounced, stale-guarded current → target → implied-adjustment display (display-only; the compiled draft stays assertion + blank `Equity:Adjustments`).

### Design decisions
- Account fields accept the **full account path** as free text (no auto-prefixing of role roots).
- Tab **ordering/persistence is out of scope** (Phase 5); the order is hardcoded here.
- A transaction's type stays **inferred from posting shape, never stored**.

### Security
An automated review flagged argv flag-smuggling in `getAccountBalance` (user-supplied account/currency into a `ledger` exec). Hardened **before** it was wired into the live UI: `isSafeLedgerArg` rejects empty/leading-dash inputs and a `--` positional separator stops flag injection (`execFile`, no shell).

### Testing
- Full suite green: **776 tests / 152 files**; type-check and lint clean.
- Per-form static smoke tests; `TypeLens` state-model tests (3 states); `isEmptyDraft`, `optionsForRoles`, and `extractAccountBalance`/`isSafeLedgerArg` unit tests; shell tests for Types-default (create) and Form-fallback (edit, unrecognized draft); TabBar a11y assertions.

### Not in this PR
- Manual in-app pass on a live `pnpm dev` instance (needs auth/DB/Garage) — to be done by the reviewer.
- A few cosmetic/by-design polish items (chip-switch discards in-progress body **by design**; redundant fallback import; empty-currency label; a `NaN` guard) deferred as Phase-5-adjacent backlog.

### Out of scope (whole feature)
Storing the type in the journal; account-role aliases; types beyond the five; natural-language quick-add.

### DIFF
--- a/features/transactions/entry/typeForms/ExpenseForm.tsx
+++ b/features/transactions/entry/typeForms/ExpenseForm.tsx
@@ -1,0 +1,89 @@
+// features/transactions/entry/typeForms/ExpenseForm.tsx
+'use client';
+
+import React, { useMemo, useState } from 'react';
+import AmountInput from '../../AmountInput';
+import type { DraftAction, DraftState } from '../draftReducer';
+import { headerOf } from '../types/adapter';
+import { expenseAdapter, type ExpenseFields } from '../types/expense';
+import { HeaderFieldsEditor } from './HeaderFields';
+import { Field, SectionLabel, AccountField } from './fields';
+import { Input } from '@/components/ui/input';
+
+export type TypeFormProps = {
+  draft: DraftState;
+  dispatch: (a: DraftAction) => void;
+  accounts: string[];
+  payees: string[];
+  defaultCurrency: string;
+};
+
+export function ExpenseForm({
+  draft,
+  dispatch,
+  accounts,
+  payees,
+  defaultCurrency,
+}: TypeFormProps): React.JSX.Element {
+  const ctx = useMemo(() => ({ defaultCurrency }), [defaultCurrency]);
+  const [fields, setFields] = useState<ExpenseFields>(
+    () =>
+      expenseAdapter.detect(draft) ?? {
+        ...expenseAdapter.emptyFields(ctx),
+        ...headerOf(draft),
+      }
+  );
+
+  const update = (next: ExpenseFields) => {
+    setFields(next);
+    dispatch({ type: 'replaceAll', state: expenseAdapter.compile(next, ctx) });
+  };
+
+  return (
+    <div className="grid gap-6 md:grid-cols-[minmax(240px,1fr)_1.6fr] md:gap-8 lg:grid-cols-[minmax(280px,360px)_1fr]">
+      <HeaderFieldsEditor
+        header={fields}
+        payees={payees}
+        onChange={(patch) => update({ ...fields, ...patch })}
+      />
+
+      <section className="flex flex-col gap-5 md:border-l md:border-border md:pl-8">
+        <SectionLabel>Expense</SectionLabel>
+
+        <Field label="Amount">
+          <div className="flex gap-2">
+            <AmountInput
+              value={fields.amount}
+              onChange={(amount) => update({ ...fields, amount })}
+              placeholder="Amount"
+              className="flex-1 text-right tabular-nums"
+            />
+            <Input
+              type="text"
+              value={fields.currency}
+              onChange={(e) => update({ ...fields, currency: e.target.value })}
+              placeholder="Currency"
+              className="w-24"
+            />
+          </div>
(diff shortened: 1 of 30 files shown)


features/transactions/entry/typeForms/ExpenseForm.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { Field, SectionLabel, AccountField } from './fields';
+import { Input } from '@/components/ui/input';
+
+export type TypeFormProps = {
+  draft: DraftState;
+  dispatch: (a: DraftAction) => void;
+  accounts: string[];
+  payees: string[];
+  defaultCurrency: string;
+};
```
naeemba: nit: The shared `TypeFormProps` contract lives inside the concrete `ExpenseForm.tsx` and is re-imported by every other form plus `TypeLens`. That makes `ExpenseForm` an implicit "types module" for the whole `typeForms/` directory, so e.g. `TransferForm` depends on `ExpenseForm` purely for a type. Consider homing it next to the adapter contract (e.g. `types/adapter.ts` or a small `typeForms/props.ts`) so the forms are siblings rather than one depending on another.

features/transactions/entry/typeForms/FixBalanceForm.tsx
```diff
@@ -0,0 +1,118 @@
...
+    const account = fields.account.trim();
+    const id = ++reqId.current;
+    let t: ReturnType<typeof setTimeout> | undefined;
+    if (account) {
+      t = setTimeout(() => {
+        void getAccountBalance(account, fields.targetCurrency).then((bal) => {
+          if (id === reqId.current) setCurrent(bal);
+        });
+      }, 300);
+    } else {
```
naeemba: nit: The empty-account branch schedules `setTimeout(() => setCurrent(null), 0)` behind the same request-id guard just to clear the preview. A direct `setCurrent(null); return;` (as the plan's original version did) is equivalent and clearer — when `account` is empty there is no in-flight request to race against, so the 0ms timer + id check adds indirection without buying anything.

features/transactions/entry/typeForms/ExpenseForm.tsx
naeemba: Fixed in d016759. Moved `TypeFormProps` into a new `typeForms/props.ts` and repointed all six importers (the five forms + `TypeLens`) at it, so the forms are siblings instead of depending on `ExpenseForm` for a type.

features/transactions/entry/typeForms/FixBalanceForm.tsx
naeemba: Addressed in d016759, with a caveat: the bare `setCurrent(null); return;` the plan used trips the `react-hooks/set-state-in-effect` lint rule (synchronous setState in an effect body), so a fully-direct clear fails pre-commit. Instead of two `setTimeout` blocks I collapsed it into one with a conditional delay (`account ? 300 : 0`) and a single request-id guard — removing the duplication while keeping the deferral the linter requires. Added a comment explaining why.

PR #53, opened 2026-06-29
feat(transactions): Phase 5 — configurable entry tab order & default

naeemba:
Final phase of the layered transaction-entry feature: a per-user preference for the order of the three entry tabs (**Types · Form · Raw**), where the first tab is the one the add/edit screen opens on. Ships the default `Types · Form · Raw`; reorderable in Settings.

## What's in it

Bottom-up, one concern per commit:

1. **Pure core** — `lib/transactions/entryTabs.ts`: `TabId`/`TAB_IDS`/`DEFAULT_TAB_ORDER` + `normalizeTabOrder`/`parseEntryTabOrder`/`serializeEntryTabOrder` + `entryTabOrderSchema`. `normalizeTabOrder` keeps valid ids (deduped, first-wins) then appends any missing canonical tab, so the full set is always returned.
2. **Storage** — nullable `entryTabOrder` text column on `userSetting` (migration `0011`), stored as a comma-joined string.
3. **Repository + service** — `upsertEntryTabOrder` (mirrors `upsertBaseCurrency`, no-clobber) and `saveEntryTabOrder` (serializes through the normalizer).
4. **Reader** — cached, `server-only` `getEntryTabOrder()`; degrades to the default order for anonymous users and on DB error. Deliberately no cookie/session-override layer (unlike base currency) — tab order is a pure stored preference.
5. **Action** — `setEntryTabOrderAction`: `requireUser` → rate-limit → `entryTabOrderSchema.safeParse` → save → `revalidatePath`. Unknown ids are rejected before any write.
6. **Shell** — `TransactionEntry` gains an optional `tabOrder` prop that drives the tab order; the default active tab becomes the first ordered tab, while the edit-mode fallback to **Form** (when the draft's type can't be detected) is preserved.
7. **Wiring** — the New/Edit server components read the order and pass it down.
8. **Settings UI** — `EntryTabOrderForm` (up/down reorder, "Default" marker on the top tab, dirty-gated Save) in a new "Transaction entry tabs" card.

## Invariants held

- All three tabs are always present regardless of stored order — enforced on both read (`parseEntryTabOrder`) and write (`serializeEntryTabOrder`), so a tab-hiding order can neither be persisted nor survive a read.
- Users with no preference get exactly `Types · Form · Raw`; a null column parses to the default. Migration is additive and backward-compatible.
- Transaction type stays inferred from posting shape — never stored. No journal/ledger-writing path or submit flow changed; the shell change touches only the tab list and the default-active-tab seed.

## Tests

Full suite green: **157 files / 799 tests** (+24 for this phase), type-check and lint clean. Coverage spans the serialize→parse round-trip, normalization healing of partial/unknown stored values, schema rejection, all four reader degradation branches, the order-driven shell (custom order + preserved default + edit-mode fallback), and the Settings card.

Manual `pnpm dev` browser pass deferred (needs live app + auth/DB), consistent with earlier phases.

Follow-up (non-blocking): consolidate the duplicated `TAB_LABELS` map into a single export from `entryTabs.ts`.

### DIFF
--- a/features/settings/EntryTabOrderForm.tsx
+++ b/features/settings/EntryTabOrderForm.tsx
@@ -4,13 +4,7 @@ import { useState, useTransition } from 'react';
 import { toast } from 'sonner';
 import { Button } from '@/components/ui/button';
 import { setEntryTabOrderAction } from '@/features/settings/actions';
-import { type TabId } from '@/lib/transactions/entryTabs';
-
-const LABELS: Record<TabId, string> = {
-  types: 'Types',
-  form: 'Form',
-  raw: 'Raw',
-};
+import { type TabId, TAB_LABELS } from '@/lib/transactions/entryTabs';
 
 type Props = { initial: TabId[] };
 
@@ -44,7 +38,7 @@ const EntryTabOrderForm = ({ initial }: Props) => {
             className="flex items-center justify-between gap-3 rounded-md border border-border px-3 py-2"
           >
             <span className="text-sm font-medium">
-              {LABELS[id]}
+              {TAB_LABELS[id]}
               {i === 0 && (
                 <span className="ml-2 text-xs text-muted-foreground">
                   Default
@@ -56,7 +50,7 @@ const EntryTabOrderForm = ({ initial }: Props) => {
                 type="button"
                 variant="outline"
                 size="sm"
-                aria-label={`Move ${LABELS[id]} up`}
+                aria-label={`Move ${TAB_LABELS[id]} up`}
                 disabled={i === 0}
                 onClick={() => setOrder((o) => move(o, i, i - 1))}
               >
@@ -66,7 +60,7 @@ const EntryTabOrderForm = ({ initial }: Props) => {
                 type="button"
                 variant="outline"
                 size="sm"
-                aria-label={`Move ${LABELS[id]} down`}
+                aria-label={`Move ${TAB_LABELS[id]} down`}
                 disabled={i === order.length - 1}
                 onClick={() => setOrder((o) => move(o, i, i + 1))}
               >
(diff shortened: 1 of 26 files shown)


features/settings/EntryTabOrderForm.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { toast } from 'sonner';
+import { Button } from '@/components/ui/button';
+import { setEntryTabOrderAction } from '@/features/settings/actions';
+import { type TabId } from '@/lib/transactions/entryTabs';
+
+const LABELS: Record<TabId, string> = {
+  types: 'Types',
+  form: 'Form',
+  raw: 'Raw',
+};
```
naeemba: This `LABELS` tab-id→label map is an exact copy of `TAB_LABELS` in `features/transactions/entry/TransactionEntry.tsx` (lines 59-63). Two sources of truth for the same canonical tab names means a relabel (e.g. "Types"→"Type forms") silently drifts between the Settings reorder UI and the entry shell. Since `TabId`/`TAB_IDS`/`DEFAULT_TAB_ORDER` already live in the pure `lib/transactions/entryTabs.ts`, export a single `TAB_LABELS: Record<TabId, string>` from there and import it in both components. (The PR description already notes this as a follow-up — worth folding in now while both call sites are in the diff.)

naeemba: Done in fdbcf08. Exported a single `TAB_LABELS: Record<TabId, string>` from `lib/transactions/entryTabs.ts` and imported it in both `TransactionEntry` and `EntryTabOrderForm`, removing the two duplicated maps so the labels can no longer drift.

PR #55, opened 2026-06-29
feat(transactions): codemirror raw entry with autocomplete, highlighting, formatter, and smart indent

naeemba:
## Summary

Replaces the raw transaction-entry `<textarea>` with a CodeMirror 6 editor, giving the Raw tab IDE-grade ergonomics while keeping the existing parse → `replaceAll` dispatch flow untouched.

What it adds:

- **Context-aware autocomplete at the caret** — payees on the header line, accounts in a posting's account region, commodities after an amount. Fed by the account/payee/currency lists already passed to `TransactionEntry` (no new server calls).
- **Ledger syntax highlighting** — date, status marker, account, amount, and comments.
- **Amount formatter** — aligns amounts to the shared column (reuses `formatPosting` / `ACCOUNT_COLUMN`), runs on blur, via a Format button, and on ⇧⌥F. Idempotent and defensive: returns the text unchanged when there's no valid header, and passes comments / uid / blank / unparsable lines through verbatim.
- **Smart indentation** — Enter starts the next posting line pre-indented (4 spaces); Tab accepts an open completion else inserts an indent; ⇧Tab dedents.

## Architecture

All testable logic lives in pure, `node`-friendly modules; CodeMirror is a thin adapter (the repo's Vitest env is `node` with no DOM harness):

- `lib/ledger/format.ts` — `formatLedgerText` (idempotent formatter)
- `lib/ledger/completionContext.ts` — `completionAt` (pure caret-context resolver)
- `lib/ledger/completions.ts` — CodeMirror `CompletionSource` adapter over `completionAt`
- `lib/ledger/highlight.ts` — ledger `StreamLanguage` tokenizer
- `features/transactions/entry/rawLensLogic.ts` — `applyRawText`, the extracted parse/validate/dispatch decision
- `features/transactions/entry/LedgerEditor.tsx` — the editor (autocomplete + highlight + format + keymap + SSR mount guard)

## Testing

- New unit tests for the formatter, the completion resolver, and `applyRawText`.
- Full suite green: **816 tests** across 160 files; `type-check`, `lint`, and `build` all clean.
- Deferred: live browser checks (dropdown render, highlight colors, format keystroke) need a running app with auth/DB, consistent with prior phases.

## Notes

Built behind a `mounted` guard so it renders a labelled placeholder during SSR (CodeMirror initialises client-side); the editable surface carries its `aria-label` via `EditorView.contentAttributes`.


lib/ledger/completionContext.ts
```diff
@@ -0,0 +1,64 @@
...
+    const from = lineStart + m[1].length;
+    return result(from, filterList(lists.payees, payeeToken));
+  }
+
+  // Posting line must be indented.
+  if (!/^\s/.test(upToCaret)) return null;
+
+  // An amount gap (2+ spaces or a tab after the account) splits the line into
+  // account region (before) and amount/commodity region (after).
+  const gap = upToCaret.match(/\s{2,}|\t+/g);
```
naeemba: nit: `gap` (line 49) is computed but never used — the `void gap;` on line 62 is only there to silence the unused-variable lint. The `hasAmountGap` boolean on line 50 already covers the account-vs-amount split, so this line and the `void` can be dropped.

PR #59, opened 2026-06-30
fix(crypto): stop unlock redirect loop on prefetch RSC requests

naeemba:
## Problem

After a deploy, the journal locks (the in-RAM DEK is dropped on server restart — expected) and the user sees a **storm of `/crypto/unlock?_rsc=` requests and a blank screen**. A hard refresh recovers and lands on the unlock page.

## Root cause

`CryptoGate` is mounted in the root `app/layout.tsx` and calls `redirect('/crypto/unlock')` when locked. On deploy every user becomes locked, so every protected route bounces to unlock. The sidebar/header nav `<Link>`s prefetch by default, and **the App Router retries a prefetch that receives a redirect in a tight loop** ([vercel/next.js#48438](https://github.com/vercel/next.js/issues/48438)) — hence the `_rsc` storm + white screen. A full reload works because it's a normal 307, not an RSC prefetch.

The earlier wordmark `prefetch={false}` fix (96f1c13) only covered two links on the unlock/setup pages, not the app-wide nav.

## Fix

Never `redirect()` on a prefetch. New `isPrefetchRequest` helper detects prefetches via `next-router-prefetch: 1` **or** `purpose: prefetch` (both, since [#85836](https://github.com/vercel/next.js/issues/85836) reports the first is sometimes missing). `CryptoGate` returns `null` on a prefetch and only redirects on real navigations / full loads. The data layer's `LockedError` remains the hard backstop. One fix point covers all prefetchable links without degrading prefetch perf for unlocked users.

## Tests

- New unit tests: gate redirects on real navigation, stays silent on prefetch; `isPrefetchRequest` header matrix.
- Full suite green (840 tests), lint + type-check clean.

## Verification note

Root cause is grounded in the Next.js issue tracker and the existing in-code comment that already diagnosed this class of bug; logic is unit-tested. The production storm was **not** reproduced end-to-end (needs a prod build + mid-session restart).

### DIFF
--- a/components/crypto/CryptoGate.test.ts
+++ b/components/crypto/CryptoGate.test.ts
@@ -54,6 +54,14 @@ describe('CryptoGate', () => {
     expect(redirectMock).not.toHaveBeenCalled();
   });
 
+  it('redirects an unset user to /crypto/setup on a real navigation', async () => {
+    setHeaders({ 'x-pathname': '/dashboard', 'RSC': '1' });
+    cryptoStatus.mockResolvedValue('unset');
+    const { CryptoGate } = await import('./CryptoGate');
+    await expect(CryptoGate()).rejects.toThrow('NEXT_REDIRECT');
+    expect(redirectMock).toHaveBeenCalledWith('/crypto/setup');
+  });
+
   it('does NOT redirect an unset user on a prefetch request', async () => {
     setHeaders({ 'x-pathname': '/dashboard', 'purpose': 'prefetch' });
     cryptoStatus.mockResolvedValue('unset');
--- a/components/crypto/CryptoGate.test.ts
+++ b/components/crypto/CryptoGate.test.ts
@@ -1,0 +1,72 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+const headerStore = new Map<string, string>();
+const redirectMock = vi.fn((_dest: string): never => {
+  // Mirror Next's redirect(), which throws to halt rendering.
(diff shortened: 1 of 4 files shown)


components/crypto/CryptoGate.test.ts
```diff
@@ -0,0 +1,72 @@
...
+describe('CryptoGate', () => {
+  it('redirects a locked user to /crypto/unlock on a real navigation', async () => {
+    setHeaders({ 'x-pathname': '/dashboard', 'RSC': '1' });
+    cryptoStatus.mockResolvedValue('locked');
+    const { CryptoGate } = await import('./CryptoGate');
+    await expect(CryptoGate()).rejects.toThrow('NEXT_REDIRECT');
+    expect(redirectMock).toHaveBeenCalledWith('/crypto/unlock');
+  });
+
+  it('does NOT redirect a locked user on a prefetch request (avoids the _rsc storm)', async () => {
```
naeemba, line 45: nit: the gate's redirect matrix is missing the `unset` + real-navigation case. You test `locked` → redirects to `/crypto/unlock` on a real nav, and `unset` → silent on prefetch, but never assert that an `unset` user on a real navigation redirects to `/crypto/setup`. Since this PR changes the code path that precedes both `redirect('/crypto/setup')` and `redirect('/crypto/unlock')`, a symmetric test for the setup redirect would lock in that the prefetch short-circuit doesn't accidentally swallow the setup branch too.

PR #60, opened 2026-07-01
feat(accounts): friendly accounts section for non-ledger users

naeemba:
## Why

The accounts page was legible to people who already know ledger-cli, but opaque to everyone else: an always-expanded raw colon-tree (`Assets:Bank:Checking`), no balances, no grouping, and ledger's credit/debit sign convention that shows a credit card as `-$512` and a salary as `-$5,000` — actively misleading to a normal person.

This reframes the page for non-ledger users, entirely in the presentation layer. The journal, ledger write paths, and data model are untouched; the only backend change is switching the read query from `ledger accounts` to tree-mode `ledger balance`.

## What changed

- **Two friendly buckets** instead of five raw roots: **Accounts** (Assets + Liabilities — what you have/owe) and **Categories** (Income + Expenses — where money comes from/goes). `Equity` and any non-standard root are tucked into a collapsed **Advanced** fold.
- **Friendly balances**: each row shows a magnitude + a direction arrow (↑ in your favour / ↓ against you) + colour, derived from the account's role and the *sign* of its balance — never hard-coded per role. A short chip appears only when a balance is reversed from its normal side:
  - reversed liability (a friend owes you) → `owed to you`
  - overdrawn asset → `overdrawn`
  - income reversal → `reduced`; expense rebate → `refunded`
- **Collapsible tree** (all collapsed except Expenses), with search that expands matches. Leaf rows still route to the existing account detail page.
- The colon-paths remain the source of truth — displayed kindly, never renamed.

## How balances stay correct

Tree-mode `ledger balance` (no `--flat`) emits every node — including synthetic parents — with the rolled-up `%T` already computed, so parent rows show correct totals with no JS rollup math. Empirically verified during design (including the parent-with-direct-postings case). Non-numeric/non-converted commodity amounts fail safe to an em-dash rather than rendering nonsensically.

## Tests

- Pure logic is TDD-covered: `balanceDisplay` (every role × sign branch, reversals, zero), `parseAmountParts` (sign/precision), `accountTree` (build/bucket/leaf-count, synthesised parents), `FriendlyBalance` (arrow/colour/chip/em-dash).
- Full suite green: 166 files / 850 tests, `tsc --noEmit` clean.
- **Manual `/accounts` browser check still owed** (three sections, search/collapse, arrows/chips on a real journal) — could not be run headless (needs auth + a journal).

## Deferred (documented in the spec)

- **"Ledger mode" setting** — a per-user toggle to render accounts raw (five roots, colon-paths, raw signs) for power users; this is also where the current raw tree would be preserved. Note: until it ships, zero-net-balance accounts are hidden (tree-mode balance omits them), which is intended for a balance-oriented view.
- Path visible-vs-tooltip choice, account-subtype icons (only via explicit tagging, never inferred), richer multi-currency.

Spec: `docs/superpowers/specs/2026-07-01-friendly-accounts-design.md`
Plan: `docs/superpowers/plans/2026-07-01-friendly-accounts.md`


features/accounts/AccountsView.tsx
```diff
@@ -25,9 +33,16 @@ const AccountsView = ({ accounts }: Props) => {
...
-        <Tree tree={tree} />
-      </div>
+      {buckets.map((bucket) => (
+        <BucketSection
+          key={bucket.key}
+          title={bucket.title}
+          count={countLeaves(bucket.roots)}
+          defaultOpen={bucket.key !== 'advanced'}
+        >
+          <AccountTreeView nodes={bucket.roots} forceOpen={searching} />
```
naeemba, line 44: Search doesn't reveal matches inside a collapsed `BucketSection`. `searching` is threaded into `AccountTreeView` (`forceOpen`), which expands the tree *nodes*, but the enclosing `BucketSection` open state is independent local state seeded once from `defaultOpen` and never reacts to `searching`. Since the `advanced` bucket defaults collapsed, searching for an Equity/unknown-role account (e.g. `Opening`) produces a match that stays hidden — the search silently appears to return nothing. Same happens for any bucket the user has manually collapsed. Consider forcing the section open (or lifting the open state) when `searching` is true.

features/accounts/accountTree.ts
```diff
@@ -0,0 +1,98 @@
...
+    categories: [],
+    advanced: [],
+  };
+  for (const root of roots) {
+    groups[BUCKET_OF[root.role]].push(root);
+  }
+  return order.map((key) => ({
+    key,
+    title: BUCKET_TITLES[key],
+    roots: groups[key].sort((a, b) => a.name.localeCompare(b.name)),
```
naeemba: nit: `roots` is already sorted alphabetically by `buildAccountTree` (`sortRec` sorts every level, including the roots array), so this `.sort(...)` inside `bucketRoots` is redundant. Not harmful, just dead work.

features/accounts/AccountsView.tsx
naeemba: Fixed in 2b06bbd. Added a forceOpen prop to BucketSection wired from the searching flag, so sections (including the default-collapsed Advanced bucket and any manually collapsed one) expand while a query is active and restore their prior open state when the search clears.

features/accounts/accountTree.ts
naeemba: Fixed in 492f29d. Dropped the redundant .sort in bucketRoots since buildAccountTree already sorts the roots array (and every level) and bucketing preserves that order.

PR #63, opened 2026-07-03
fix(review): A2 — carry cost/assertion through save-as-template

naeemba:
## A2 — Save-as-template silently drops `@@` cost and `=` assertion annotations

Saving a transaction row as a template dropped its `@@` cost and `=` balance-assertion annotations. `TransactionRow`'s posting shape was narrowed to `{account, amount, currency}`, and both `toTransactionRow` and `toTemplateDraft` mapped only those three fields — even though `templateDraftSchema` legally stores the annotations (it reuses `postingSchema`).

Impact: a multi-currency transaction balanced via `@@` cost hydrated into a template whose postings no longer summed to zero, so `computeBalance`/`canSubmit` left the Save buttons permanently disabled — the template was unusable for its primary purpose. Assertion-only postings degraded into blank auto-balance rows, silently changing semantics.

### Fix
- Widen the `TransactionRow` posting shape to carry optional `cost`/`assertion`, and pass them through in `toTransactionRow` and `toTemplateDraft`.
- Move `toTemplateDraft` out of the `RowActions` client component into the pure `transactionRow` module so it is unit-testable alongside `toTransactionRow`.
- Regression tests cover cost and assertion round-tripping through both mappers.

### Verification
- `npx vitest run` — 869 passed
- `npx tsc --noEmit` — clean
- `npx eslint` — clean

Tracked in `REVIEW.md` (workstream A, item A2).

### DIFF
--- a/features/transactions/transactionRow.ts
+++ b/features/transactions/transactionRow.ts
@@ -1,3 +1,4 @@
+import { carryAnnotations } from './carryAnnotations.util';
 import type { Annotation, Transaction } from '@/lib/journal/parser';
 import type { TemplateDraft } from '@/lib/templates/schema';
 
@@ -27,8 +28,7 @@ export const toTransactionRow = (t: Transaction): TransactionRow => ({
     account: p.account,
     amount: p.amount,
     currency: p.currency,
-    ...(p.cost ? { cost: p.cost } : {}),
-    ...(p.assertion ? { assertion: p.assertion } : {}),
+    ...carryAnnotations(p),
   })),
 });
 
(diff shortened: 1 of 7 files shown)


features/transactions/transactionRow.ts
```diff
@@ -20,5 +27,29 @@ export const toTransactionRow = (t: Transaction): TransactionRow => ({
...
+  payee: t.payee,
+  status: t.status,
+  note: t.note ?? undefined,
+  postings: t.postings.map((p) => ({
+    account: p.account,
+    amount: p.amount,
+    currency: p.currency,
+    ...(p.cost ? { cost: p.cost } : {}),
+    ...(p.assertion ? { assertion: p.assertion } : {}),
   })),
```
naeemba, line 53: nit: the annotation-carry spread `...(p.cost ? { cost: p.cost } : {})` / `...(p.assertion ? { assertion: p.assertion } : {})` is now duplicated three times — here, in `toTransactionRow` above, and in `entry/transactionToDraft.ts` (A1). With the A3/A4 hydration paths still to come, this idiom is worth lifting into a small shared helper (e.g. `carryAnnotations(p)` in a `*.util.ts`) so every posting mapper stays in sync. Not blocking.

PR #64, opened 2026-07-03
fix(review): A3 — carry cost/assertion through entry-form save-as-template

naeemba:
### Problem
When building the `TemplateDraft` from the live entry draft, `TransactionEntry` mapped each posting to only `{account, amount, currency}`, silently dropping the `@@` cost and `=` balance-assertion annotations that `DraftPosting` carries and `templateDraftSchema` legally accepts. A multi-currency draft balanced via `@@` cost then rehydrated into a template whose postings no longer sum to zero per currency, leaving the Save buttons disabled — the template was unusable for its primary purpose. This was the second independent stripping site (beyond RowActions / A2).

### Fix
Extracted the inline `templateDraft` construction into a pure, unit-tested `draftToTemplateDraft(draft)` helper (`features/transactions/entry/draftToTemplateDraft.ts`) that carries both annotations through the shared `carryAnnotations` idiom — the same helper the edit (A1) and row (A2) paths already use, keeping all posting mappers in sync. `draftToTemplateDraft.test.ts` pins cost/assertion round-tripping plus field trimming.

A4 (template hydration in `NewTransaction`) remains the last stripping hop in this pipeline.

### Verification
- `npx vitest run` — 872 tests green (3 new)
- `npx tsc --noEmit` — clean
- `npx eslint` on changed files — clean

Refs REVIEW.md item **A3**.

### DIFF
--- a/features/transactions/entry/draftToTemplateDraft.ts
+++ b/features/transactions/entry/draftToTemplateDraft.ts
@@ -1,0 +1,22 @@
+import { carryAnnotations } from '../carryAnnotations.util';
+import type { DraftState } from './draftReducer';
+import type { TemplateDraft } from '@/lib/templates/schema';
+
+/**
+ * Build the `TemplateDraft` saved by "Save as template" from the live entry
+ * draft. Trims text fields and carries each posting's `@@` cost and `=`
+ * balance-assertion annotations through {@link carryAnnotations} — dropping
+ * them would rehydrate a cost-balanced multi-currency template into an
+ * unbalanced, unsubmittable draft.
+ */
+export const draftToTemplateDraft = (draft: DraftState): TemplateDraft => ({
+  payee: draft.payee.trim() || '—',
+  status: draft.status,
+  note: draft.note.trim() || undefined,
+  postings: draft.postings.map((p) => ({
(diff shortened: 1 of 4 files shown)


features/transactions/entry/draftToTemplateDraft.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Build the `TemplateDraft` saved by "Save as template" from the live entry
+ * draft. Trims text fields and carries each posting's `@@` cost and `=`
+ * balance-assertion annotations through {@link carryAnnotations} — dropping
+ * them would rehydrate a cost-balanced multi-currency template into an
+ * unbalanced, unsubmittable draft.
+ */
+export const draftToTemplateDraft = (draft: DraftState): TemplateDraft => ({
+  payee: draft.payee.trim() || '—',
+  status: draft.status,
+  note: draft.note.trim() || undefined,
```
naeemba, line 15: nit: the test pins trimming + the annotation carry-through, but not the two fallback branches on this line — empty `payee` collapsing to `—` and empty `note` collapsing to `undefined`. A case with a blank payee/note would fully lock down this helper's behavior.

PR #65, opened 2026-07-03
refactor(model): transaction model consolidation — one Txn class (P1–P5)

naeemba:
## Transaction model consolidation

Collapses the entry/transaction subsystem's tangle of near-identical types and scattered converters into **one canonical `Txn` model** used everywhere. Full design + plan are in-branch under `docs/superpowers/`.

### The problem
The subsystem had **five near-identical "transaction-with-postings" container types**, the **posting shape declared four times**, and **six bespoke posting-converter functions** scattered across five files. Every converter independently had to remember to carry `@@` cost / `=` assertion annotations — forgetting is exactly the A1–A4 bug class.

### The shape now
- **One `Posting` type** in a dependency-free leaf `lib/transactions/posting.ts`; `ParsedPosting`, `DraftPosting`, and the `TransactionRow` posting all alias it (re-exports keep every importer working; no import cycle).
- **One `Txn` class** (`lib/transactions/model.ts`) is the entry-form `useReducer` state, with named constructors (`fromTransaction`, `fromParsedBlock`, `fromTemplate`), immutable updates (`withField`/`withPosting`/`addPosting`/`removePosting`), and outputs (`toWire`, `toTemplate`). The six scattered converters are gone.

### Phases (all on this branch)
| Phase | What | Notes |
|---|---|---|
| P1 | Add `Txn` + tests | wired nothing |
| P2 | `Txn` becomes the reducer state | **closes A3** (entry-form save-as-template now carries cost/assertion) |
| P3a | client-side edges → `Txn` | raw lens, RowActions; deleted `parsedBlockToDraft`, `toTemplateDraft` |
| P3b | server-side seeds → `Txn.…toWire()` | **closes A4** (template hydration carries cost/assertion); deleted `transactionToDraft` |
| P4 | collapse posting declarations | **supersedes G3** |
| P5 | drop 3 orphaned `Txn` methods | `empty`/`fromJSON`/`toSubmit` had no callers |

Closes review items **A3**, **A4**, **G3**. **Supersedes PR #64** (A3) — that branch's standalone `draftToTemplateDraft` fix is now covered here via `Txn.toTemplate`; #64 can be closed.

### Design notes
- **RSC boundary respected.** `EditTransaction`/`NewTransaction` are server components; they pass only plain `.toWire(...)` output (a plain `TxnJSON`) as `initialDraft`, never a `Txn` instance — the client rebuilds the instance in the reducer.
- **`??`→`||` date fallback** (TransactionEntry): required by the new template seed (`toWire('create')` emits `date: ''` → falls back to today). Safe on every path — edit seeds a real date, plain-new is `undefined`, and a date string is never legitimately falsy except `''` (= unset).
- **Harmless behavior-widening (informational):** the deleted edit/row converters didn't trim or apply the `payee || '—'` default; `toWire`/`toTemplate` do. This is a no-op on parser-clean data (the header regex trims payee and rejects blanks; posting fields carry no surrounding whitespace), so persisted transactions are unaffected.
- G3's status notes the four **TypeScript** posting declarations are unified; the zod `postingSchema` remains the independent runtime validator (its inferred type is already structurally identical — the P1 work dropped the DTO casts proving this). A follow-up could assert that equivalence at compile time.

### Verification
- **869 tests green**, `tsc --noEmit` clean, `eslint` clean.
- The annotation-carry path (the A1–A4 motivation) is covered at the single hub in `model.test.ts` across every constructor and output.
- Independent whole-branch review (behavior-preservation, RSC boundary, A3/A4 closure, canonical `Posting`, date-fallback safety) → ready to merge.


lib/transactions/model.ts
```diff
@@ -0,0 +1,223 @@
...
+  postings: Posting[];
+};
+
+const carry = (p: {
+  cost?: Annotation;
+  assertion?: Annotation;
+}): Pick<Posting, 'cost' | 'assertion'> => ({
+  ...(p.cost ? { cost: p.cost } : {}),
+  ...(p.assertion ? { assertion: p.assertion } : {}),
+});
```
naeemba: This `carry` helper is a byte-for-byte duplicate of the existing `carryAnnotations` in `features/transactions/carryAnnotations.util.ts`, whose own docstring states the idiom "lives here once to keep all mappers in sync." Reintroducing a private copy is exactly the drift risk that util was created to prevent, and it undercuts the consolidation goal this PR series is pursuing.

The wrinkle is layering: `carryAnnotations` currently sits under `features/`, and this new model lives under `lib/` (non-UI shared code), so `lib` importing from `features` would be an inversion. The clean fix is to lift `carryAnnotations` down into `lib/transactions/` now and re-point the feature mappers (`transactionToDraft`, `parsedBlockToDraft`, `transactionRow`) at it — which is the P3/P4 direction anyway — rather than adding a fourth copy here. At minimum, call out this copy as a deliberate temporary that P4 collapses.

lib/transactions/model.ts
```diff
@@ -0,0 +1,223 @@
...
+
+export type Posting = {
+  account: string;
+  amount: string;
+  currency: string;
+  cost?: Annotation;
+  assertion?: Annotation;
+};
+
+export type TxnStatus = 'cleared' | 'pending' | 'none';
```
naeemba: nit: `TxnStatus` re-spells the `'cleared' | 'pending' | 'none'` union that already exists three times — the parser (`Transaction['status']` / `ParsedBlock['status']`) and both zod enums (`transactionDraftSchema`, `templateDraftSchema`). Consider deriving it (`type TxnStatus = TransactionDraft['status']`) so the canonical model tracks the schema instead of holding a fourth independent copy.

lib/transactions/model.ts
naeemba: Done in 3fd0a41. Lifted carryAnnotations down into lib/transactions/ (avoiding the lib->features inversion) and re-pointed the three feature mappers (transactionRow, parsedBlockToDraft, transactionToDraft) at the new location. Removed the private carry copy from the model so there's a single canonical idiom.

lib/transactions/model.ts
naeemba: Done in 0912f9f. TxnStatus is now derived as TransactionDraft['status'] so it tracks the zod schema instead of holding a fourth independent copy.

lib/journal/service.ts
```diff
@@ -405,8 +402,8 @@ export class JournalService {
       };
     }
 
-    const text = await this.repo.readFile(tx.file);
-    const fileTxs = parseJournalFile(tx.file, text);
+    const text = await this.repo.readFile(tx.file!);
+    const fileTxs = parseJournalFile(tx.file!, text);
```
naeemba: Consolidating the parser transaction into the unified \`Transaction\` model widened its identity fields (\`file\`, \`startLine\`, \`endLine\`, \`rawBlock\`, \`fingerprint\`) to optional, since a scratch draft has none. The cost is that every consumer that *knows* it holds a parsed transaction must now assert non-null: this method alone adds `!` in ~10 spots (\`tx.file!\`, \`current.startLine!\`, \`current.endLine!\`), and it feeds the write path (\`readFile(tx.file!)\`, \`writeFileAtomic(tx.file!, ...)\`). On \`main\` the parser `Transaction` guaranteed these at compile time; the PR trades that guarantee for `!`.

It is safe at runtime here — \`repo.find\` -> \`parseJournal\` always populates them — but the compiler no longer protects these file-write paths, and a future refactor that passes a draft-derived \`Transaction\` in would silently write to \`undefined\`.

Consider restoring the guarantee without abandoning the unified class: have \`parseJournalFile\`/\`repo.find\`/\`repo.list\` return a narrowed type where the identity fields are required (e.g. a \`ParsedTransaction = Transaction & Required<Pick<Transaction, 'file' | 'startLine' | 'endLine' | 'rawBlock' | 'fingerprint'>>\` alias, or a type guard), so the \`!\` disappears from the service layer.

features/transactions/entry/draftReducer.ts
```diff
@@ -1,80 +1,53 @@
...
-  uid: input.uid,
-  postings: input.postings ?? emptyPostings(defaultCurrency),
-});
+): Transaction =>
+  new Transaction({
+    date: input.date,
+    payee: input.payee ?? '',
+    status: input.status ?? 'none',
+    note: input.note ?? '',
+    postings: input.postings ?? Transaction.empty(defaultCurrency).postings,
```
naeemba: nit: `input.postings ?? Transaction.empty(defaultCurrency).postings` constructs a full throwaway `Transaction` (two-posting blank draft) just to read its `.postings`. `emptyPostings(currency)` used to serve exactly this. Consider a small exported `blankPostings(currency)` helper on the model (or keep `Transaction.empty` but expose the posting seed directly) so the default path doesn't allocate and discard an instance.

lib/journal/service.ts
naeemba: Done in e1842b2. Added a ParsedTransaction alias (Transaction & Required<Pick<Transaction, 'file' | 'startLine' | 'endLine' | 'rawBlock' | 'fingerprint'>>) and narrowed parseJournalFile, ParsedJournal.transactions, repo.list/find, and service.findTransaction to it. parseJournalFile is the sole producer and vouches for the identity fields via a single cast, so the ~10 non-null assertions in performEdit/performDelete are gone and the file-write paths are compiler-protected again.

features/transactions/entry/draftReducer.ts
naeemba: Done in 1a71d33. Added a static Transaction.blankPostings(currency) that returns the two-posting seed, had Transaction.empty build from it, and pointed the draft initializer at it. The default path now allocates only the two postings instead of constructing and discarding a full blank Transaction.

PR #66, opened 2026-07-04
fix(review): drop blank posting rows when saving a template (A5)

naeemba:
## What & why (REVIEW.md item A5)

Saving a transaction as a template failed with an opaque **"Validation failed."** whenever the draft still carried a blank posting row — the default two-row scaffold or a leftover "Add posting" row. `canSaveTemplate` only counts postings with non-empty accounts (`>= 2`), but `Transaction.toTemplate()` mapped **every** posting into the template. An empty account fails `accountSchema` (`min(1)`), so a draft with two filled rows plus one blank row was rejected server-side while both filled rows looked perfectly valid — a UX dead-end for the feature's primary purpose.

## Fix

- **Root cause** — `Transaction.toTemplate()` (`lib/transactions/model.ts`), the single template builder, now filters out postings whose account is blank, so the saved shape matches what `canSaveTemplate` counted. The other caller (`RowActions`) builds from parsed transactions, which never carry blank accounts, so it is unaffected.
- **Defensive fallback** — `SaveAsTemplateDialog` previously rendered only `result.message` and dropped `fieldErrors` entirely. It now composes the message through a new pure `saveTemplateErrorMessage(result)` helper (`features/templates/saveTemplateError.ts`) that appends the distinct field-error messages, so any remaining validation failure names its cause instead of a bare "Validation failed."

## Tests

- `lib/transactions/model.test.ts` — `toTemplate` drops blank filler posting rows.
- `features/templates/saveTemplateError.test.ts` — message composition (bare message, default fallback, deduplicated field errors).

Full suite green (875 tests), `tsc --noEmit` clean, `eslint` clean.

### DIFF
--- a/features/templates/saveTemplateError.ts
+++ b/features/templates/saveTemplateError.ts
@@ -1,0 +1,21 @@
+import type { SaveTemplateResult } from './actions/saveTemplate';
+
+type SaveTemplateFailure = Extract<SaveTemplateResult, { ok: false }>;
+
+/**
+ * Human-readable error for a failed template save. The action reports the
+ * offending fields in `fieldErrors` (e.g. a blank posting account), but the
+ * dialog only had room for `message` — leaving the user with a bare
+ * "Validation failed." and no hint. Append the distinct field messages so the
+ * cause is at least named.
+ */
+export const saveTemplateErrorMessage = (
+  result: SaveTemplateFailure
+): string => {
+  const base = result.message ?? 'Could not save';
+  const fieldMessages = result.fieldErrors
+    ? [...new Set(Object.values(result.fieldErrors))]
+    : [];
+  if (fieldMessages.length === 0) return base;
+  return `${base} ${fieldMessages.join(' ')}`;
+};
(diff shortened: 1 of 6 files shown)


features/templates/saveTemplateError.ts
```diff
@@ -0,0 +1,21 @@
...
+
+type SaveTemplateFailure = Extract<SaveTemplateResult, { ok: false }>;
+
+/**
+ * Human-readable error for a failed template save. The action reports the
+ * offending fields in `fieldErrors` (e.g. a blank posting account), but the
+ * dialog only had room for `message` — leaving the user with a bare
+ * "Validation failed." and no hint. Append the distinct field messages so the
+ * cause is at least named.
+ */
```
naeemba, line 11: nit: colocated pure helper modules in this codebase use the `.util.ts` suffix (`features/activity/formatWhen.util.ts`, `features/transactions/AmountInput.util.ts`, `lib/transactions/carryAnnotations.util.ts`). For consistency this would be `saveTemplateError.util.ts` (and `saveTemplateError.util.test.ts`). Not blocking.

PR #67, opened 2026-07-05
fix(entry): keep guided type form open while its draft is incomplete

naeemba:
## Problem

On the Add Transaction page, picking **Expense** on the Types tab and typing a payee *before* filling the accounts/amount collapsed the form into the "This transaction's shape doesn't map to a quick type" notice, breaking the entry flow.

## Root cause

`TypeLens` computed its gate (`chipsDisabled`) purely from the live draft: `!isEmptyDraft(draft) && !detected`. A partially filled form produces a draft that is:

- no longer empty (`isEmptyDraft` returns `false` the moment `payee` is set), yet
- undetectable (empty accounts classify as `unknown`, so `detectType` returns `null`).

So the gate flipped to `true` and swapped the in-progress form out. It conflated *an incoming draft whose shape is unrecognized* with *a form the user is mid-way through filling*.

## Fix

- Extract the decision into a pure, tested helper `typeLensState.ts`:
  - `resolveTypeLensState(draft, picked)` — the unrecognized notice now fires **only when no type is picked**. Once a type is engaged, the guided form owns the draft and stays open.
  - `initialPickForDraft(draft)` — seeds `picked` from the detected type on mount. Because the Types tab unmounts when inactive, this re-seeds each visit, so editing a *recognized* transaction (e.g. clearing a field mid-edit) also no longer collapses the form — same bug class, now covered.
- `TypeLens` seeds `picked` via the initializer and reads its state from the helper.

## Tests

- New `typeLensState.test.ts` (7 cases), including the regression: partial expense + picked type keeps the form open; unrecognized incoming draft still disables the chips.
- Full `features/transactions/entry` suite green (106 tests); `tsc` and eslint clean.

### DIFF
--- a/features/transactions/entry/typeLensState.test.ts
+++ b/features/transactions/entry/typeLensState.test.ts
@@ -1,0 +1,99 @@
+// features/transactions/entry/typeLensState.test.ts
+import { describe, it, expect } from 'vitest';
+import { initDraft } from './draftReducer';
+import { initialPickForDraft, resolveTypeLensState } from './typeLensState';
+import { expenseAdapter } from './types/expense';
+
+const ctx = { defaultCurrency: 'USD' };
+
+const emptyDraft = () => initDraft({ date: '2026-07-05' }, 'USD');
+
+// A guided expense that the user has only started filling: a payee is typed
+// but the accounts and amount are still blank, so no adapter can detect it yet.
+const partialExpenseDraft = () =>
+  expenseAdapter.compile(
+    { ...expenseAdapter.emptyFields(ctx), payee: 'Cafe' },
+    ctx
+  );
+
+const expenseDraft = () =>
+  initDraft(
+    {
+      date: '2026-07-05',
+      payee: 'Cafe',
+      postings: [
+        { account: 'Expenses:Food', amount: '5', currency: 'USD' },
+        { account: 'Assets:Checking', amount: '-5', currency: 'USD' },
+      ],
+    },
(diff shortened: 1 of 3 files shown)


features/transactions/entry/typeLensState.test.ts
```diff
@@ -0,0 +1,99 @@
...
+  it('keeps a picked form open while its draft is still being filled', () => {
+    // Regression: typing a payee before the accounts/amount makes the draft
+    // non-empty yet undetectable. The picked form must stay open, not collapse
+    // into the "doesn't map to a quick type" notice.
+    const { selectedId, chipsDisabled } = resolveTypeLensState(
+      partialExpenseDraft(),
+      'expense'
+    );
+    expect(selectedId).toBe('expense');
+    expect(chipsDisabled).toBe(false);
```
naeemba, line 56: nit: this regression assertion passes for *any* non-null `picked` — `chipsDisabled` is `picked === null && ...`, so with `picked='expense'` it is `false` regardless of the draft. The load-bearing premise the docstring names ("non-empty yet undetectable") is never pinned: nothing asserts `isEmptyDraft(partialExpenseDraft())` is `false` or that `detectType(partialExpenseDraft())` is `null`. If `expenseAdapter.compile({payee}) ` ever produced an empty-looking draft the test would still pass for the wrong reason. Consider asserting those two premises so the test actually guards the partial-vs-unrecognized distinction that motivated the fix.

PR #68, opened 2026-07-05
fix(amounts): parse commodity in either position; visible text selection

naeemba:
## Problem

Two bugs surfaced while looking at a journal that uses a suffix-rendered commodity (`Kirt`) and no exchange rate for BTC:

1. **Amounts lose their number.** On the Periodic Balance page every row rendered just `Kirt` with no figure (while the total looked right by accident). Root cause: amount rendering assumed the commodity always precedes the number (`$ 100`), but ledger prints *word* commodities **after** the number (`71,214.5302 Kirt`). Everything that grabbed "the 2nd whitespace token" picked up the commodity instead of the value — affecting the balance rows, the chart bars (`NaN`), and the dashboard highest-expense picker. The same mis-parse made un-converted holdings (e.g. un-priced BTC/shares) render as an em-dash `—` on the Accounts page, i.e. **holdings appeared to vanish**.

2. **Text selection invisible in light mode.** `::selection` reused `--accent`, which is a near-white neutral in light mode, so the selection highlight was imperceptible on the near-white page. Text was selectable; you just couldn't see it.

## Fix

- Add `utils/amountParts.ts`: a parser that locates the numeric token regardless of commodity placement — prefix symbol (`$ 100` / `$100`), prefix code (`USD 100`), suffix code (`71,214.5302 Kirt`, `0.5 BTC`).
- Route `formatAmount`, the periodic-balance chart, the dashboard highest-expense picker, and `FriendlyBalance` through it. `features/accounts/amountParts.ts` now re-exports from `utils/` (no import churn for existing consumers).
- Un-converted commodities now show their original amount inline, matching the `BaseCurrencyBanner` promise ("affected reports show original currencies inline"). Updated the `FriendlyBalance` test that had codified the em-dash bug.
- `::selection` now uses `--fg`, visible in both light and dark themes.

## Verification

- `vitest run` — 888 passed (added regression tests for suffix commodities, negative suffix, glued prefix symbol, and the Kirt-row render).
- `tsc --noEmit` clean, eslint clean.

## Not included

The multi-commodity **continuation-line drop** in `lib/balance/parse.ts` (`if (!line.includes('|')) continue;` discards ledger's stacked extra-commodity lines) is a separate latent bug, left for a follow-up.

### DIFF
--- a/features/dashboard/Dashboard.utils.ts
+++ b/features/dashboard/Dashboard.utils.ts
@@ -1,3 +1,4 @@
+import { parseAmountParts } from '@/utils/amountParts';
 import runLedger from '@/utils/runLedger';
 
 export const getHighestExpense = (stdout: string): string => {
@@ -5,9 +6,8 @@ export const getHighestExpense = (stdout: string): string => {
   stdout.split('\n').forEach((expense) => {
     if (!expense) return;
     const amountField = expense.split('|')[1];
-    const amountToken = amountField?.split(' ')[1];
-    if (!amountToken) return;
-    const amount = Number(amountToken.replaceAll(',', ''));
+    if (!amountField) return;
+    const amount = parseAmountParts(amountField).signed;
     if (Number.isFinite(amount) && amount > highestExpense.amount) {
       highestExpense = { amount, str: expense };
     }
--- a/utils/formatAmount.tsx
+++ b/utils/formatAmount.tsx
@@ -1,3 +1,5 @@
+import { parseAmountParts } from '@/utils/amountParts';
+
 // Adds comma thousands separators to the integer part of a numeric string
 // while preserving its original decimal places. Operates on the string (not a
 // parsed Number) so ledger's display precision — including trailing zeros — is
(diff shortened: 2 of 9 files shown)


utils/formatAmount.tsx
```diff
@@ -30,14 +32,17 @@ const formatAmount = (str: string | undefined | null, withUnit: boolean) => {
...
-  } else if (withUnit) {
-    return formatAmountWithoutUnit(splitted[1], splitted[0]);
-  } else {
-    return formatAmountWithoutUnit(splitted[1]);
   }
+  const signedNumber = `${negative ? '-' : ''}${magnitude}`;
+  return formatAmountWithoutUnit(
+    signedNumber,
+    withUnit ? unit || undefined : undefined
+  );
```
naeemba, line 45: nit: `formatAmountWithoutUnit` always renders the unit as a prefix (`{unit}&nbsp;{grouped}`), so a suffix commodity now shows as `Kirt 71,214.5302` (and `Kirt (1,234.56)` when negative) rather than ledger's native `71,214.5302 Kirt`. The number is correctly recovered — this is purely placement — and `FriendlyBalance` makes the same prefix choice, so it is at least internally consistent. The two new suffix tests only assert containment, so the ordering is untested. Worth deciding whether suffix commodities should render on the trailing side.

features/dashboard/Dashboard.utils.ts
```diff
@@ -1,13 +1,13 @@
...
   let highestExpense = { amount: 0, str: '' };
   stdout.split('\n').forEach((expense) => {
     if (!expense) return;
     const amountField = expense.split('|')[1];
-    const amountToken = amountField?.split(' ')[1];
-    if (!amountToken) return;
-    const amount = Number(amountToken.replaceAll(',', ''));
+    if (!amountField) return;
+    const amount = parseAmountParts(amountField).signed;
     if (Number.isFinite(amount) && amount > highestExpense.amount) {
```
naeemba, line 11: nit: `parseAmountParts(...).signed` always returns a finite number (`Number(...) || 0`), so the `Number.isFinite(amount)` guard is now always true and can be dropped. It also quietly changed behavior: a row whose amount field has no numeric token used to be skipped (`if (!amountToken) return`), whereas it now yields `signed: 0` and just loses the `> highestExpense.amount` comparison — equivalent given the `amount: 0` seed, but the guard no longer does anything.

PR #69, opened 2026-07-06
fix(currencies): show journal commodities in currency picker

naeemba:
## Problem

The currency combobox searched only CoinGecko + fiat, so a user's own journal commodities (e.g. `KIRT`, `NIM`) no longer appeared and the picker showed **"Start typing to search"** on open. The `currencies` prop from `getAvailableCurrencies` still flowed down the tree but was dropped at `FormLens` before reaching the picker — a regression from the CoinGecko provider refactor (`7fcaf2d`).

## Fix

`searchCommoditiesAction` now merges the user's journal commodities into the results:

- Journal commodities rank **first**, tagged "in your journal".
- An existing user `commodity_mapping` wins over auto-classify, so selecting one never clobbers a hand-set kind.
- Results are de-duplicated by symbol (a journal `USD` isn't also offered as fiat or "use as manual").
- **Empty query** returns the journal list, so opening the picker surfaces the currencies actually in use; typing layers fiat + CoinGecko discovery on top.
- New users (empty journal) still get full online discovery — no regression.

The merge logic is a pure, unit-tested helper (`buildCommoditySuggestions`). The combobox now fetches on open and renders empty-query results.

## Behavior

| State | Before | After |
|---|---|---|
| Open picker | "Start typing to search" | Journal currencies listed |
| Typing | CoinGecko + fiat only | Journal matches first, then CoinGecko + fiat |

## Tests

- `buildCommoditySuggestions.test.ts` — 7 new cases (empty-query journal-only, substring filter, ranking, mapping preservation, fiat auto-classify, dedup, manual fallback).
- Full suite green (123 in currencies/prices/settings), `tsc` + eslint clean.

## Notes

- The now-dead `currencies` prop chain (`NewTransaction → TransactionEntry → FormLens`) can be removed in a follow-up.
- The combobox also gains journal suggestions in `PricesView`/`CurrenciesView` — intended, they're commodity pickers.

### DIFF
--- a/features/currencies/actions/searchCommodities.ts
+++ b/features/currencies/actions/searchCommodities.ts
@@ -1,9 +1,9 @@
 'use server';
 
 import { buildCommoditySuggestions } from './buildCommoditySuggestions';
+import type { CommodityContext } from './loadCommodityContext';
 import type { CommoditySuggestion } from './types';
 import { requireUser } from '@/lib/auth/require-user';
-import { commodityMappingRepository } from '@/lib/prices';
 import {
   getCoinSymbolMap,
   searchCoins,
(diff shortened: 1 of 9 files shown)


features/currencies/actions/searchCommodities.ts
```diff
@@ -13,48 +20,36 @@ export async function searchCommoditiesAction(
...
   const trimmed = query.trim();
-  if (!trimmed) return [];
-  const upper = trimmed.toUpperCase();
 
-  const suggestions: CommoditySuggestion[] = [];
+  // Journal commodities + the user's own mappings ground the list in what they
+  // actually use; CoinGecko/fiat only add discovery once they start typing.
+  const [{ currencies }, mappings] = await Promise.all([
+    getAvailableCurrencies(),
+    commodityMappingRepository.mapForUser(user.id),
```
naeemba: nit: this now runs `getAvailableCurrencies()` (which spawns a `ledger commodities` subprocess) plus `mapForUser` on every debounced keystroke, not just on open. The journal list and mappings are invariant for the duration of an open session, so the per-keystroke subprocess spawn + DB round-trip is pure overhead once the user starts typing. `getAvailableCurrencies` is wrapped in React `cache()`, but that only dedupes within a single request, so each server-action call re-runs it. Consider fetching the journal + mappings once when the picker opens (a separate action or passing them down) and letting `buildCommoditySuggestions` filter the already-loaded list, reserving the per-keystroke call for CoinGecko discovery. Non-blocking.

PR #70, opened 2026-07-07
chore(deps): update dependencies + adopt starter Postal transport

naeemba:
## Summary

Updates dependencies to latest compatible versions and migrates email delivery onto the next-starter 0.10.0 built-in Postal transport.

## Dependency updates

**In-range (patch/minor):** next 16.2.10, react/react-dom 19.2.7, vitest 4.1.10, @vitest/coverage-v8 4.1.10, prettier 3.9.4, lucide-react 1.23.0, @base-ui/react 1.6.0, @sentry/nextjs 10.63.0, resend 6.17.1, node-cron 4.6.0, tailwindcss 4.3.2, and others. Prettier 3.9's collapsed-union style reformatted a handful of union types.

**Majors verified safe:** recharts 3.8.0 → 3.9.2, @types/node 25 → 26, @naeemba/next-starter 0.9.1 → 0.10.0 (additive Postal email transport, no breaking changes).

**Held back — would break:** `eslint 9 → 10`. eslint-config-next 16.2.10 transitively pins eslint-plugin-react@7.37.5, which calls `context.getFilename()` (removed in eslint 10) and throws a `TypeError` on every lint. Stays on 9.39.4 until Next ships eslint-10-compatible plugins.

`@react-email/components` shows a "Deprecated" tag but is a peer dependency of next-starter — the starter drives its version, not this repo. Left as-is.

## Email transport migration

next-starter 0.10.0 ships a built-in Postal transport plus an `EMAIL_TRANSPORT` provider resolver equivalent to the app's hand-rolled one, so the custom `lib/email-transport.ts` is removed:

- `lib/auth.ts` no longer injects a `transport`; magic-link delivery resolves postal/console from `EMAIL_TRANSPORT` inside the starter.
- Encryption-reset and account-deletion code emails now go through the starter's `sendEmail`.
- `lib/email-transport.ts` deleted (starter's `sendViaPostal` is identical logic — same endpoint, `X-Server-API-Key`, `html_body`/`plain_body`, `status: 'success'` check).

No env changes required — the starter reads the same `POSTAL_API_URL` / `POSTAL_API_KEY` / `EMAIL_TRANSPORT` / `EMAIL_FROM` from `process.env`.

**Behavior delta:** dev console output loses the box-drawing layout for a plainer block ending in `[magic-link-log] email=... url=...`. Link still clearly visible.

## Verification

- type-check ✓
- lint ✓ (1 pre-existing `react-hooks/incompatible-library` warning, unchanged)
- 907 tests pass ✓
- production build ✓

Not live-tested against the actual Postal server — recommend one real magic-link + reset-code send before merge, since that path isn't unit-testable.


lib/env/index.ts
```diff
@@ -28,8 +28,9 @@ const envSchema = clientEnvSchema
     JOURNAL_QUOTA_MB: z.coerce.number().int().positive().default(100),
 
     // Email (magic link). 'postal' delivers via the self-hosted Postal server
-    // (see lib/email-transport.ts); 'console' prints the link/code to the dev
-    // server console and sends nothing — for local development. When 'postal',
+    // through the starter's built-in Postal transport; 'console' prints the
+    // link/code to the dev server console and sends nothing — for local
+    // development (both resolved by the starter from EMAIL_TRANSPORT). When 'postal',
```
naeemba, line 33: The `EMAIL_TRANSPORT: z.enum(['postal','console']).default('postal')` on the next line no longer drives actual email routing — this is a silent-failure trap in production.

Before this PR the app selected the transport itself: the deleted `lib/email-transport.ts` read the *parsed* `env.EMAIL_TRANSPORT` (default `'postal'`) and passed the resulting transport into `createAuth({ transport })` and the two `sendCode` helpers. Omitting `EMAIL_TRANSPORT` from the environment therefore resolved to Postal.

After this PR, routing is decided inside the starter's `resolveProvider`, which reads **`process.env.EMAIL_TRANSPORT` directly** (node_modules/@naeemba/next-starter/dist/chunk-KG6ZE5PP.js): when it's unset it returns `RESEND_API_KEY ? 'resend' : 'console'`. This app never sets `RESEND_API_KEY`, so with `EMAIL_TRANSPORT` unset the starter falls to the **console** transport and writes magic links + account-deletion / encryption-reset codes to the server log instead of sending them.

The Zod default lives only on the parsed `env` object; it is never written back to `process.env`, so the app and the starter now disagree. A deployer reading this schema reasonably concludes "EMAIL_TRANSPORT defaults to postal, I only need POSTAL_*" — and the `superRefine` below even enforces POSTAL_* under that default — yet every email silently goes to the log (the starter's only signal is a single `console.warn` when `NODE_ENV=production`). That breaks sign-in, account deletion, and encryption reset.

Fix: either drop `.default('postal')` and require `EMAIL_TRANSPORT` explicitly, or mirror the parsed value back into `process.env.EMAIL_TRANSPORT` after parsing so the starter reads the same value the app validated.

naeemba (commented):
Reviewed the dependency bumps and the Postal transport migration. The union-type reflows, the popover `collisionAvoidance` default, and the `journal/repository` reformat are all fine. One substantive concern on email routing (inline), plus one nit below.

nit: `vitest-setup.ts:16` still reads `Magic-link delivery (lib/email-transport.ts) reads these via the validated env schema` — that file is deleted by this PR. Update the pointer (delivery now goes through the starter's built-in Postal transport) so the comment doesn't dangle.

lib/env/index.ts
naeemba: Fixed in 74a957d. Confirmed the trap: the starter resolves the provider via resolveProvider(process.env), reading process.env.EMAIL_TRANSPORT directly, and the Zod .default('postal') only lands on the parsed env object. Rather than drop the default (which would force every deployer to set the var explicitly), lib/env/index.ts now mirrors the validated values back — process.env.EMAIL_TRANSPORT = env.EMAIL_TRANSPORT and process.env.EMAIL_FROM = env.EMAIL_FROM — so the app and starter share one source of truth. EMAIL_FROM had the identical defect (starter reads process.env.EMAIL_FROM directly, its Zod default auth@example.com never reached it) so it is mirrored too. Added a regression test asserting process.env carries the resolved default after load.

naeemba:
nit (vitest-setup.ts:16) addressed in 74a957d — the dangling pointer to the deleted lib/email-transport.ts is replaced with: "Magic-link delivery goes through the starter's built-in Postal transport, which reads these from process.env; placeholders satisfy the env-schema validators without contacting Postal."

PR #71, opened 2026-07-07
fix(transactions): correct fix-balance current balance and format amounts

naeemba:
## Problem

On the add-transaction **Types → Fix balance** form, selecting an account showed two bugs:

1. **Wrong current balance** — always displayed `Now: 0`, regardless of the account's real balance.
2. **Unformatted figures** — the "Now" and "Implied adjustment" amounts rendered without thousands separators (e.g. `134097806.00`).

## Cause

1. `getAccountBalance` ran `ledger balance` with `--collapse`, which relabels the emitted row to the top-level account (e.g. `Assets`). `extractAccountBalance` matches the exact account name (`Assets:Bank:Blubank`), so the row never matched and the fallback `"0"` was returned.
2. The `current` and `implied` values were interpolated raw into the display string.

## Fix

- Drop `--collapse`. The exact-match row now carries the correct rolled-up `%T` total for the account (verified against a real journal: `Assets:Bank:Blubank|KIRT 1,779.540`).
- Format both displayed figures with `groupAmountInput` — the same grouping helper the `AmountInput` field uses — so display never drifts from the input.

## Verification

- `tsc --noEmit` clean.
- `fixBalancePreview` + `AmountInput.util` tests pass (22).
- End-to-end lookup against a real journal returns the correct balance.

### DIFF
--- a/features/transactions/entry/typeForms/FixBalanceForm.tsx
+++ b/features/transactions/entry/typeForms/FixBalanceForm.tsx
@@ -3,7 +3,7 @@
 
 import React, { useEffect, useMemo, useRef, useState } from 'react';
 import AmountInput from '../../AmountInput';
-import { groupAmountInput } from '../../AmountInput.util';
+import { groupAmountInput, decimalPlaces } from '../../AmountInput.util';
 import { headerOf } from '../types/adapter';
 import { fixBalanceAdapter, type FixBalanceFields } from '../types/fixBalance';
 import { HeaderFieldsEditor } from './HeaderFields';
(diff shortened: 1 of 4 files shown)


features/transactions/entry/typeForms/FixBalanceForm.tsx
```diff
@@ -105,10 +106,11 @@ export function FixBalanceForm({
         <div className="text-xs text-muted-foreground tabular-nums">
           {current === null
             ? 'Enter an account to see its current balance.'
-            : `Now: ${current} ${fields.targetCurrency}`}
+            : `Now: ${groupAmountInput(current)} ${fields.targetCurrency}`}
           {implied !== null && (
             <span className="ml-2">
-              · Implied adjustment: {implied} {fields.targetCurrency}
+              · Implied adjustment: {groupAmountInput(implied)}{' '}
```
naeemba, line 114: nit: now that both figures are grouped onto the same line, the precision mismatch is more visible: `current` renders at its native ledger precision (e.g. `Now: 1,779.540` for a 3-decimal commodity) while `implied` is hard-coded to `.toFixed(2)` (`Implied adjustment: 220.46`). Consider deriving the implied precision from `current`/`targetAmount` so the two amounts line up. The `.toFixed(2)` predates this PR, so this is optional.

PR #72, opened 2026-07-07
feat(prices): known-prices list and per-commodity history page

naeemba:
## What

Adds a read-only view of the latest price `ledger` knows for every commodity you hold, plus a per-commodity price-history page.

- **`/prices` is now tabbed:**
  - **Known prices** (default): a table of every held commodity with its latest known price, the date that price is from, its age (with a stale marker past 7 days), and the source of the price (`fetched` / `manual` / `journal` / `base` / no price).
  - **Manual entry**: the existing manual-price form, unchanged.
- **`/prices/[symbol]`**: a price-history detail page — a line chart over time plus a dated table.

## How

- Prices come straight from `ledger prices`, which merges the journal's own `P`/`@` prices with the fetched + manual price database (via `--file` + `--price-db`), so the view reflects exactly what ledger values each commodity at.
- Only commodities you actually hold are listed (`ledger commodities`), including the base currency (USD), shown as `base` at 1.00. Held-but-unpriced commodities show a "no price" gap row.
- Provenance (the Source column) is derived by correlating the latest price against the manual- and fetched-price tables, since ledger output carries no source.

## Notable correctness points

- **Staleness dates from when the price last changed, not the newest data row.** `ledger prices` forward-carries the prevailing price onto every date a commodity is transacted, so a naive "last row" read would show a recent posting date and make a stale price look fresh. A dedicated helper (`latestGenuinePrice`) tracks the last actual price change; a regression test guards it.
- **Input hardening:** the per-commodity ledger call uses a `--` end-of-options separator plus an empty/newline guard so a crafted commodity symbol cannot smuggle a ledger flag.
- Prices render with thousands separators; quotes display normalized (USD).

## Testing

- 931 tests passing; `tsc`, `eslint`, and `next build` all clean (both routes emit as dynamic).
- New coverage: parser + provenance unit tests, and integration tests for base/journal/fetched/manual sources, gap rows, and the forward-carry staleness scenario.

## Deferred (backlog)

- `listKnownPrices` fans out one `ledger` process per held commodity; fine at personal scale, worth a concurrency limiter if portfolios grow large.
- Stale shown as text rather than a styled badge.

Design + plan: `docs/superpowers/specs/2026-07-08-known-prices-page-design.md`, `docs/superpowers/plans/2026-07-08-known-prices-page.md`.

### DIFF
--- a/features/prices/PriceHistoryView.tsx
+++ b/features/prices/PriceHistoryView.tsx
@@ -1,0 +1,98 @@
+'use client';
+
+import {
+  CartesianGrid,
+  Line,
+  LineChart,
+  ResponsiveContainer,
+  Tooltip,
+  XAxis,
+  YAxis,
+} from 'recharts';
+import { TableScroll } from '@/components/ui/table';
+import type { PricePoint } from '@/lib/prices';
+import Link from 'next/link';
+
+type Props = { symbol: string; points: PricePoint[] };
+
+export const PriceHistoryView = ({ symbol, points }: Props) => {
+  const quote = points.at(-1)?.quote ?? '';
+
+  return (
+    <div className="mx-auto w-full max-w-3xl space-y-6 p-4">
+      <header className="space-y-1">
+        <Link
+          href="/prices"
+          className="text-sm text-muted-foreground hover:underline"
+        >
+          ← Back to prices
+        </Link>
+        <h1 className="text-2xl font-semibold">{symbol} price history</h1>
+      </header>
+
+      {points.length === 0 ? (
+        <p className="text-muted-foreground text-sm">
+          Ledger has no price history for {symbol}.
+        </p>
+      ) : (
+        <>
+          <div className="h-72 w-full rounded-2xl border border-border bg-card p-4 shadow-sm">
+            <ResponsiveContainer width="100%" height="100%">
+              <LineChart
+                data={points}
+                margin={{ top: 8, right: 16, bottom: 8, left: 8 }}
+              >
+                <CartesianGrid
+                  strokeDasharray="3 3"
+                  className="stroke-border"
+                />
+                <XAxis dataKey="date" tick={{ fontSize: 12 }} minTickGap={24} />
+                <YAxis
+                  tick={{ fontSize: 12 }}
+                  width={72}
+                  domain={['auto', 'auto']}
+                />
+                <Tooltip
+                  formatter={(value) => [`${value ?? ''} ${quote}`, 'Price']}
+                />
+                <Line
+                  type="monotone"
+                  dataKey="price"
+                  dot={false}
+                  strokeWidth={2}
+                />
+              </LineChart>
+            </ResponsiveContainer>
+          </div>
+
+          <div className="overflow-hidden rounded-2xl border border-border bg-card shadow-sm">
+            <TableScroll bleed={false}>
+              <table>
+                <thead>
+                  <tr>
+                    <th>Date</th>
+                    <th className="text-right">Price</th>
+                    <th>Quote</th>
+                  </tr>
+                </thead>
+                <tbody>
+                  {[...points].reverse().map((point) => (
+                    <tr key={`${point.date}-${point.price}`}>
+                      <td className="whitespace-nowrap text-muted-foreground">
+                        {point.date}
+                      </td>
+                      <td className="text-right tabular-nums whitespace-nowrap">
+                        {point.price}
+                      </td>
+                      <td className="whitespace-nowrap">{point.quote}</td>
+                    </tr>
+                  ))}
+                </tbody>
+              </table>
+            </TableScroll>
+          </div>
+        </>
+      )}
+    </div>
+  );
+};
--- a/features/prices/PriceHistoryView.tsx
+++ b/features/prices/PriceHistoryView.tsx
@@ -15,6 +15,10 @@ import Link from 'next/link';
 
 type Props = { symbol: string; points: PricePoint[] };
 
+const priceFormatter = new Intl.NumberFormat('en-US', {
+  maximumFractionDigits: 8,
+});
+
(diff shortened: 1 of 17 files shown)


features/prices/PriceHistoryView.tsx
```diff
@@ -0,0 +1,108 @@
...
+  XAxis,
+  YAxis,
+} from 'recharts';
+import { TableScroll } from '@/components/ui/table';
+import type { PricePoint } from '@/lib/prices';
+import Link from 'next/link';
+
+type Props = { symbol: string; points: PricePoint[] };
+
+const priceFormatter = new Intl.NumberFormat('en-US', {
```
naeemba: This `priceFormatter` (`Intl.NumberFormat('en-US', { maximumFractionDigits: 8 })`) is defined identically here and in `KnownPricesView.tsx:9`. Both are new files in this PR, so the two copies can silently drift (e.g. someone changes `maximumFractionDigits` in one). Extract it once into a colocated `features/prices/format.util.ts` (or similar) and import it in both views.

features/prices/PriceHistoryView.tsx
```diff
@@ -0,0 +1,108 @@
...
+  maximumFractionDigits: 8,
+});
+
+export const PriceHistoryView = ({ symbol, points }: Props) => {
+  const quote = points.at(-1)?.quote ?? '';
```
naeemba: nit: the quote shown here is the raw ledger quote (`$`), because `listPriceHistory` returns `parsePriceHistory` output unchanged, whereas the Known-prices list normalizes it (`quote: quoteNormalized` in `service.ts`, rendering `USD`). Same commodity therefore reads `50,000 USD` on the list but `50,000 $` on the detail page/tooltip. Consider normalizing here too for consistency.

lib/prices/knownPrices.ts
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * The current price and the date it last changed. `ledger prices` forward-
+ * carries the prevailing price onto every posting date, so the final row's
+ * date can be a transaction date rather than when the price was actually set.
+ * Returns the last point's price/quote paired with the date the price value
+ * last changed (the start of the final constant-price run), so staleness
+ * reflects how long the current price has stood. Returns null for empty input.
+ */
+export const latestGenuinePrice = (points: PricePoint[]): PricePoint | null => {
```
naeemba, line 71: nit: `latestGenuinePrice` treats any later row carrying the same price value as "not a change", which correctly discounts forward-carried *posting* rows — but it can't distinguish those from a genuinely re-fetched-but-unchanged price. A commodity that is fetched daily yet holds an identical value (e.g. a stablecoin quoting exactly `1`) would take the *first* occurrence's date and be marked stale despite fresh data. Float quotes rarely repeat exactly, so this is a low-probability edge; flagging for awareness rather than a required change.

naeemba (commented):
Read-only known-prices list plus per-commodity history page. Correctness core (injection guard via `--`, staleness from `latestGenuinePrice` to defeat forward-carried posting rows, provenance correlation) is sound and well-tested. The double `decodeURIComponent` in the `[symbol]` route matches the established sibling convention (`app/accounts/[account]`, `app/registers/monthly/[account]`) so it is intentional. A few minor polish items below.

features/prices/PriceHistoryView.tsx
naeemba: Extracted into features/prices/format.util.ts and imported in both views. Fixed in 03be1d7.

features/prices/PriceHistoryView.tsx
naeemba: Normalized the quote in the detail view (chart tooltip and Quote column) via the same normalizeCommoditySymbol used by the list, so $ now reads as USD on both pages. Fixed in a42cb61.

lib/prices/knownPrices.ts
naeemba: Leaving as-is. This is the accepted tradeoff: without provenance from ledger we can't distinguish a forward-carried posting row from a genuinely re-fetched identical value, and discounting repeated values is the safer default for the far more common posting-carry case. As you note, float quotes rarely repeat exactly, so the stablecoin-quoting-exactly-1 case is a low-probability edge. If it becomes real we can key staleness off fetch-run provenance rather than the price series.

features/prices/PriceHistoryView.tsx
naeemba: Confirmed resolved: features/prices/format.util.ts now owns priceFormatter and both KnownPricesView and PriceHistoryView import it. Resolving.

features/prices/PriceHistoryView.tsx
naeemba: Confirmed resolved: PriceHistoryView now runs the quote through displayQuote (normalizeCommoditySymbol) in both the tooltip and the Quote column, matching the list. Resolving.

lib/prices/service.ts
```diff
@@ -301,6 +371,84 @@ export class PriceService {
...
+        priceKey(row.symbol, row.quote, utcDate(row.pricedAt))
+      )
+    );
+    const fetchedKeys = new Set(
+      fetched.map((row) => priceKey(row.symbol, row.quote, row.fetchedDate))
+    );
+
+    const today = utcDate(new Date());
+
+    const rows = await Promise.all(
```
naeemba: nit: this fans out one `ledger prices <symbol>` subprocess per held commodity with no concurrency bound (`Promise.all` over `held`), and each invocation re-parses the full journal + price-db. On a `force-dynamic`, uncached page that means N concurrent ledger processes per /prices view. Fine at personal-finance scale, but if a user accumulates many commodities consider a bounded map (e.g. p-limit) or a single-pass parse. The per-symbol choice itself is sound — it sidesteps the quote-vs-symbol ambiguity the design doc flagged for the combined `pricedb` report — so this is purely about the unbounded fan-out.

PR #73, opened 2026-07-08
feat(prices): toggle known prices between original quote and base currency

naeemba:
## Summary

Adds a toggle on the **Known prices** tab to switch each row between its original journal quote and a value normalized into the base currency (USD).

Previously every commodity showed its latest price in whatever quote its newest `P` directive used — fetched prices in USD, but journal-sourced ones in their original quote (BTC in DAI, Nim/Sekke in KIRT). The result was a column of mixed currencies that couldn't be compared at a glance.

## How it works

- **Original quote** (default) — unchanged behavior.
- **In USD** — every held commodity is valued into the base via a single `ledger balance -X <base>` call driven by a throwaway probe journal (one balanced `1 <commodity>` transaction each). Ledger chains cross-rates through the full price graph (BTC→DAI→USD, Nim→KIRT→USD). Commodities with no path to the base show "no price" — never a fabricated 0.
- Date / Age / Source columns are carried over unchanged from the original-quote view (keep-raw): a USD value derived from a stale underlying price is still flagged stale.
- The toggle is a `?base=usd` search param (server refetch), rendered as a segmented control shown only on the Known prices tab.

`ledger prices -X` does not convert (verified) — only a `balance -X` report chains rates. Because ledger treats `$` and the literal `USD` as distinct commodities, the probe journal injects a `$` = 1 USD bridge (dated in the distant past so any real `$`/USD price wins) so dollar-denominated holdings — the common journal convention — value correctly.

## Cost

One extra ledger subprocess beyond the existing per-commodity fan-out, regardless of how many commodities are held.

## Testing

- `parseBaseBalance` unit tests (index keying, unconvertible commodity, thousands stripping, offset/garbage rejection, exact format-string assertion).
- Real-ledger integration tests for `listKnownPricesInBase`: chained conversion, no-path → null, keep-raw columns, base row, dollar-denominated holding via the bridge.
- Full suite 948/948 green; type-check + lint clean; `next build` compiles `/prices` as a dynamic route.

## Not covered

Live in-browser verification against a real journal was not run (requires an authenticated session).

### DIFF
--- a/features/prices/KnownPricesView.tsx
+++ b/features/prices/KnownPricesView.tsx
@@ -40,14 +40,14 @@ export const KnownPricesView = ({ rows, baseMode, baseCurrency }: Props) => (
       >
         <Link
           href="/prices"
-          aria-current={!baseMode}
+          aria-current={!baseMode ? 'page' : undefined}
           className={segmentClass(!baseMode)}
         >
           Original quote
         </Link>
         <Link
           href="/prices?base=usd"
-          aria-current={baseMode}
+          aria-current={baseMode ? 'page' : undefined}
           className={`${segmentClass(baseMode)} border-l border-border`}
         >
           In {baseCurrency}
--- a/features/prices/KnownPricesView.tsx
+++ b/features/prices/KnownPricesView.tsx
@@ -5,7 +5,15 @@ import { TableScroll } from '@/components/ui/table';
 import type { KnownPrice } from '@/lib/prices';
 import Link from 'next/link';
 
-type Props = { rows: KnownPrice[] };
+type Props = { rows: KnownPrice[]; baseMode: boolean; baseCurrency: string };
+
+const segmentClass = (active: boolean): string =>
+  [
+    'px-3 py-1.5 font-medium transition-opacity',
+    active
+      ? 'bg-accent text-accent-foreground'
+      : 'opacity-60 hover:opacity-100',
+  ].join(' ');
 
 const sourceLabel: Record<KnownPrice['source'], string> = {
   fetched: 'Fetched',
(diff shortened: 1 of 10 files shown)


lib/prices/service.ts
```diff
@@ -457,6 +461,105 @@ export class PriceService {
...
+    const bridge = base === 'USD' ? 'P 2000-01-01 $ 1 USD\n\n' : '';
+
+    // One balanced `1 <symbol>` transaction per commodity, indexed by position
+    // so the account name carries no commodity-specific characters. Blank lines
+    // separate transactions; an explicit offset balances each one.
+    const journal =
+      bridge +
+      probeSymbols
+        .map((symbol, index) => {
+          const needsQuote = /[^A-Za-z0-9_]/.test(symbol);
```
naeemba: The `needsQuote` predicate only quotes symbols containing a non-`[A-Za-z0-9_]` char, but ledger requires a commodity name to be double-quoted whenever it contains **digits** too. A digit-bearing symbol like `1INCH` (memory notes 1inch is a real held commodity) or `SOL2` is emitted unquoted as `1 1INCH`, which ledger rejects:

```
Error: Unexpected char '1' (Note: inline math requires parentheses)
```

Verified against Ledger 3.4.1. `normalizeCommoditySymbol` permits digits (`/^[A-Za-z0-9]+$/`), so such symbols flow straight into the probe journal. Because the whole probe journal is valued in a single `balance` call, one unquotable holding fails the entire parse, the `catch` degrades **every** row to `price: null`, and the USD view silently shows "no price" for all commodities — not just the offending one.

Quote whenever the symbol is not purely alphabetic (e.g. `!/^[A-Za-z]+$/.test(symbol)`), and add a regression test with a digit-bearing ticker (the current tests use only letter-only symbols: BTC, Nim, KIRT, DAI, XOF, which is why this slipped through).

lib/prices/service.ts
```diff
@@ -457,6 +461,105 @@ export class PriceService {
...
+      .map((row) => row.symbol);
+
+    if (probeSymbols.length === 0) return raw.map(toBaseRow);
+
+    // Ledger treats `$` and the literal `USD` as distinct commodities and will
+    // not bridge them on its own, so a holding priced in `$` (the common
+    // journal convention) would fail to value into a `USD` base. Seed the probe
+    // journal with the identity `$` = 1 USD, dated in the distant past so any
+    // real `$`/USD price in the user's journal takes precedence over it.
+    const bridge = base === 'USD' ? 'P 2000-01-01 $ 1 USD\n\n' : '';
```
naeemba: The `$` = 1 USD bridge is build-dependent and can abort the entire valuation. On a Ledger build that canonicalizes `$` to `USD` (my local 3.4.1 does), `P 2000-01-01 $ 1 USD` is a self-price and hard-aborts parsing:

```
Error: Assertion failed in .../history.cc line 282: source != price.commodity()
```

When that fires, the `balance` call throws, the `catch` degrades every row to "no price", and USD mode is fully broken. The whole feature rests on the premise that `$` and `USD` are distinct commodities in the target ledger build — true on your CI (hence green tests) but not universally. Please confirm the production ledger binary treats them as distinct (and ideally guard/probe for it) before relying on the bridge; on builds where they're already the same, the bridge is both unnecessary and fatal.

Separately (nit): the bridge hardcodes the literal `'USD'`, which the plan's own global constraint says to avoid — though it's defensible here since `$` is inherently dollar-specific.

lib/prices/service.ts
```diff
@@ -457,6 +461,105 @@ export class PriceService {
...
+        );
+      } catch {
+        // Ledger failed → no valuation available; degrade to no-price rows.
+        return raw.map(toBaseRow);
+      }
+
+      const valued = parseBaseBalance(stdout);
+      return raw.map((row) => {
+        if (normalizeCommoditySymbol(row.symbol) === base) return row;
+        const index = probeSymbols.indexOf(row.symbol);
```
naeemba: nit: `probeSymbols.indexOf(row.symbol)` inside `raw.map(...)` is O(n²) over held commodities. Building the probe journal already assigns a stable index per symbol — capture a `Map<string, number>` (or map probe results back by symbol directly) to keep it linear. Negligible for small holdings, but free to fix.

lib/prices/service.ts
naeemba: Fixed in 511e7f5, though the mechanism differs from the description. Digit-bearing commodities are only legal double-quoted in a journal (`1 1INCH` and even `1 SOL2` fail to parse), and `ledger commodities` surfaces them back with the quotes intact (`"1INCH"`). So `row.symbol` already carries the quote chars and was excluded by the existing `!/["\n\r]/` guard — the failure mode was a silent per-row `price: null` for a convertible holding, not a whole-journal abort (the offending symbol never reached the probe, quoted or not).

The suggested `!/^[A-Za-z]+$/` tweak would not have helped, since the symbol never reaches that line unquoted. The fix instead strips the surrounding quote pair and always re-quotes the bare name in the probe (`1 "1INCH"`), so 1inch now values into the base. Added a regression test with a digit-bearing ticker asserting the quoted row identity.

lib/prices/service.ts
naeemba: Fixed in 511e7f5. Confirmed against the production binary (Ledger 3.4.1, same build the app shells out to): `P 2000-01-01 $ 1 USD` does abort with `Assertion failed ... source != price.commodity()`, so the concern is real on our own target build, not just hypothetical.

Rather than probe for the behaviour up front (an extra ledger invocation), the valuation now retries without the bridge when the first attempt throws. On builds that canonicalize `$` to `USD` the bridge directive is both fatal and unnecessary — they already value `$` holdings into `USD` natively — so the retry recovers cleanly (verified: the no-bridge probe emits commodity `$`, which the existing `normalizeCommoditySymbol(hit.commodity) === base` comparison already folds to USD). On builds that keep them distinct, the first attempt with the bridge succeeds. Correct across both.

On the hardcoded `USD` literal: leaving it, since the bridge is inherently dollar-specific ($ has no meaning for any other base) — as you noted, defensible here.

lib/prices/service.ts
naeemba: Fixed in 511e7f5. The re-value pass now builds a `Map<string, number>` from symbol to probe index once (first index wins on duplicate symbols) and does an O(1) `symbolIndex.get(row.symbol)` per row instead of the O(n^2) `indexOf` scan.

features/prices/KnownPricesView.tsx
```diff
@@ -22,70 +30,94 @@ const ageLabel = (row: KnownPrice): string => {
...
+      >
+        <Link
+          href="/prices"
+          aria-current={!baseMode ? 'page' : undefined}
+          className={segmentClass(!baseMode)}
+        >
+          Original quote
+        </Link>
+        <Link
+          href="/prices?base=usd"
```
naeemba: nit: the toggle label is dynamic (`In {baseCurrency}`) but the link target and the page-level branch are both hardcoded to the literal `usd` (`href="/prices?base=usd"` here, `base === 'usd'` in page.tsx, and `base === 'USD'` for the bridge in service.ts). `resolveBaseCurrency` is written to be swappable (there's a base-currency-selector plan in docs/superpowers/plans/), so the day it returns anything but USD, this segment would render "In EUR" while still routing through the USD-only param/bridge path. Scoped to USD today per the spec, so fine to leave — but a `?base=1` / `?base=base` flag decoupled from the currency code would avoid the latent trap.

lib/prices/service.ts
```diff
@@ -457,6 +461,140 @@ export class PriceService {
...
+    // Ledger builds that keep `$` and the literal `USD` distinct will not
+    // bridge a `$`-priced holding into a `USD` base on their own, so seed the
+    // identity `$` = 1 USD, dated in the distant past so any real `$`/USD price
+    // in the user's journal takes precedence over it. Builds that canonicalize
+    // `$` to `USD` (e.g. Ledger 3.4.x) reject that directive as a self-price
+    // ("Assertion failed … source != price.commodity()") and abort the whole
+    // parse — but they also already value `$` into `USD` natively, so the
+    // bridge is redundant there. Try with the bridge, then retry without it on
+    // failure to stay correct across both build behaviours.
+    const bridge = base === 'USD' ? 'P 2000-01-01 $ 1 USD\n\n' : '';
```
naeemba: nit / cost note: on the production binary (Ledger 3.4.1, which per your own comment on the bridge canonicalizes `$`→`USD` and aborts on the self-price directive), the bridge attempt on line 575 *always* throws and the retry on line 578 *always* runs — so every `?base=usd` render spends two ledger subprocesses here, not the "one extra subprocess" the PR body advertises. The order is correct and must not simply be flipped (probing without the bridge first would misreport a `$`-priced holding on a distinct-`$` build as already-in-base, since `normalizeCommoditySymbol('$') === base`). If the steady-state double-spawn matters, memoize the detected build behaviour once (mindful of the Next module-singleton caveat — pin it on `globalThis`) so subsequent renders skip the known-fatal bridge attempt. Optional.

features/prices/KnownPricesView.tsx
naeemba: Fixed in d6532b4. The toggle now routes through the mode flag `?base=base` ("value into the resolved base currency") with the page branching on `base === 'base'`, so the label and route no longer diverge when resolveBaseCurrency returns a non-USD code. The `base === 'USD'` guard in service.ts stays — it's genuinely currency-semantic (the `$` bridge only applies to USD), not a routing literal.

lib/prices/service.ts
naeemba: Fixed in d6532b4. The binary's bridge behaviour is now memoized on globalThis (`lib/prices/bridgeViability.ts`, pinned there per the Next module-singleton caveat). The first USD-base valuation still probes with the bridge, but once observed to abort, every later render skips the known-fatal attempt and its retry — steady state is one subprocess on a canonicalizing build. Order is unchanged for the first observation, so the distinct-`$` correctness you flagged is preserved (that build records 'viable' and keeps using the bridge).

PR #74, opened 2026-07-08
feat(prices): value known-prices toggle into the selected display currency

naeemba:
## Summary

Follow-up to #73. The Known-prices "In …" toggle valued every row into the hardcoded USD **pricing** base (`resolveBaseCurrency`), ignoring the user's currency selector — so it always read "In USD" no matter what currency was chosen.

This wires the toggle to the **display** currency (`getBaseCurrency()` — session cookie → saved setting → default, the same resolver the balance/net-worth pages use). The USD pricing base is untouched: prices are still fetched and stored in USD; only the on-screen valuation target changes.

## Changes

- `listKnownPricesInBase(userId, targetCurrency?)` — new optional target; values via `balance -X <target>`. Ledger inverts a fetched `EUR-in-USD` leg to reach a non-USD target (e.g. BTC → USD → EUR).
- The `$` = 1 USD bridge is no longer gated on a USD target — it is a universal identity, so dollar-priced holdings reach any target as `$` → USD → target. Ledger-build viability still gates it.
- `app/prices/page.tsx` resolves `getBaseCurrency()` and passes it as both the valuation target and the toggle label; the label now reads "In <selected currency>".

## Testing

- New integration test: values BTC into EUR (40000 USD / 1.10 = 36363.64 EUR).
- Full `lib/prices` suite 107/107; type-check + lint clean.

## Note

Commodities with no price path to the selected currency still show "no price" (honest — no fabricated value).

### DIFF
--- a/lib/prices/service.ts
+++ b/lib/prices/service.ts
@@ -467,14 +467,21 @@ export class PriceService {
   }
 
   /**
-   * Latest known price for every held commodity, valued into the base currency.
+   * Latest known price for every held commodity, valued into a target currency.
    * Reuses the raw rows from `listKnownPrices` for provenance and staleness,
-   * then re-values each non-base holding through ledger's full price graph in a
-   * single `balance -X <base>` call driven by a throwaway probe journal. A
-   * holding with no conversion path to the base yields `price: null`.
+   * then re-values each non-target holding through ledger's full price graph in
+   * a single `balance -X <target>` call driven by a throwaway probe journal. A
+   * holding with no conversion path to the target yields `price: null`.
+   *
+   * `targetCurrency` is the user-facing display currency (their selector) — NOT
+   * the USD pricing base that `resolveBaseCurrency` returns for storage. It
+   * defaults to that pricing base only so existing callers stay valid.
    */
-  async listKnownPricesInBase(userId: string): Promise<KnownPrice[]> {
-    const base = await this.resolveBaseCurrency(userId);
+  async listKnownPricesInBase(
+    userId: string,
+    targetCurrency?: string
+  ): Promise<KnownPrice[]> {
+    const base = targetCurrency ?? (await this.resolveBaseCurrency(userId));
     const raw = await this.listKnownPrices(userId);
 
     // The base row keeps whatever symbol ledger prints for it (e.g. `$`), so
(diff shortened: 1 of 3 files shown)


naeemba (commented):
Reviewed the toggle rewiring, the bridge-universality change, and the new integration test.

lib/prices/service.ts
```diff
@@ -576,8 +583,12 @@ export class PriceService {
     // failure to stay correct across both build behaviours — and remember which
     // build this binary is, so a canonicalizing build skips the known-fatal
     // bridge attempt (and its guaranteed retry) on every later render.
+    // The bridge states `$` = 1 USD, a universal identity (the pricing base is
+    // always USD), so it is valid for any target: a dollar-priced holding
+    // reaches a non-USD target as `$`->USD->target. Only the ledger-build
+    // viability gates it, not the target currency.
     const bridge = 'P 2000-01-01 $ 1 USD\n\n';
-    const wantsBridge = base === 'USD' && getBridgeViability() !== 'aborts';
+    const wantsBridge = getBridgeViability() !== 'aborts';
```
naeemba, line 594: The specific behavior this line now enables — a `$`-denominated holding reaching a **non-USD** target via the `$ = 1 USD` bridge — is not covered by any test. The new EUR test (service.test.ts:867) prices BTC directly in the `USD` literal, so it never touches the bridge; the only bridge test (service.test.ts:815) targets USD, the path that already worked before the `base === 'USD'` gate was removed. Add a case that holds a `$`-priced commodity and values it into EUR (e.g. `P BTC $40000` + `P EUR 1.10 USD`, hold 1 BTC, target `EUR`, expect ~36363.64) so the exact interaction this change unlocks is pinned down.

lib/prices/service.ts
```diff
@@ -467,14 +467,21 @@ export class PriceService {
...
+   * the USD pricing base that `resolveBaseCurrency` returns for storage. It
+   * defaults to that pricing base only so existing callers stay valid.
    */
-  async listKnownPricesInBase(userId: string): Promise<KnownPrice[]> {
-    const base = await this.resolveBaseCurrency(userId);
+  async listKnownPricesInBase(
+    userId: string,
+    targetCurrency?: string
+  ): Promise<KnownPrice[]> {
+    const base = targetCurrency ?? (await this.resolveBaseCurrency(userId));
```
naeemba, line 484: nit: downstream (the base-row early return, ~line 607) now keeps the target currency's own held row at its raw USD-quoted price. If the user actually holds EUR while the toggle reads "In EUR", the EUR row renders as e.g. `1.10 USD` (quote `USD`) among rows all valued in EUR — inconsistent. Harmless for the USD default (the USD numeraire row is effectively identity/absent), but newly reachable with a non-USD target. Consider representing the base row as identity (`1 <target>`) or excluding it when `target !== USD`.

lib/prices/service.ts
naeemba: Added in ab0cc41: a test holding a `$`-priced BTC (`P BTC $40000`) with `P EUR 1.10 USD`, valued into `EUR`, asserting ~36363.64. It only reaches the USD pricing base through the injected `$` = 1 USD bridge and then inverts the EUR/USD leg, so it exercises the exact `$`->USD->non-USD-target path the removed `base === 'USD'` gate unlocks (distinct from the existing EUR test, which prices BTC in the `USD` literal, and the existing bridge test, which targets USD).

lib/prices/service.ts
naeemba: Fixed in 82c617a: the base row is now emitted as the identity `1 <target>` (both the early no-probe path and the final re-value pass) instead of its raw quote. For the USD default this is a no-op — the held USD row is already `1 USD` — but with a non-USD target the held target currency no longer renders at e.g. `1.10 USD` among rows all valued in the target; it now shows `1 EUR`. Symbol (e.g. `$`), date, staleness and provenance are preserved.

PR #75, opened 2026-07-08
fix(prices): normalize display-currency target so mixed-case selections value

naeemba:
## Summary

Follow-up to #74. Selecting a display currency whose case differs from how it is stored/priced (e.g. **Kirt** vs the `KIRT` commodity) made every row show "no price".

Held symbols and ledger output are normalized (uppercased, `$`→USD) before every `=== base` comparison, but the target currency from `getBaseCurrency()` was passed through raw. So `normalize("KIRT") === "Kirt"` was always false — every valued row got nulled and the base row was never detected.

## Fix

Normalize the target once (`normalizeCommoditySymbol(rawBase) ?? rawBase`) and use that canonical form for both the `-X <target>` valuation and every comparison. `Kirt` → `KIRT` now values correctly (ADA = 0.169753 / 0.00568182 ≈ 29.88 KIRT; KIRT itself is the base row at 1).

## Testing

- New integration test: `listKnownPricesInBase(user, 'Kirt')` values ADA into `KIRT` and treats `KIRT` as the base row.
- Full `lib/prices` suite 109/109; type-check + lint clean.

### DIFF
--- a/lib/prices/service.ts
+++ b/lib/prices/service.ts
@@ -481,7 +481,13 @@ export class PriceService {
     userId: string,
     targetCurrency?: string
   ): Promise<KnownPrice[]> {
-    const base = targetCurrency ?? (await this.resolveBaseCurrency(userId));
+    // Normalize the target the same way held symbols are normalized (uppercased,
+    // `$`→USD). getBaseCurrency can hand back a currency in any case (e.g.
+    // `Kirt`), and every downstream `=== base` check compares against a
+    // normalized symbol — and `-X` must target the same canonical form ledger
+    // valued into — so an un-normalized base would null every row.
+    const rawBase = targetCurrency ?? (await this.resolveBaseCurrency(userId));
+    const base = normalizeCommoditySymbol(rawBase) ?? rawBase;
     const raw = await this.listKnownPrices(userId);
 
     // The base row is the target currency valued in itself, so represent it as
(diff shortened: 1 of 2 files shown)


lib/prices/service.ts
```diff
@@ -481,7 +481,13 @@ export class PriceService {
...
     targetCurrency?: string
   ): Promise<KnownPrice[]> {
-    const base = targetCurrency ?? (await this.resolveBaseCurrency(userId));
+    // Normalize the target the same way held symbols are normalized (uppercased,
+    // `$`→USD). getBaseCurrency can hand back a currency in any case (e.g.
+    // `Kirt`), and every downstream `=== base` check compares against a
+    // normalized symbol — and `-X` must target the same canonical form ledger
+    // valued into — so an un-normalized base would null every row.
+    const rawBase = targetCurrency ?? (await this.resolveBaseCurrency(userId));
+    const base = normalizeCommoditySymbol(rawBase) ?? rawBase;
```
naeemba, line 490: nit: with a mixed-case target the returned rows now carry `quote: base` in canonical form (`KIRT`), but the prices page still passes the raw `getBaseCurrency()` value as the column label (`In {baseCurrency}` in KnownPricesView renders "In Kirt"). So a valued row reads "29.88 KIRT" under an "In Kirt" header — a visible case mismatch that only appears now that these rows value instead of showing "no price". Consider normalizing the label the same way (or surfacing the canonical base back to the page) so header and row agree.

PR #77, opened 2026-07-08
fix(prices): stop the fetcher from dropping user commodity definitions

naeemba:
## Problem

An account holding **3,103 DAI** valued at **Kirt 347,424** instead of the expected **~545,952**. DAI's current price is ~176 Kirt, but the balance used ~112.

Root cause: the price fetcher owns `price-db.ledger` and rewrites it with `P` lines on every refresh. Early users kept `commodity`/`alias`/`account` declarations in that file. The first migration harvested only the `P` prices into the DB and **silently dropped every declaration** (backing the original up as `price-db_old.ledger`).

Losing `commodity KIRT / alias Kirt` makes ledger treat the fetched-price commodity `KIRT` and the journal/display commodity `Kirt` as **distinct commodities** (ledger is case-sensitive). A `-X Kirt` valuation then can't reach the fresh `KIRT` rate and falls back to a stale one:

```
-X KIRT  → 545,952.126   (correct)
-X Kirt  → 347,424.0934   (stale fallback — what the user saw)
```

Every holding in a split currency is mispriced, not just DAI.

## Fix

- **Isolate** the generated file: the fetcher now writes `generated-prices.ledger` (system-owned); `--price-db` prefers it and falls back to the legacy `price-db.ledger`. A user-authored price DB can never be clobbered again.
- **`extractDefinitions`** — pure splitter separating declarations from `P` directives.
- **`relocateLegacyDefinitions`** — moves dropped declarations into an included `definitions.ledger`, regenerates prices under the new filename, removes the superseded legacy files, verifies, and pushes. Idempotent; **request-context only** so an encrypted journal is never overwritten with plaintext (its `pull` fails without a DEK → skipped). Reachable via `POST /api/maintenance/relocate-definitions` for the caller's own journal.
- **`verifyJournalParseable`** now runs hermetically (ignores ambient `~/.ledgerrc` and `LEDGER_*`), so a personal `LEDGER_PRICE_DB` can't fail an otherwise-valid parse.
- Quota accounting and main-file detection exclude the generated file.

## After deploy

The affected user triggers the repair once (authenticated): `POST /api/maintenance/relocate-definitions` → their balance corrects to 545,952.

## Tests

- `definitions.test.ts` — splitter (6 cases)
- `service-definitions.test.ts` — relocation end to end via the in-memory store: relocate + push-to-canonical, idempotency, and a no-op for unaffected users
- Full suite: 963 passing


naeemba (commented):
Review of the definitions-relocation fix. The isolation of the generated price DB and the hermetic verify are solid. A few concerns on the relocation write ordering and the idempotency pre-check.

lib/prices/service.ts
```diff
@@ -643,4 +663,113 @@ export class PriceService {
...
+        `${extractDefinitions(source)}\n`
+      );
+      await this.prependInclude(layout.mainPath, defsPath);
+
+      // Rebuild the generated price file under its own name, then drop the
+      // superseded legacy files so the declarations aren't loaded twice
+      // (ledger aborts on a duplicate `alias`).
+      await this.regenerateUserPriceDb(userId);
+      await fs.rm(path.join(layout.dir, PRICE_DB_NAME), { force: true });
+      await fs.rm(path.join(layout.dir, PRICE_DB_OLD_NAME), { force: true });
```
naeemba: The relocation mutates the journal (writes `definitions.ledger`, prepends the `include` to the main file, regenerates prices) and then `rm`s both legacy files *before* `verifyJournalParseable` runs. There is no rollback if verify fails, and `push` never runs, so local and canonical diverge in a way a later `pull` cannot heal cleanly:

- The relocated `alias`/`commodity` can collide with a declaration the user already re-added to their journal after noticing the mispricing. The comment two lines up even notes "ledger aborts on a duplicate `alias`" — so a user who manually re-added `alias Kirt` and then triggers this repair hits exactly that: verify fails, the function throws.
- On that throw, `main.ledger` locally now carries `include ./definitions.ledger` but the manifest still holds the old etag. The next `pullToLocal` sees main unchanged (etag matches) so it keeps the modified local main, but deletes `definitions.ledger` (absent from canonical). Main now includes a file that no longer exists → every subsequent read/report fails, and it re-breaks on each retry.

Run `verifyJournalParseable` against the new layout *before* deleting the legacy files and (ideally) before mutating the committed main — or restore the pre-relocation state on failure — so a failed verify is a clean no-op rather than a persistently broken journal.

lib/prices/service.ts
```diff
@@ -643,4 +663,113 @@ export class PriceService {
...
+   * lock/pull entirely for the common case (already relocated, or nothing to
+   * relocate), so an unaffected user never pays for a mirror-pull.
+   */
+  async relocateLegacyDefinitions(
+    userId: string
+  ): Promise<'relocated' | 'skipped'> {
+    const dir = getJournalDir(userId);
+    // Cheap local pre-checks outside the lock: already relocated, or nothing to
+    // relocate. Both avoid the destructive mirror-pull for unaffected users.
+    if (await fileExists(path.join(dir, DEFINITIONS_NAME))) return 'skipped';
```
naeemba: This pre-check treats the mere existence of `definitions.ledger` as "already relocated" and returns `skipped`. But `definitions.ledger` is a common hand-authored split in ledger journals (accounts.ledger / commodities.ledger / definitions.ledger). A user who already keeps such a file AND still has dropped declarations in a legacy `price-db.ledger` will silently never be repaired. Consider gating on a marker we own (e.g. our banner) rather than the filename, or checking whether the legacy source still has un-relocated definitions before skipping.

app/api/maintenance/relocate-definitions/route.ts
```diff
@@ -0,0 +1,20 @@
...
+ * One-shot repair: relocate commodity/account declarations that an early price
+ * migration dropped out of the fetcher-owned price DB into an included
+ * `definitions.ledger`. Operates only on the caller's own journal. Idempotent —
+ * safe to hit more than once; returns `skipped` once there is nothing to move.
+ *
+ * Pulls first so the local cache (and, for an encryption-enabled caller, the
+ * session DEK) is current before the relocation reads its source files.
+ */
+export async function POST(): Promise<NextResponse> {
+  const user = await requireUser();
```
naeemba, line 20: `relocateLegacyDefinitions` can throw (the verify-failure path throws `Error('definitions relocation left journal unparseable: ...')`), and this handler has no try/catch, so a repair that leaves an unparseable journal surfaces as an unstructured 500. Wrap it and return a structured error/status so the caller can distinguish `relocated` / `skipped` / `failed`.

app/api/maintenance/relocate-definitions/route.ts
```diff
@@ -0,0 +1,20 @@
...
+  await pullLocked(user.id);
```
naeemba: nit: the handler `pullLocked`s, then `relocateLegacyDefinitions` re-`pull`s inside its own `withUserLock` — two full mirror-pulls (and two lock acquisitions) per call. The route-level pull is only needed so the service's cheap pre-checks read current local state; the authoritative pull already happens inside the lock. Minor redundant round-trip, but worth collapsing.

lib/prices/service.ts
naeemba: Fixed in c286965. The verify now runs against the new layout *before* the legacy files are deleted, and on failure the pre-relocation state is restored: `main.ledger` is written back to its captured original (dropping the just-prepended include), and `definitions.ledger` is removed (or restored to its prior hand-authored content) before the error is re-thrown. `push` never runs on that path, so the legacy files remain the untouched source of truth and a failed verify is a clean, fully reversible no-op rather than a persistently broken journal.

lib/prices/service.ts
naeemba: Fixed in c286965. "Already relocated" is now determined by our own banner in `definitions.ledger` (via `hasGeneratedBanner`), not by the filename. A hand-authored `definitions.ledger` (no banner) is no longer treated as done: we append the recovered declarations to it instead of overwriting, so the user's own split survives, and if that produces a duplicate declaration the verify catches it and rolls back. Idempotency for our own file is preserved by the banner check (and, after a successful relocation, the legacy source is gone so the source pre-check also short-circuits).

app/api/maintenance/relocate-definitions/route.ts
naeemba: Fixed in 078c111. The handler now wraps the call and returns `{ result: 'failed', message }` with a 500 on throw, so the caller can distinguish `relocated` / `skipped` / `failed`. The message on the verify-failure path is already path-redacted by `verifyJournalParseable`.

app/api/maintenance/relocate-definitions/route.ts
naeemba: Fixed in c286965 (service) and 078c111 (route). The route-level pull is gone and the outside-lock pre-checks it existed to feed were removed; the service now takes the lock once and pulls once inside it, doing all checks against that fresh state. So it's a single lock acquisition and a single mirror-pull per call instead of two.

lib/prices/service.ts
```diff
@@ -643,4 +660,146 @@ export class PriceService {
...
+      // user split we must never clobber; we append to it below instead.
+      const existingDefs = await this.deps.journalRepo
+        .readFile(defsPath)
+        .catch(() => null);
+      if (existingDefs !== null && hasGeneratedBanner(existingDefs)) {
+        return 'skipped';
+      }
+
+      const source = await this.readDefinitionsSource(layout.dir);
+      if (!source) return 'skipped';
```
naeemba, line 712: The two branches added by the fix commits — the verify-failure rollback (this block: restore `definitions.ledger`/`main.ledger`, leave legacy files intact, re-throw) and the append-to-a-hand-authored-`definitions.ledger` path (the `existingDefs !== null && existingDefs.trim()` branch above) — are the crux of round-1 feedback but have no test coverage. `service-definitions.test.ts` only exercises relocate/idempotent/skip.

Worth adding: (1) seed a journal whose main already declares a commodity that collides with the extracted `alias`, assert the call throws, `definitions.ledger` is gone (or restored to its prior content), `main.ledger` no longer carries the prepended `include`, both legacy files still exist, and nothing was pushed; (2) seed a hand-authored `definitions.ledger` (no banner) plus a legacy source, assert the recovered declarations are appended (user split preserved) rather than overwritten.

naeemba: Added in 10eab4d. Two tests now cover the branches:

1. `rolls back cleanly when the new layout fails to parse` — seeds a main journal that already re-declares `commodity KIRT` / `alias Kirt` (the "user re-added after noticing the mispricing" case). Relocating the same `alias` produces a duplicate that ledger aborts on, so verify fails. Asserts the call throws, `definitions.ledger` is gone, `main.ledger` is restored (prepended `include` dropped), both legacy files still exist, and nothing was pushed (wipe-and-repull reproduces the pre-relocation canonical state).

2. `appends to a hand-authored definitions.ledger instead of overwriting it` — seeds a no-banner `definitions.ledger` (`commodity BTC`) plus a legacy source. Asserts the recovered declarations are appended after the user's own (BTC precedes KIRT), the user split survives, no banner is stamped, and a second call is a no-op.

Note: the collision has to be a duplicate `alias` (not a colliding commodity name) — ledger 3.4.1 tolerates `commodity Kirt` alongside `alias Kirt`, but aborts on a second `alias Kirt`.

app/api/maintenance/relocate-definitions/route.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+export async function POST(): Promise<NextResponse> {
+  const user = await requireUser();
+  try {
+    const result = await priceService.relocateLegacyDefinitions(user.id);
+    return NextResponse.json({ result });
+  } catch (error) {
+    const message =
+      error instanceof Error ? error.message : 'definitions relocation failed';
+    return NextResponse.json({ result: 'failed', message }, { status: 500 });
```
naeemba: The catch-all returns the raw `error.message` to the client for *every* throw, not just the verify-failure path. The reply to the earlier thread noted the verify message is path-redacted by `verifyJournalParseable` — but that only covers one path. Any other throw out of `relocateLegacyDefinitions` (an `fs` error from `writeFileAtomic`/`fs.rm`, a DB error from `commodityRepo.insert`, a `push` failure) surfaces its message verbatim, and `fs` errors embed the absolute server path (e.g. `ENOENT: ... '/data/journals/<id>/...'`). That leaks internal filesystem layout.

The sibling route `app/api/account/export/route.ts` is the canonical pattern here: it `log.error({ err: e }, ...)`s the real error and returns a generic redacted string. This handler neither logs nor redacts. Suggest: log the error and return a generic message while keeping the `result: 'failed'` discriminant (e.g. `{ result: 'failed', message: 'Could not repair your journal' }`).

naeemba: Fixed in 4eea30b. The handler now follows the export-route pattern: it logs the real error server-side (`log.error({ err: error }, ...)`) and returns a generic `{ result: 'failed', message: 'Could not repair your journal' }` for every throw. No error message reaches the client, so fs/DB/push failures can no longer leak the absolute journal path, while the structured `failed` discriminant is preserved so callers can still tell it apart from relocated/skipped.

lib/prices/service.ts
```diff
@@ -643,4 +660,146 @@ export class PriceService {
...
+    defsPath: string
+  ): Promise<void> {
+    const main = await this.deps.journalRepo.readFile(mainPath).catch(() => '');
+    let rel = path
+      .relative(path.dirname(mainPath), defsPath)
+      .split(path.sep)
+      .join('/');
+    if (!rel.startsWith('.')) rel = `./${rel}`;
+    const directive = `include ${rel}`;
+    if (main.split('\n').some((line) => line.trim() === directive)) return;
```
naeemba: nit: `prependInclude` only recognizes an existing include when the line reads exactly `include ./definitions.ledger`. A user with a hand-authored split who wrote `include definitions.ledger` (bare, no `./`) is on the append path, so this prepends a second, `./`-prefixed include of the same file. Ledger then loads `definitions.ledger` twice, the relocated `alias` collides, verify fails and rolls back — that user can never self-repair via this endpoint (fails cleanly, but permanently). Consider normalizing the comparison (resolve each include target to a path) rather than matching the exact directive string.

PR #78, opened 2026-07-08
fix(prices): make relocated `format` directives journal-valid

naeemba:
Follow-up to #77.

## Problem

`relocateLegacyDefinitions` copied `commodity` blocks verbatim from the legacy price DB. A hand-maintained price DB commonly writes a **symbol-less `format` sample** (e.g. `format 1,000`). ledger accepts that inside a `--price-db` but **rejects it in a journal include**:

```
commodity directive symbol KIRT and format directive symbol '' should be the same
```

So the post-relocation `verifyJournalParseable` failed, the relocation rolled back, and the repair could never complete — `POST /api/maintenance/relocate-definitions` returned `failed`. (The rollback worked correctly; no journal was damaged.)

## Fix

`extractDefinitions` now gives a numeric-only `format` sample its enclosing commodity, **quoted** so a symbol containing a separator (e.g. `د.إ`, whose `.` would otherwise be mis-tokenized) matches its `commodity` directive:

```
format 1,000        →  format 1,000 "KIRT"
format 1,000.00     →  format 1,000.00 "د.إ"
```

A sample that already carries its symbol (`format KIRT 1,000` / `format $1,000.00`) is left untouched.

## Verification

Reproduced against the **real** 211-line `price-db_old.ledger` from prod: after the fix the relocated `definitions.ledger` passes `ledger stats` (exit 0) and values cleanly, where before every `format` line failed. Unit tests cover symbol injection, the separator-quoting case, and the already-symboled no-op.

### DIFF
--- a/lib/prices/definitions.ts
+++ b/lib/prices/definitions.ts
@@ -22,11 +22,43 @@ const PRICE_LINE =
  */
 export const extractDefinitions = (text: string): string => {
   const kept: string[] = [];
+  let currentCommodity: string | null = null;
   for (const raw of text.split('\n')) {
     const line = raw.replace(/\s+$/, '');
     const trimmed = line.trim();
     if (PRICE_LINE.test(trimmed)) continue;
     if (trimmed.includes(BANNER_MARKER)) continue;
+
+    // Track the enclosing `commodity <symbol>` block. A non-indented, non-blank
+    // line closes it; blank lines and comments leave it open.
+    const commodityMatch = /^commodity\s+(.+)$/.exec(trimmed);
+    if (commodityMatch) {
+      // Store unquoted so we can re-quote consistently in the format below.
+      currentCommodity = commodityMatch[1].trim().replace(/^"(.*)"$/, '$1');
+      kept.push(line);
+      continue;
+    }
+    if (trimmed && !/^\s/.test(line)) currentCommodity = null;
+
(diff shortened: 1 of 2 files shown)


lib/prices/definitions.ts
```diff
@@ -22,11 +22,43 @@ const PRICE_LINE =
...
   const kept: string[] = [];
+  let currentCommodity: string | null = null;
   for (const raw of text.split('\n')) {
     const line = raw.replace(/\s+$/, '');
     const trimmed = line.trim();
     if (PRICE_LINE.test(trimmed)) continue;
     if (trimmed.includes(BANNER_MARKER)) continue;
+
+    // Track the enclosing `commodity <symbol>` block. A non-indented, non-blank
+    // line closes it; blank lines and comments leave it open.
```
naeemba: nit: This comment contradicts itself for a top-level comment line. It says "A non-indented, non-blank line closes it; blank lines and comments leave it open" — but a column-0 comment (`; foo`) is a non-indented, non-blank line, so the reset on line 41 (`trimmed && !/^\s/.test(line)`) fires and closes the block. Only *indented* comments leave it open. That happens to match ledger's own block semantics (a column-0 line ends the commodity block), so the behavior is fine — just reword to "indented comments and blank lines leave it open" to avoid implying column-0 comments are preserved inside the block.

PR #79, opened 2026-07-08
fix(prices): make relocated definitions layout parse under the read path

naeemba:
## Summary

The definition-relocation feature (#77) could leave a journal that `verifyJournalParseable` accepted but the dashboard couldn't read, returning a 500 with a ledger `pool.cc:110` assertion. This branch makes the relocated layout valid under the actual read path.

## The bug

ledger parses `--price-db` **before** the main journal and auto-creates a commodity from each `P` line's symbol/quote. When the generated price DB named a commodity the journal also **aliases**, the later `alias` directive collided with that auto-created commodity and aborted every read:

- base currency: price quotes `USD` while the journal declares `commodity $` / `alias USD`
- symbol alias: price symbol `BTC` while the journal declares `commodity BITCOIN` / `alias BTC`

It only surfaced with the price DB loaded. `verifyJournalParseable` ran `ledger stats` **without** `--price-db`, so it never reproduced the collision and pushed the broken layout.

## Changes

- **Canonicalize generated prices** — `regenerateUserPriceDb` maps every price symbol/quote through the aliases declared in `definitions.ledger` (new `parseAliasMap`), at render time, so the price DB always names the canonical commodity. Covers both cron-fetched and relocated rows; the base-currency quote stays intact for `listForQuote`'s filter.
- **Verify like the read path** — `verifyJournalParseable(mainPath, priceDbPath?)` loads the price DB, so a definitions/price-db collision fails the verify before any push.
- **Hermetic ledger invocation** — `runLedgerForUser` now passes `--init-file /dev/null` and strips ambient `LEDGER_*`; adds an `includePriceDb` option so commodity listing skips the price DB (a broken one can no longer zero the symbol set during a repair).
- **Repair vehicle** — `repairUserPriceDb` + `POST /api/maintenance/repair-price-db` regenerate, verify, and push a corrected price DB for the caller, encrypting with the session DEK. Repairs journals written before this fix (idempotent).
- Earlier commits on the branch: relocated `format` directives made journal-valid, and a commodity-block comment correction.

## Repair for an already-broken journal

After deploy, the affected user (authenticated) hits `POST /api/maintenance/repair-price-db` once → `{result:"repaired"}`. Works for encryption-enabled accounts because it runs in the user's session.

## Tests

- `parseAliasMap` unit coverage (indented block aliases, quoting, account-alias/block-close exclusion).
- Relocation with a symbol-alias legacy price (`BTC`→`BITCOIN`).
- The base-currency-alias case (`USD`→`$`) end-to-end via `repairUserPriceDb`.
- Full suite green (975 tests).

### DIFF
--- a/lib/prices/service.ts
+++ b/lib/prices/service.ts
@@ -140,8 +140,15 @@ export class PriceService {
       await this.listNormalizedSymbolsForUser(userId)
     );
     // Match on the canonical symbol: a stored alias-symbol row still belongs to
-    // a journal that only ever names the canonical commodity.
-    const fetched = all.filter((r) => userSymbols.has(canonical(r.symbol)));
+    // a journal that only ever names the canonical commodity. Normalize the
+    // canonical name before the membership test — `userSymbols` is upper-cased
+    // (listNormalizedSymbolsForUser), while canonical() preserves the journal's
+    // raw case, so a lower/mixed-case declaration (`commodity Bitcoin` / `alias
+    // BTC`) would otherwise never match and drop the row silently.
+    const fetched = all.filter((r) => {
+      const canonicalSymbol = normalizeCommoditySymbol(canonical(r.symbol));
+      return canonicalSymbol !== null && userSymbols.has(canonicalSymbol);
+    });
     const manual = await this.deps.manualRepo.listForUser(userId);
     const manualRows: CommodityPriceRow[] = manual.map((m) => ({
       symbol: m.symbol,
--- a/lib/prices/service.ts
+++ b/lib/prices/service.ts
@@ -53,8 +53,12 @@ import {
   PRICE_DB_NAME,
   getJournalCacheTag,
 } from '@/lib/journal/layout';
+import { resolveIncludes } from '@/lib/journal/loader';
 import { withUserLock } from '@/lib/journal/mutex';
-import type { JournalRepository } from '@/lib/journal/repository';
+import type {
+  JournalLayout,
+  JournalRepository,
+} from '@/lib/journal/repository';
 import { verifyJournalParseable } from '@/lib/journal/verify';
 import { createLogger } from '@/lib/log';
 import { pull, push } from '@/lib/storage';
@@ -128,7 +132,7 @@ export class PriceService {
     // canonicalize at render, not at insert, so the base-currency quote stays
     // intact for listForQuote's filter above and cron-fetched rows are covered
     // too. See parseAliasMap and lib/journal/verify.ts.
-    const aliasMap = await this.readCommodityAliasMap(layout.dir);
+    const aliasMap = await this.readCommodityAliasMap(layout);
     const canonical = (symbol: string): string =>
       aliasMap.get(symbol) ?? symbol;
 
@@ -168,19 +172,44 @@ export class PriceService {
   }
 
   /**
-   * Alias→canonical-commodity map declared in the user's `definitions.ledger`
-   * (empty when there is none). This is where the relocation puts recovered
-   * `commodity`/`alias` declarations, so it is the authoritative source for
-   * canonicalizing generated prices. Aliases declared elsewhere in the journal
-   * are not consulted; move them into definitions.ledger if they must apply.
+   * Alias→canonical-commodity map for the whole journal (empty when it declares
+   * none). ledger honors `commodity`/`alias` blocks wherever they appear — a
+   * common layout declares `commodity $` / `alias USD` directly in the main
+   * journal, never in `definitions.ledger` — so consulting only the relocated
+   * definitions file would miss them and let a generated `P … USD` collide with
+   * the journal's `alias USD` (the pool.cc assertion this canonicalization
+   * exists to prevent). We therefore parse every file reachable from the main
+   * journal via its `include` directives, the same set ledger loads.
+   *
+   * `definitions.ledger` is read explicitly as well so the map still resolves if
+   * include-resolution fails (an include cycle or a missing include, which would
+   * make ledger itself fail to parse anyway). Later files win on a duplicate
+   * alias, but a genuine duplicate `alias` would abort ledger regardless.
    */
   private async readCommodityAliasMap(
-    dir: string
+    layout: JournalLayout
   ): Promise<Map<string, string>> {
-    const text = await this.deps.journalRepo
-      .readFile(path.join(dir, DEFINITIONS_NAME))
-      .catch(() => '');
-    return parseAliasMap(text);
+    const defsPath = path.join(layout.dir, DEFINITIONS_NAME);
+    const files = [defsPath];
+    try {
+      for (const file of await resolveIncludes(layout.mainPath)) {
+        // definitions.ledger is already first in the list; skip the duplicate.
+        if (path.resolve(file) !== path.resolve(defsPath)) files.push(file);
+      }
+    } catch {
+      // Include cycle or missing include — ledger would fail to parse too. Fall
+      // back to the definitions.ledger declarations collected above.
+    }
+
+    const map = new Map<string, string>();
+    for (const file of files) {
+      const text = await this.deps.journalRepo.readFile(file).catch(() => '');
+      if (!text) continue;
+      for (const [alias, commodity] of parseAliasMap(text)) {
+        map.set(alias, commodity);
+      }
+    }
+    return map;
   }
 
   /**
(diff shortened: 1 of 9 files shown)


lib/prices/service.ts
```diff
@@ -115,10 +119,25 @@ export class PriceService {
...
+
+    // Canonicalize every price symbol/quote through the aliases the journal
+    // declares. ledger auto-creates a commodity from a `P` line's symbol/quote,
+    // so a price that names an alias — quote `USD` when the journal declares
+    // `commodity $ / alias USD`, or symbol `BTC` aliased to `BITCOIN` — collides
+    // with the `alias` directive and aborts every read (pool.cc assertion). We
+    // canonicalize at render, not at insert, so the base-currency quote stays
+    // intact for listForQuote's filter above and cron-fetched rows are covered
+    // too. See parseAliasMap and lib/journal/verify.ts.
+    const aliasMap = await this.readCommodityAliasMap(layout.dir);
```
naeemba: The canonicalization only consults `definitions.ledger` (`readCommodityAliasMap` reads `DEFINITIONS_NAME` alone), but ledger honors `commodity`/`alias` blocks wherever they appear in the journal. A very common real-world layout declares `commodity $` / `alias USD` directly in the main journal (or any hand-authored include), never in `definitions.ledger` and never touched by relocation. For such a user, `listForQuote("USD")` returns rows quoting `USD`, `canonical("USD")` no-ops (empty map), and the generated DB emits `P ... USD` — which auto-vivifies commodity `USD` and collides with the journal `alias USD`, aborting every read with the exact pool.cc assertion this PR sets out to eliminate.

Worse, this reintroduces on the two paths that have no verify gate: the daily cron (`runOnce` -> `regenerateUserPriceDb`, no `verifyJournalParseable`) and manual-price edits (`addManualPrices`/`deleteManualPrice` -> `regenerateUserPriceDb`). Only `repairUserPriceDb`/`relocateLegacyDefinitions` verify — and `repairUserPriceDb` reads the same `definitions.ledger`-only map, so it cannot fix such a user either.

Consider deriving the alias map from the whole journal (e.g. parse `ledger commodities`/the resolved journal) rather than the single `definitions.ledger` file, so canonicalization matches what ledger actually resolves.

lib/prices/service.ts
```diff
@@ -132,7 +151,12 @@ export class PriceService {
     const merged = [...fetched, ...manualRows].sort(
       (a, b) => a.fetchedAt.getTime() - b.fetchedAt.getTime()
     );
-    const body = renderPriceDb(merged);
+    const canonicalized = merged.map((r) => ({
+      ...r,
+      symbol: canonical(r.symbol),
+      quote: canonical(r.quote),
```
naeemba, line 168: nit: `canonical()` can substitute a commodity name that requires quoting. `parseAliasMap` deliberately unquotes the `commodity` target (it handles `commodity "د.إ"`), so `aliasMap.get(sym)` can return a symbol containing whitespace or special characters (e.g. `commodity "Real Estate" / alias RE`). `renderPriceDb` emits it bare — `P DATE SYM PRICE Real Estate` — which ledger cannot parse. Fetched rows never trigger this (their symbols are already canonical tickers), but a canonical substitution can. On repair/relocate the verify catches it (repair silently returns `failed`), but cron/manual regeneration would push the unparseable DB. If a canonical symbol needs quoting, `renderPriceDb` should re-quote it.

lib/prices/service.ts
naeemba: Fixed in 1c59413. `readCommodityAliasMap` now parses every file reachable from the main journal via its `include` directives (the same set ledger loads), instead of only `definitions.ledger`. So a user who declares `commodity $` / `alias USD` in the main journal (or any hand-authored include, never relocated) now gets a non-empty map, `canonical("USD")` resolves to `$`, and the generated DB no longer emits `P … USD` to collide with the `alias USD`. This covers the cron and manual-price paths too, since the canonicalization happens in `regenerateUserPriceDb`. `definitions.ledger` is still read explicitly as a fallback if include-resolution throws (cycle / missing include), which is a state where ledger itself would fail to parse anyway.

lib/prices/service.ts
naeemba: Fixed in 67ab585. `renderPriceDb` now re-quotes a symbol or quote when it carries whitespace, a digit, or ledger's reserved amount-syntax punctuation, so a canonical substitution like `commodity "Real Estate"` / `alias RE` (or `د.إ`) emits `P DATE "Real Estate" …` rather than the unparseable bare form. Bare-safe tickers (`$`, `BTC`, `USD`) are untouched, so fetched/manual rows render exactly as before. Added formatter tests covering both the quoted and bare-safe cases.

lib/prices/service.ts
```diff
@@ -115,10 +123,25 @@ export class PriceService {
...
+    const canonical = (symbol: string): string =>
+      aliasMap.get(symbol) ?? symbol;
+
     const userSymbols = new Set(
       await this.listNormalizedSymbolsForUser(userId)
     );
-    const fetched = all.filter((r) => userSymbols.has(r.symbol));
+    // Match on the canonical symbol: a stored alias-symbol row still belongs to
+    // a journal that only ever names the canonical commodity.
+    const fetched = all.filter((r) => userSymbols.has(canonical(r.symbol)));
```
naeemba: `canonical(r.symbol)` returns the commodity name exactly as written in the journal (`parseAliasMap`/`extractDefinitions` only unquote — they never upper-case), but `userSymbols` is the normalized, upper-cased set from `listNormalizedSymbolsForUser` (`normalizeCommoditySymbol` upper-cases everything). So the membership test mixes raw-case against an upper-cased set.

For the common aliases this happens to line up because the canonical commodity is already upper-case (`BITCOIN`, `KIRT`) — which is exactly what the `gina` test exercises. But a perfectly legal lower/mixed-case declaration breaks it: `commodity Bitcoin` / `alias BTC` with a stored row `symbol='BTC'` yields `canonical('BTC') = 'Bitcoin'`, while `ledger commodities` → `Bitcoin` → normalized `'BITCOIN'`. `userSymbols.has('Bitcoin')` is false, so the row is silently dropped and that commodity gets no price line. Consider comparing on the normalized canonical (e.g. `normalizeCommoditySymbol(canonical(r.symbol))`) so both sides are in the same space.

naeemba: Fixed in eeea8f7. The membership test now normalizes the canonical name before the lookup: `userSymbols.has(normalizeCommoditySymbol(canonical(r.symbol)))` (with a null guard). `userSymbols` is the upper-cased set from `listNormalizedSymbolsForUser`, while `canonical()` preserves the journal's raw case, so a lower/mixed-case declaration like `commodity Bitcoin` / `alias BTC` (stored row `symbol='BTC'`) previously yielded `canonical('BTC') = 'Bitcoin'`, missed `'BITCOIN'` in the set, and silently dropped the row. Both sides now compare in the normalized space. The upper-case aliases (`BITCOIN`, `KIRT`) are unaffected since they were already normalized.

naeemba: Follow-up (additive, code fix confirmed and this thread resolved): the `normalizeCommoditySymbol(canonical(r.symbol))` normalization has no regression test. Both new integration tests use already-uppercase canonical commodities — gina `commodity BITCOIN / alias BTC` (canonical `BITCOIN`) and hank `commodity KIRT` / `commodity $` (canonical `KIRT`/`$`) — so the normalize wrapper is a no-op in both; deleting it would not fail any test. The guarded case (a mixed-case `commodity Bitcoin / alias BTC`, stored `symbol='BTC'` → canonical `Bitcoin` must match the upper-cased `BITCOIN` in `userSymbols`, else the row is silently dropped and that commodity gets no price line) is the exact data-loss path this round fixed. Add a regenerate test with a mixed-case canonical commodity so a future refactor cannot silently regress it.

naeemba: Added in 7e575ad. New regenerate test in service-definitions.test.ts seeds a mixed-case canonical (`commodity Bitcoin` / `alias BTC`) with a stored row `symbol='BTC'`, so `canonical('BTC') = 'Bitcoin'` must normalize to `BITCOIN` to match the upper-cased `userSymbols`. It asserts the row survives and renders as `P … Bitcoin 50000 USD` (never the alias). Verified it fails when the `normalizeCommoditySymbol(canonical(...))` wrapper is removed — the row is dropped and no price line is emitted — so a future refactor can no longer silently regress this path.

PR #80, opened 2026-07-09
feat(entry): add extra items to guided transaction forms

naeemba:
## What

Adds a dynamic **extra items** list to the four guided transaction-entry forms (Expense, Income, Transfer, Exchange), so a single transaction can carry an arbitrary number of extra postings — a tip on a meal, a shipping fee on a purchase, a broker fee on a trade, a processor cut on income. FixBalance is untouched.

Previously these guided forms compiled to exactly two postings; extra lines were only reachable by dropping to the Form/Raw lens.

## How

- **`types/extraItems.ts`** — shared helper: `ExtraItem`, `formatAmount`, `residualByCurrency`, `balancingPostings`, `extraItemPostings`, `toExtraItems`, `singleAccount`.
- **Each adapter** gains `extraItems: ExtraItem[]`. `compile` appends the extras and lets the form's absorbing account balance them — one posting per currency (handles the `@@` cost path for Exchange). `detect` recovers extras by partitioning postings by account role.
- **`ExtraItemsField.tsx`** — dynamic row component (account + amount + currency + remove; add button capped at `MAX_POSTINGS`), wired into all four forms.

### Multi-currency example (Expense)

```
Expenses:Dining   100 USD
Expenses:Tips      20 USD
Expenses:Fees       2 EUR
Assets:Checking  -120 USD
Assets:Checking    -2 EUR
```

## Invariants

- **Regression-safe:** when `extraItems` is empty, `compile` output is byte-identical to before (explicit empty-extras branch using the prior string form — computed absorbing postings never rewrite existing users' ledger output).
- **Deterministic detect:** returns null unless the whole draft balances, the absorbing account is a single account, and there are no stray account roles. Cross-type detection is locked in by tests (a transfer-with-fee stays Transfer, income-with-fee stays Income, etc.).

## Notes

- Type detection now recognizes plain asset↔liability 2-posting transactions (e.g. a credit-card payment) as **Transfer** — previously they fell through to the Form fallback. Non-destructive (round-trips identically); consistent with the Transfer/Income forms already offering liability accounts.
- Extra-item account picker is scoped to expense-role accounts (matching what `detect` recovers).

## Testing

- Full suite: **977/977** green; entry suite **130/130**; `tsc` and `eslint` clean.
- New coverage: helper functions, per-adapter compile/detect round-trips (incl. multi-currency and exchange cost), cross-type detection ordering, and the `ExtraItemsField` component.

### DIFF
--- a/features/transactions/entry/typeForms/ExtraItemsField.tsx
+++ b/features/transactions/entry/typeForms/ExtraItemsField.tsx
@@ -3,7 +3,7 @@
 import React from 'react';
 import AmountInput from '../../AmountInput';
 import type { ExtraItem } from '../types/extraItems';
-import { CurrencyCombobox, SectionLabel } from './fields';
+import { CurrencyCombobox, optionsForRoles, SectionLabel } from './fields';
 import Combobox from '@/components/Combobox';
 import { Button } from '@/components/ui/button';
 
@@ -23,6 +23,7 @@ export function ExtraItemsField({
   onChange: (items: ExtraItem[]) => void;
 }): React.JSX.Element {
   const atCap = baseCount + items.length >= MAX_POSTINGS;
+  const expenseAccounts = optionsForRoles(accounts, 'expense');
 
   const setItem = (index: number, patch: Partial<ExtraItem>) =>
     onChange(
@@ -49,7 +50,7 @@ export function ExtraItemsField({
           <Combobox
             value={item.account}
             onChange={(account) => setItem(index, { account })}
-            options={accounts}
+            options={expenseAccounts}
             placeholder="Account, e.g. Expenses:Fees"
           />
           <AmountInput
--- a/features/transactions/entry/typeForms/ExtraItemsField.tsx
+++ b/features/transactions/entry/typeForms/ExtraItemsField.tsx
@@ -26,15 +26,19 @@ export function ExtraItemsField({
 
   const setItem = (index: number, patch: Partial<ExtraItem>) =>
     onChange(
-      items.map((item, i) => (i === index ? { ...item, ...patch } : item))
+      items.map((item, itemIndex) =>
+        itemIndex === index ? { ...item, ...patch } : item
+      )
     );
-  const addItem = () =>
+  const addItem = () => {
+    if (atCap) return;
     onChange([
       ...items,
       { account: '', amount: '', currency: defaultCurrency },
     ]);
+  };
   const removeItem = (index: number) =>
-    onChange(items.filter((_, i) => i !== index));
+    onChange(items.filter((_, itemIndex) => itemIndex !== index));
 
   return (
     <section className="flex flex-col gap-3">
--- a/features/transactions/entry/typeForms/ExtraItemsField.tsx
+++ b/features/transactions/entry/typeForms/ExtraItemsField.tsx
@@ -1,0 +1,93 @@
+'use client';
+
+import React from 'react';
+import AmountInput from '../../AmountInput';
+import type { ExtraItem } from '../types/extraItems';
+import { CurrencyCombobox, SectionLabel } from './fields';
+import Combobox from '@/components/Combobox';
+import { Button } from '@/components/ui/button';
+
+const MAX_POSTINGS = 50;
+
+export function ExtraItemsField({
+  items,
+  accounts,
+  defaultCurrency,
+  baseCount,
+  onChange,
+}: {
+  items: ExtraItem[];
+  accounts: string[];
+  defaultCurrency: string;
+  baseCount: number;
+  onChange: (items: ExtraItem[]) => void;
+}): React.JSX.Element {
+  const atCap = baseCount + items.length >= MAX_POSTINGS;
+
+  const setItem = (index: number, patch: Partial<ExtraItem>) =>
+    onChange(
+      items.map((item, i) => (i === index ? { ...item, ...patch } : item))
+    );
+  const addItem = () =>
+    onChange([
+      ...items,
+      { account: '', amount: '', currency: defaultCurrency },
+    ]);
+  const removeItem = (index: number) =>
+    onChange(items.filter((_, i) => i !== index));
+
+  return (
+    <section className="flex flex-col gap-3">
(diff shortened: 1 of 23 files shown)


features/transactions/entry/typeForms/ExtraItemsField.tsx
```diff
@@ -0,0 +1,98 @@
...
+  };
+  const removeItem = (index: number) =>
+    onChange(items.filter((_, itemIndex) => itemIndex !== index));
+
+  return (
+    <section className="flex flex-col gap-3">
+      <SectionLabel>Extra items (fees, tips…)</SectionLabel>
+
+      {items.map((item, index) => (
+        <div key={index} className="flex items-center gap-2">
```
naeemba: Using the array index as the React `key` for a list that supports mid-list removal causes state to be misassociated. `Combobox` keeps its own uncontrolled internal state (`const [open] = useState(false)` and `const [search] = useState('')` in `components/Combobox/Combobox.tsx`) that is not derived from `item.account`. When a middle row is removed, every following row shifts down one index; React reuses the component instance bound to that index, so the shifted row inherits the previous row's open/typed-query state (e.g. a half-typed "Groc" filter or an open dropdown) even though its `value` prop updates. Give each row a stable identity (a per-row id, or `crypto.randomUUID()` captured when the row is created) and key on that instead of `index`.

features/transactions/entry/typeForms/ExtraItemsField.tsx
```diff
@@ -0,0 +1,98 @@
...
+  baseCount,
+  onChange,
+}: {
+  items: ExtraItem[];
+  accounts: string[];
+  defaultCurrency: string;
+  baseCount: number;
+  onChange: (items: ExtraItem[]) => void;
+}): React.JSX.Element {
+  const atCap = baseCount + items.length >= MAX_POSTINGS;
```
naeemba: The cap counts only extra rows (`baseCount + items.length`), but `compile` emits one balancing posting per distinct currency via `balancingPostings`, so a multi-currency draft produces more postings than `items.length`. With extras spread across several currencies the compiled transaction can exceed `MAX_POSTINGS = 50` (the `lib/transactions/schema.ts` limit) while the UI still reported spare capacity, and submission then fails validation. Edge case, but the plan's "total posting count stays bounded by 50" invariant is what the cap is supposed to guarantee — consider accounting for the currencies in play (or at least the worst case) when computing `atCap`.

features/transactions/entry/types/extraItems.ts
```diff
@@ -0,0 +1,78 @@
+import type { Posting } from '@/lib/transactions/posting';
+
+export type ExtraItem = { account: string; amount: string; currency: string };
+
+/** Format a computed numeric amount to a compact string (trailing zeros trimmed, no -0, no scientific notation). */
+export const formatAmount = (n: number): string => {
```
naeemba, line 11: nit: the doc comment promises "no scientific notation", but `Number.prototype.toFixed(10)` returns exponential form for magnitudes >= 1e21 (e.g. `(1e21).toFixed(10)` -> `"1e+21"`), and the trailing-zero regex won't strip it. Not reachable for realistic currency sums, so the string is fine — just tighten the comment or clamp if you want the guarantee to hold literally.

features/transactions/entry/types/extraItems.ts
```diff
@@ -0,0 +1,78 @@
...
+  const out: Posting[] = [];
+  for (const [currency, total] of net) {
+    if (Math.abs(total) <= 1e-9) continue;
+    out.push({ account, amount: formatAmount(-total), currency });
+  }
+  return out;
+};
+
+/** Map extra-item rows to postings, dropping fully-blank rows. */
+export const extraItemPostings = (items: readonly ExtraItem[]): Posting[] =>
```
naeemba, line 66: nit: a row with an account filled in but a blank amount survives the filter (the predicate is OR), so `extraItemPostings` emits a posting with `amount: ''`. `residualByCurrency` treats it as 0, so the computed balancing postings ignore it, and the blank posting turns the compiled draft into `computeBalance` -> `auto-balance` rather than the intended fee line. The preview does surface the auto-balance state so it isn't silent, but a partially-entered fee row is easy to hit; consider requiring both fields (drop rows missing an amount) to match user intent.

naeemba (commented):
Extra-items feature review. Adapters compile/detect round-trip cleanly and the empty-extras regression guard is solid. A few observations below, mostly low-severity.

features/transactions/entry/typeForms/ExtraItemsField.tsx
naeemba: Fixed in 27044ee. Rows now carry a stable per-row id (crypto.randomUUID captured at creation, kept in lockstep with the items array on add/remove) and are keyed on that instead of the array index, so removing a middle row no longer lets a shifted row inherit the previous row's open dropdown or half-typed Combobox search.

features/transactions/entry/typeForms/ExtraItemsField.tsx
naeemba: Fixed in 27044ee. The cap now bounds the worst case by counting the base residual currency plus each row's currency: `baseCount - 1 + items.length + distinctResidualCurrencies >= MAX_POSTINGS`. Single-currency drafts are unchanged (`baseCount + items.length`), but a multi-currency spread now consumes capacity for the extra balancing postings so the compiled transaction stays within the 50-posting limit.

features/transactions/entry/types/extraItems.ts
naeemba: Tightened the comment in 4614cc2. It now states the plain-decimal guarantee holds for realistic currency magnitudes and notes that toFixed only switches to exponential form at |n| >= 1e21, rather than claiming "no scientific notation" universally.

features/transactions/entry/types/extraItems.ts
naeemba: Fixed in e0345d3. `extraItemPostings` now requires both fields (AND instead of OR), so a row with an account but a blank amount is dropped rather than emitting an `amount: ''` posting that silently folds into auto-balance. Added test coverage for the partial-row cases.

PR #83, opened 2026-07-09
fix(prices): show freshest price, default to ledger-valued view

naeemba:
## Problem

The Known prices tab showed stale, wrong prices. BTC displayed **107,424 DAI dated 2025-01-20 (535 days · stale)** while a fresh fetched price (`$62,513`, 2026-07-09) sat ignored.

Two causes:

1. **`latestGenuinePrice` picked the wrong row.** `ledger prices <symbol>` emits one date-ascending run *per quote commodity*, concatenated. It grabbed `points[length-1]` — the last row, which belongs to whichever quote sorts last (here a lone `@@ DAI` cost annotation from a 2025 posting), not the most recent price. Ledger orders by commodity group, not global date; our code assumed "last printed = newest."
2. **The default view showed raw journal quotes.** The `?base=base` (ledger-valued) view existed but wasn't the default, so users saw our-picked directive prices instead of ledger's valuation.

## Fix

- **`latestGenuinePrice` is now quote-aware**: collapse each quote to its current price (keeping the forward-carry change-date logic per run), then pick the freshest run. Base quote wins an equal-date tie. The price value is still ledger's `P` directive verbatim — we only select which emitted row is current.
- **Default the Known prices view to the ledger-valued path** (`-X` into the display currency): the price shown is ledger's own valuation — freshest, bridged, identical to every balance report. `?quote=original` opts into the raw per-commodity journal quote. Toggle relabeled ("In {base}" default, "Original quote" secondary).

## Who computes what

| Piece | Source |
|---|---|
| Price value | Ledger — `-X` / `market()`, freshest + bridged |
| Price date + staleness | Ledger's dated `P` directives (`ledger prices`); our code selects the latest genuine run |

Ledger has **no** accessor for the date of the price it selected — `value_date` is the valuation as-of date (today / `--now`), not the set-date (verified: `--now=2027-01-01` moved `value_date` to 2027 while the price held). So price recency stays sourced from the dated directives.

## Verification

- Reproduced against prod data: BTC now values **62,523 DAI** (62,513 `$` bridged through the fresh `$`/DAI quote), dated 2026-07-09, source fetched, **not stale**.
- 138 prices tests pass; new coverage: quote-aware selection (3 unit), base tie-break, and an end-to-end case (fresh `$` beats stale `@@ DAI`) through real ledger.
- `tsc --noEmit` + eslint clean.

### DIFF
--- a/lib/prices/knownPrices.ts
+++ b/lib/prices/knownPrices.ts
@@ -14,6 +14,11 @@ export const PRICES_FORMAT =
  * `account|quantity|commodity` line per probe holding. `quantity` is the
  * full-precision value, `commodity` is the currency it resolved to (the base
  * when a conversion path existed, otherwise the holding's own commodity).
+ *
+ * Note: ledger has no accessor for the date of the price it selected —
+ * `value_date` is the valuation "as of" date (today / `--now`), not when the
+ * price was set — so recency for a valued holding is dated separately, from the
+ * dated price directives `ledger prices` reports (see latestGenuinePrice).
  */
 export const BALANCE_BASE_FORMAT =
   '%(account)|%(quantity(scrub(display_total)))|%(commodity(scrub(display_total)))\n';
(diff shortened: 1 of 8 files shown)


lib/prices/knownPrices.ts
```diff
@@ -70,23 +75,68 @@ export const ageInDays = (dateIso: string, todayIso: string): number => {
...
+      if (run[index].price !== run[index - 1].price) {
+        changeDate = run[index].date;
+      }
+    }
+    const isBase =
+      normalizedBase !== null &&
+      (normalizeCommoditySymbol(last.quote) ?? last.quote) === normalizedBase;
+    if (
+      !best ||
+      last.date > best.lastDate ||
```
naeemba, line 139: nit: run freshness is compared on `last.date` (the last emitted row of each quote's run), but that date can be a forward-carried posting date rather than when the price was set — which is exactly why `changeDate` is computed separately below. In the base-vs-cross-quote case this resolves correctly: ledger forward-carries every active quote onto the same last posting date, so both runs tie on `last.date` and the `isBase` tie-break picks the base run. But for a commodity with **no** base-quote run and two cross-quotes (e.g. a KIRT-style holding priced only in DAI and EUR), both runs still tie on the forward-carried last date, `isBase` is false for both, and the winner falls to Map insertion order (ledger's commodity sort). That can surface the stale cross-quote's price while its `changeDate` correctly flags it stale. Not a regression (the old code was strictly worse), and it needs the narrow no-base-quote + multi-cross-quote shape — but worth a comment or a test noting the tie is order-dependent there. Consider comparing runs on `changeDate` (real price-set recency) rather than the forward-carried `last.date`.

PR #85, opened 2026-07-09
fix(cash-flow): sum monthly totals across all accounts

naeemba:
## Problem

JOURNAL-AUDIT.md finding **A1** (highest impact). `reg --monthly` emits one line **per account** per month (plus `-X` `<Adjustment>` revaluation postings). The parser keyed each row by date with `map.set`, so every month collapsed to its **alphabetically-last account** instead of the true total.

Verified against the production journal:

| Month | Shown (buggy) | Actual |
|-------|---------------|--------|
| Nov 2025 expenses | $0.89 (`Expenses:Tip`) | **$50.89** |
| May 2025 expenses | $0.70 (`Expenses:Wage`) | **$5,026.27** |

Income side had the same shape. Net effect: the cash-flow chart numbers were garbage every month.

## Fix

- Add `%A` (account) to the register format.
- Skip injected `<Adjustment>` / `<Revalued>` rows (they carry a real payee under `-X` and would pollute the total).
- Accumulate per month with `+=` instead of overwriting.

## Verification

Reproduced buggy vs fixed output against the real journal — Nov 2025 now sums `50.00 + 0.89 = $50.89`, skipping the `<Adjustment>` row. Typecheck + lint clean.


features/monthlyComparison/MonthlyComparison.utils.ts
```diff
@@ -14,13 +14,17 @@ const fetchMonthly = async (
...
   const map = new Map<string, number>();
   for (const line of stdout.split('NNN')) {
-    const [date, amount] = line.split('|').map((s) => s.trim());
-    if (date && amount) map.set(date, parseAmount(amount));
+    const [date, account, amount] = line.split('|').map((s) => s.trim());
+    if (!date || !amount) continue;
+    // `-X` injects `<Adjustment>` / `<Revalued>` postings that inherit a real
+    // payee; they must not be summed into the month's total.
+    if (account.startsWith('<')) continue;
+    map.set(date, (map.get(date) ?? 0) + parseAmount(amount));
```
naeemba: The exact bug this PR fixes (per-month collapse) is parsing logic, but it lives inline inside the async `fetchMonthly` (IO + parse mixed) with no test. Every sibling that parses `NNN`-split ledger register output extracts a pure exported function with a colocated test: `lib/payees/parse.ts` (`parsePayeeRows`) + `parse.test.ts`, `lib/netWorth/parse.ts`, `lib/balance/parsePeriodic.ts`, and `features/reconcile/Reconcile.utils.ts` (`parseReconcileRows`) + `Reconcile.utils.test.ts`. Consider extracting the stdout->`Map<string, number>` parse into a pure `parseMonthly(stdout)` function and adding a `MonthlyComparison.utils.test.ts` case covering multiple accounts per month + a skipped `<Adjustment>` row. Without a test, this regression is unguarded and the fix diverges from the established convention.

features/monthlyComparison/MonthlyComparison.utils.ts
```diff
@@ -14,13 +14,17 @@ const fetchMonthly = async (
...
   );
   const map = new Map<string, number>();
   for (const line of stdout.split('NNN')) {
-    const [date, amount] = line.split('|').map((s) => s.trim());
-    if (date && amount) map.set(date, parseAmount(amount));
+    const [date, account, amount] = line.split('|').map((s) => s.trim());
+    if (!date || !amount) continue;
+    // `-X` injects `<Adjustment>` / `<Revalued>` postings that inherit a real
+    // payee; they must not be summed into the month's total.
+    if (account.startsWith('<')) continue;
```
naeemba: nit: two small reuse points. (1) `parseAmount` here is byte-identical to the copies in `lib/payees/parse.ts` and `lib/netWorth/parse.ts` — a shared `lib/ledger` amount parser would collapse the triplication. (2) JOURNAL-AUDIT A3 (Payees) and A4 (Reconcile) need this same `<Adjustment>`/`<Revalued>` skip; a shared `isSyntheticAccount(account)` helper would let all three sites reuse one predicate instead of re-spelling `startsWith('<')`.

naeemba (commented):
The accumulation + `<Adjustment>`/`<Revalued>` filter fix looks correct. Main point below is structural: the fixed parsing logic should follow the sibling parser pattern so it is regression-guarded.

features/monthlyComparison/MonthlyComparison.utils.ts
naeemba: Addressed in 525b370. Parsing moved to a pure, unit-tested `lib/monthly/parse.ts` (`parseMonthlyTotals`), matching the sibling `lib/payees/parse.ts` / `lib/netWorth/parse.ts` convention; `fetchMonthly` now only does IO. Tests cover per-account summation and the `<Adjustment>`/`<Revalued>` skip.

features/monthlyComparison/MonthlyComparison.utils.ts
naeemba: Addressed in 525b370. `parseAmount` collapsed into a shared `lib/ledger/parseAmount.ts` consumed by the monthly, payees and net-worth parsers (with its own test). Left the synthetic-account predicate inline — it has a single call site, so a shared helper would be premature.

lib/ledger/parseAmount.ts
```diff
@@ -0,0 +1,13 @@
...
+ * cell is either a bare number (`12,000.00`) or `<commodity> <number>`
+ * (`USD 12,000.00`); we take the numeric part, strip thousands separators, and
+ * fall back to 0 on anything unparsable so a malformed row can never poison a
+ * running total.
+ */
+export const parseAmount = (raw: string): number => {
+  if (!raw) return 0;
+  const parts = raw.trim().split(/\s+/);
+  const numericPart = parts.length > 1 ? parts[1] : parts[0];
+  return Number(numericPart.replaceAll(',', '')) || 0;
```
naeemba: This new shared parser duplicates an existing, more capable one: `utils/parseAmountColumn.ts` already extracts the numeric token from a ledger amount cell via regex and handles the shapes this whitespace-split cannot.

The whitespace split assumes commodity-as-prefix (`USD 100.00`). It silently returns 0 for suffix-rendered commodities: `parseAmount("0.5 BTC")` → `parts = ["0.5","BTC"]` → `parts[1]="BTC"` → `Number("BTC")||0` → **0**. Same for `71,214.53 Kirt` and symbol-prefixed `$100`.

That is not hypothetical for this feature. `getCashFlow` runs `reg ... -X <currency>` where `<currency>` is the user-selected base from `getBaseCurrency()`, i.e. any journal commodity — including BTC/Kirt, which ledger renders as a suffix. When a user picks such a base, every row parses to 0 and the cash-flow chart is zeroed — the same class of "garbage numbers" this PR set out to fix (A1). `parseAmountColumn("0.5 BTC")` correctly returns `0.5`.

Suggest deleting `lib/ledger/parseAmount.ts` and routing `parseMonthlyTotals`, `parsePayeeRows` and `parseNetWorthRows` through the existing `parseAmountColumn` (it already returns 0 on empty/garbage, so the `if (!raw) return 0` and `|| 0` guards are covered). This also collapses a fourth amount parser instead of adding one.

naeemba: Addressed in 34357c5. Deleted `lib/ledger/parseAmount.ts` and routed the monthly, payees and net-worth parsers through the existing `parseAmountColumn`, which extracts the numeric token via regex regardless of commodity placement. This fixes the zeroing of suffix-rendered (`0.5 BTC`, `71,214.53 Kirt`) and symbol-prefixed (`$100`) amounts when a non-prefix base currency is selected, and collapses the duplication rather than adding a fourth parser. Added a regression case to `lib/monthly/parse.test.ts` covering suffix and symbol amounts.

PR #86, opened 2026-07-10
fix(ledger): replace JS reimplementations with ledger (audit #4-8, #11-13)

naeemba:
Works through `LEDGER-AUDIT.md`, replacing JS reimplementations of ledger math with ledger itself. One commit per finding; each verified against ledger 3.4.1 on synthetic journals, with `pnpm type-check` / `lint` / `test` green (1011 tests pass).

## Fixed (8)

| # | Area | Change |
|---|------|--------|
| 4 | Portfolio total | Read the grand total from a dedicated `bal --depth 1` rollup row instead of the last line of the per-account dump (an unpriced holding no longer shows as the Total). |
| 5 | Payees | `reg ^Expenses --by-payee --collapse --sort '-display_amount'` — ledger collapses, converts and ranks; drops the `Commodities revalued` pseudo-payee; avoids the `-X` register segfault. |
| 6 | Dashboard month/year | `bal --collapse` single rollup row instead of guessing the grand-total line of a periodic register. |
| 7 | Cash-flow net | `net` computed once on the row (CSV and table can no longer diverge); regex amount parsing (no silent 0). |
| 8 | Portfolio CSV | ledger emits `quantity`/`commodity` via `%(quantity(...))`/`%(commodity(...))` + `--no-total`; the regex that mis-split `BTC 0.09` is gone. |
| 11 | Reconcile | `--sort date` in ledger; JS re-sort removed. |
| 12 | Cash-flow window/sign | `--invert` (income positive) + `-p 'last 36 months'` (calendar window) instead of JS negation + `slice(-36)`. |
| 13 | Net worth | `--display 'date>=[cutoff]'` windows the chart while `%T` keeps accumulating from journal start (`-b` would reset it). |

## Deferred (5) — see the "Remediation status" section in `LEDGER-AUDIT.md`

- **#1–#3 balancing trio (high):** correct fix is a debounced server-side ledger check as the submit/save authority while keeping JS for instant feedback — a feature with its own UX/latency design, not a mechanical swap.
- **#9 fiat-USD pivot:** the audit itself calls it "a design decision, not a drop-in" — changes the prices UI and manual-price-precedence model for "low divergence risk." Needs a product call.
- **#10 transaction magnitudes:** the list is entirely JS-parsed (`parseJournalFile`); a register replacement needs a `(file, beg_line)` join with real line-alignment risk for a display-only value.

The 4 "unavoidable selection" items are left as-is per the audit.


features/payees/Payees.tsx
```diff
@@ -34,6 +34,10 @@ const Payees = async ({ from: fromParam, to: toParam }: Props) => {
   const user = await requireUser();
   const existingViewNames = await savedViewService.listNames(user.id);
   const currentPath = `/payees/${fromParam}/${toParam}`;
+  // `--by-payee --collapse` makes ledger emit one converted row per payee;
+  // `--sort '-display_amount'` ranks them descending by converted value. This
+  // replaces JS-side summing/sorting and avoids the plain-register variant that
+  // segfaults ledger 3.4.1 under `-X`. See LEDGER-AUDIT.md #5.
   const stdout = await runLedger([
```
naeemba: This call sets `--sort -display_amount` but does not pass `{ sortByDate: false }`, so `runLedger` still injects its default `--sort -date` into `baseArgs` ahead of these args. The command ends up with two `--sort` flags (`--sort -date ... --sort -display_amount`), and the payee ranking only comes out correct because ledger happens to let the later flag win. Every other reordered call in this PR — the reconcile export, `NetWorth.tsx`, and `MonthlyComparison.utils.fetchMonthly` — suppresses the injected sort with `{ sortByDate: false }` when it sets its own order. If ledger's precedence ever differs (or on a version where the first `--sort` wins), payees would silently rank by date instead of amount and `.slice(0, TOP_N)` would pick the most *recent* payees rather than the biggest spenders. Pass `{ sortByDate: false }` here to match the sibling pattern and drop the ambiguity.

The payees export route (`app/api/payees/export/route.ts`) has the identical omission — same fix there.

naeemba: Fixed in ee8ec80. Both call sites now pass `{ sortByDate: false }` so runLedger no longer injects `--sort -date`, leaving `--sort -display_amount` as the sole sort. This matches the reconcile, net-worth, and monthly-comparison call sites and removes the reliance on ledger's later-flag-wins precedence.

lib/payees/parse.ts
```diff
@@ -1,27 +1,36 @@
...
+ * Pull the numeric magnitude out of a rendered ledger amount, regardless of
+ * where the commodity sits: `$ 1,000.00`, `USD 7`, `1,234.50`, `$ -5.00`.
+ * Grabs the first signed number token and strips thousands separators.
+ */
 const parseAmount = (raw: string): number => {
-  if (!raw) return 0;
-  const parts = raw.trim().split(/\s+/);
-  const numericPart = parts.length > 1 ? parts[1] : parts[0];
-  return Number(numericPart.replaceAll(',', '')) || 0;
+  const match = raw.match(/-?\d[\d,]*(?:\.\d+)?/);
```
naeemba: This rewrites \`parseAmount\` into a regex-strip-commas parser that is functionally the existing shared helper \`utils/parseAmountColumn.ts\` (already imported in \`app/registers/monthly/[account]/page.tsx\`), whose regex \`-?[\d,]+(?:\.\d+)?\` + \`replaceAll(',', '')\` + finite-coerce does exactly this over the same three ledger amount shapes. The PR now has three copies of this logic — this one, the identical \`parseAmount\` in \`features/monthlyComparison/MonthlyComparison.utils.ts\`, and the shared util — plus the older variant still in \`lib/netWorth/parse.ts\`. Since both call sites are being rewritten here anyway, import \`parseAmountColumn\` in both and drop the local definitions rather than adding a fourth spelling of the same parser.

naeemba: Fixed in 11f1fe4. Both call sites now import the shared `parseAmountColumn` from `@/utils/parseAmountColumn` and the local `parseAmount` copies are gone — one spelling of the parser across `lib/payees/parse.ts`, `MonthlyComparison.utils.ts`, and the monthly register page. (Left `lib/netWorth/parse.ts` alone since it is untouched by this PR.)

PR #87, opened 2026-07-10
feat(prices): price fiat through ledger via Tether pivot (audit #9)

naeemba:
Resolves LEDGER-AUDIT #9 in full: **no fiat→USD arithmetic happens in JavaScript** — ledger performs every conversion.

## Before
CoinGecko has no direct fiat→USD quote, so the provider divided in JS (`tetherUsd / perFiat`) and stored a computed `EUR→USD` row.

## After
Store the two **raw** Tether legs CoinGecko returns and let ledger's price graph bridge `fiat → USDT → USD`:

- **Provider** emits raw pivot legs — one shared `USDT→USD` anchor per run plus a `USDT→<fiat>` leg per fiat. No division.
- **Price DB regen** includes the anchor + each held fiat's leg (bypassing the held-symbols filter, since USDT is rarely held).
- **Valuation** uses the existing `-X` probe in `listKnownPricesInBase` — ledger bridges the pivot.
- **History** (`/prices/[symbol]`): `ledger prices <fiat>` can't bridge, so a pivot-priced fiat's history is reconstructed by valuing `1 <fiat>` at each pivot date through a throwaway probe journal (`reg -X base`). Ledger computes every point; the probe only supplies dates.
- **Provenance** registers the derived `(fiat, base, legDate)` key so a fetched fiat reads `source: fetched`.

## Subtleties handled (verified against ledger 3.4.1)
- **Precision**: ledger inherits an input price's decimal count when it *inverts* the `USDT→<fiat>` leg — a short quote like `0.9` would truncate `EUR→USD` to `1.1`. Pivot legs render with padded precision (`priceText`, 12 dp).
- **Same-day timing**: pivot `P` directives are dated at UTC midnight so the history probe's midnight posting picks them up (a wall-clock `fetchedAt` sorted *after* the posting and never applied — caught by the end-to-end test).
- **Manual overrides** still win: a direct `P <fiat> x USD` beats the derived pivot path.

## Tests
Full suite green (1020). New: provider emits raw legs + single anchor; regenerate/value a held fiat via pivot with fetched provenance; history reconstructs points through ledger; refreshAll end-to-end from mocked CoinGecko rates.

### DIFF
--- a/lib/prices/service.ts
+++ b/lib/prices/service.ts
@@ -83,6 +83,11 @@ const PRICE_HISTORY_CONCURRENCY = 8;
 // this many places keeps the derived `fiat → USD` rate accurate.
 const PIVOT_PRICE_DECIMALS = 12;
 
+// `date|account|quantity|commodity` per probe posting — used to reconstruct a
+// pivot-priced fiat's history by valuing `1 <fiat>` at each pivot date.
+const FIAT_HISTORY_FORMAT =
+  "%(format_date(date,'%Y-%m-%d'))|%(account)|%(quantity(scrub(display_amount)))|%(commodity(scrub(display_amount)))\n";
+
 export type RefreshResult =
   | { status: 'success'; fetched: number }
   | { status: 'partial'; fetched: number; failed: string[] }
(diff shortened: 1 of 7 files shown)


lib/prices/service.ts
```diff
@@ -452,7 +507,80 @@ export class PriceService {
...
+    if (legs.length === 0) return [];
+
+    const base = await this.resolveBaseCurrency(userId);
+    const dates = [...new Set(legs.map((r) => r.fetchedDate))].sort();
+    // Distinct account per date so ledger reports one row per point; the bare
+    // `1 "<fiat>"` is always double-quoted (legal for every commodity name).
+    const journal = dates
+      .map(
+        (date, index) =>
+          `${date} * probe\n  Probe:c${index}    1 "${symbol}"\n  Offset:c${index}   -1 "${symbol}"\n`
```
naeemba: `fiatPivotHistory` interpolates the raw `symbol` into `1 "${symbol}"`, unlike the sibling probe builder in `listKnownPricesInBase` (line ~825) which routes the symbol through `probeName()` first to strip the surrounding quote pair that `ledger commodities` adds to non-bare names (e.g. it prints `"1INCH"`).

If `symbol` arrives here already quoted, this emits `1 ""1INCH""`, ledger fails to parse the probe journal, the `catch` swallows it, and the history silently comes back empty instead of reconstructed. The route feeds `symbol` straight from `listHeldCommodities()` output, so a quoted held name reaches here verbatim.

Fiat codes are usually bare 3-letter alpha, so this is narrow in practice, but the two probe builders should treat the symbol identically — reuse `probeName()` (or share the builder) so a quoted commodity name doesn't diverge between the two paths.

lib/prices/service.ts
```diff
@@ -452,7 +507,80 @@ export class PriceService {
...
+
+  /**
+   * History for a fiat priced only through the `USDT → USD` / `USDT → <fiat>`
+   * pivot, where `ledger prices <fiat>` is empty. Values `1 <fiat>` at every
+   * pivot date via a throwaway probe journal and reads the bridged amount back
+   * — so the displayed rate is computed by ledger, never in JS. Returns `[]`
+   * when the symbol has no pivot legs (i.e. it simply has no price).
+   */
+  private async fiatPivotHistory(
+    userId: string,
```
naeemba, line 552: nit: `listKnownPrices` already fetches `listBySymbol(PIVOT_SYMBOL)` once (line ~658), but each held fiat that reaches `fiatPivotHistory` re-runs the same `listBySymbol(PIVOT_SYMBOL)` query independently (line 527). For a user holding several pivot-priced fiats that is one redundant full-table read per fiat inside the concurrency fan-out. The leg set is identical across all of them; consider threading it in or memoizing per request. Low impact given the leg count is small.

lib/prices/service.ts
naeemba: Fixed in 04d12ae. Hoisted the quote-stripping/rejecting guard into a shared module-level `probeName` and routed `fiatPivotHistory` through it, so an already-quoted held symbol no longer emits an unparseable `1 ""FOO""` probe journal. Both probe builders now treat the symbol identically.

lib/prices/service.ts
naeemba: Fixed in 04d12ae. `listPriceHistory` now takes the pivot legs `listKnownPrices` already fetched once and threads them into `fiatPivotHistory`, so the fan-out no longer re-runs `listBySymbol(PIVOT_SYMBOL)` per held pivot-priced fiat. The single-symbol callers (history page) keep the fallback fetch.

PR #90, opened 2026-07-10
fix(transactions): let ledger fill the balancing amount (#3)

naeemba:
Audit #3 — the last of the balancing trio, and the only one that wrote a JS-computed number **into the journal**.

## Before
The four structured entry forms (expense/income/transfer/exchange) with extra items appended a balancing posting whose amount `balancingPostings` computed in JS floats (`formatAmount(-residual)`). A rounding divergence from ledger's exact residual → the transaction gets rejected on save.

## After
Emit **one amount-less posting**; ledger fills the exact amount — multi-currency and cost-aware — when it parses the saved transaction. Verified a single blank absorbs a USD+EUR residual (`ledger stats` accepts it, `print` fills both). This is the pattern `fixBalanceAdapter` already uses.

- `balancingPostings` → single `{account, amount:'', currency:''}` when there's a residual; `residualByCurrency` stays as a *predicate* only (is a line needed?), never a written amount. `formatAmount` deleted.
- `transfer.detect` keyed `from` off a `< 0` amount, which an amount-less outflow no longer has → identify it by account instead.
- The four `detect`s accept `auto-balance` (a saved amount-less posting parses back as a blank line), so the Types tab still recognises its own transactions on edit.

## Trade-off
The form preview now shows the balancing line **blank** ("Blank posting will auto-balance") instead of a pre-computed figure — ledger owns the number. `canSubmit` already accepts `auto-balance`, so submit is unaffected.

Full suite green (1018). Trio complete: #1 stays as the client-side live-feedback hint, #2 (schema) and #3 now defer to ledger.

### DIFF
--- a/features/transactions/entry/types/exchange.test.ts
+++ b/features/transactions/entry/types/exchange.test.ts
@@ -60,7 +60,8 @@ describe('exchangeAdapter.compile', () => {
         cost: { amount: '100', currency: 'USD' },
       },
       { account: 'Expenses:BrokerFee', amount: '2', currency: 'USD' },
-      { account: 'Assets:Bank', amount: '-102', currency: 'USD' },
+      // Amount-less: ledger fills the outflow (100 + 2 = 102) on save.
+      { account: 'Assets:Bank', amount: '', currency: '' },
     ]);
   });
 });
--- a/features/transactions/entry/types/exchange.ts
+++ b/features/transactions/entry/types/exchange.ts
@@ -99,7 +99,9 @@ export const exchangeAdapter: TransactionTypeAdapter<ExchangeFields> = {
     const gaveFrom = singleAccount(gavePostings);
     if (!gaveFrom) return null;
     if (got.amount === '' || !(Number(got.amount) > 0)) return null;
-    if (computeBalance(postings).kind !== 'balanced') return null;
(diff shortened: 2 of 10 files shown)


naeemba:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

The four `detect` roundtrips correctly reconstruct fields: the amount-less balancing posting always lands on an asset/liability account excluded from `extraItems`/source reconstruction, and `transfer.detect`'s switch from sign-based to account-based `from` identification is the right fix for a blank posting that no longer carries a negative amount. `computeBalance` already classifies `amount:'' currency:''` as `auto-balance`, and the state is wired through preview, submit, and schema validation. No dangling references to the deleted local `formatAmount`.

This is a clean fulfillment of the [CLAUDE.md HARD RULE](https://github.com/naeemba/ledger-cli-ui/blob/e9d56be4209b69f14be6c31677ca3dcb4417ec50/CLAUDE.md#L3-L11): it removes the last JS-computed amount written into the journal and defers to ledger. `residualByCurrency` remains a JS sum but is now only a predicate (is a line needed?), which is the allowed client-side hint. Follows the existing `fixBalanceAdapter` amount-less pattern.

PR #91, opened 2026-07-12
feat(transactions): quick expense dialog in app header

naeemba:
## What

Adds a minimal, expense-only entry dialog reachable from **every page** via an "Add Expense" button in the app header (next to the currency picker).

## Why

The full transaction entry form is a generic, multi-posting editor — more than a typical user needs to jot down a single expense. This gives a fast, low-friction path for the common case.

## Details

- Collects **Amount + currency**, **Category**, **Paid from**, **Date** (defaults to today), and an optional **For** note.
- Sensible defaults so a user can save with just amount + category: paid-from pre-fills to the first asset/liability account; payee falls back to the category leaf.
- Client-side guards are hints only; the server `createTransactionAction` (→ ledger) remains the authority on validity, per the project's ledger-is-the-source-of-truth rule.

## Reuse (not rebuilt)

- `createTransactionAction` — same server save path as the full entry form.
- `expenseAdapter.compile` — existing expense → balanced-draft compiler.
- `AmountInput`, `AccountField`, `CurrencyCombobox`, `Dialog` — existing primitives.

## Files

- `features/transactions/QuickExpenseDialog.tsx` — client dialog
- `features/transactions/QuickExpenseSlot.tsx` — server loader (accounts + base currency)
- `app/layout.tsx` — mounts the slot in `headerSlot`

## Notes

- `QuickExpenseSlot` runs `ledger accounts` per page render; marked with a `ponytail:` comment to add a request-cache if it shows up in profiling.
- Verified: `tsc --noEmit` clean, eslint clean. No end-to-end run yet.

### DIFF
--- a/app/layout.tsx
+++ b/app/layout.tsx
@@ -4,6 +4,7 @@ import AppShell from '@/components/AppShell';
 import BaseCurrencyBanner from '@/components/BaseCurrencyBanner';
 import { BaseCurrencyPickerSlot } from '@/components/BaseCurrencyPicker';
 import { CryptoGate } from '@/components/crypto/CryptoGate';
+import QuickExpenseSlot from '@/features/transactions/QuickExpenseSlot';
 import { APP_NAME } from '@/lib/app';
 import { cn } from '@/lib/utils';
 import { Geist } from 'next/font/google';
@@ -53,7 +54,12 @@ export default function RootLayout({
       <body>
         <CryptoGate />
         <AppShell
-          headerSlot={<BaseCurrencyPickerSlot />}
+          headerSlot={
+            <>
+              <QuickExpenseSlot />
+              <BaseCurrencyPickerSlot />
+            </>
+          }
           bannerSlot={<BaseCurrencyBanner />}
         >
           {children}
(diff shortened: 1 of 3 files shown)


naeemba (commented):
### Code review

No blocking issues found. Ran the correctness/bug-hunt pass plus the structure/pattern checklist.

Verified:
- Layering matches the canonical sibling `NewTransaction.tsx` — server slot loads (`getAccountSuggestions` + `getAvailableCurrencies`), client dialog presents, save routes through the existing `createTransactionAction`.
- No reinvention: reuses `expenseAdapter.compile`, `AccountField`/`CurrencyCombobox`/`AmountInput`, `serializeDraftJson`, and the same account/currency loaders as the full form.
- `defaultCurrency={base}` from `getAvailableCurrencies` is the same source `NewTransaction` uses.
- CLAUDE.md ledger-is-source-of-truth rule respected: the client `compile` is a hint only; the server action (→ ledger) remains the authority. No JS accounting math introduced.
- No new unauthed-page failure surface: `runLedger` self-guards via `requireUser`, `getAccountSuggestions` swallows the throw, and `BaseCurrencyPickerSlot` already invokes `getAvailableCurrencies()` (React `cache()`d) in this same header slot, so the added call does not double-hit ledger.

nit: the dialog's own logic (validation guards, payee fallback to the category leaf, `fieldErrors` extraction) has no test, unlike the sibling `ExpenseForm.test.tsx` / `expense.test.ts`. Thin enough to skip, but a small render+save test would lock in the payee-fallback behavior.

PR #92, opened 2026-07-12
feat(transactions): quick-entry split button for all entry types

naeemba:
## What

Turns the header **Add Expense** button into a **split button**:

- **Primary click** → logs an expense (unchanged).
- **Caret dropdown** → Income, Transfer, Exchange, Fix balance.

Each type opens its own clean dialog, as minimal as the expense one — a few fields only.

## How

A single dialog engine (`QuickEntry.tsx`) drives every type from its **existing type adapter** (`expenseAdapter`, `incomeAdapter`, `transferAdapter`, `exchangeAdapter`, `fixBalanceAdapter`). The adapter's `compile()` turns the simple fields into a balanced ledger draft, then it goes through the **same `createTransactionAction`** server path as the full entry form — so ledger stays the authority on balancing/validity, per the project's ledger-is-the-source-of-truth rule.

The heavy typeForm *components* (header editor, extra items, two-column layout) are **not** reused — only the pure adapters. Per-type field sets live in `quickEntrySpecs.tsx`; the previous expense-only dialog is folded into the engine.

## Per-type fields

| Type | Fields |
|------|--------|
| Expense | amount, category, paid-from |
| Income | amount, received-into, source |
| Transfer | amount, from, to |
| Exchange | gave (amount+ccy+from), got (amount+ccy+into) |
| Fix balance | account, correct balance |

All share a date (defaults to today) + optional description, sensible account prefill, and per-type validation. Empty description falls back to a sensible payee (category leaf, or the type label).

## Files

- `features/transactions/QuickEntry.tsx` — split button + dropdown + dialog engine
- `features/transactions/quickEntrySpecs.tsx` — per-type field renderers + validation
- `features/transactions/QuickEntrySlot.tsx` — server loader (accounts + base currency)
- `app/layout.tsx` — mounts the slot in `headerSlot`
- removes the old `QuickExpenseDialog` / `QuickExpenseSlot`

## Verification

- `tsc --noEmit` clean, eslint clean.
- `vitest run features/transactions/entry` → 128 passed.
- No live click-through yet.


features/transactions/QuickEntry.tsx
```diff
@@ -0,0 +1,190 @@
...
+            />
+          </Field>
+          <div className="flex flex-col gap-1.5">
+            <Label htmlFor="quick-entry-payee">
+              Description{' '}
+              <span className="text-muted-foreground">(optional)</span>
+            </Label>
+            <Input
+              id="quick-entry-payee"
+              value={fields.payee}
```
naeemba, line 107: The "Description (optional)" field is bound to `fields.payee`, but `makeEmpty` seeds `payee` from each adapter's `emptyFields`. For transfer/exchange/fixBalance those defaults are non-empty (`Transfer`, `Currency exchange`, `Balance adjustment`), so the optional field opens pre-filled with a value the user has to delete. Two consequences:

- Inconsistent UX: expense/income open with an empty description, the other three do not.
- The `fields.payee.trim() || spec.resolvePayee?.(fields) || spec.label` fallback in `save` becomes dead code for those three types — `payee` is never empty, so `resolvePayee`/`label` never fires.

Consider blanking `payee` in `seed()` (or `makeEmpty`) so all types start empty and the fallback chain drives the payee, or drop the unused `resolvePayee`/label fallback for the types that keep a default.

features/transactions/quickEntrySpecs.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+const exchangeSpec: QuickEntrySpec<ExchangeFields> = {
+  kind: 'exchange',
+  label: 'Exchange',
+  icon: '💱',
+  adapter: exchangeAdapter,
+  makeEmpty: (ctx) => ({
+    ...seed(exchangeAdapter, ctx),
+    gaveFrom: firstMoneyAccount(ctx.accounts),
+    gotInto: firstMoneyAccount(ctx.accounts),
```
naeemba: Exchange seeds `gaveFrom` and `gotInto` to the same `firstMoneyAccount(...)`, and unlike the transfer spec (which guards `f.from === f.to`) the exchange `validate` has no `gaveFrom !== gotInto` check. If the user only fills amounts/currencies and leaves both account pickers at the default, they save an exchange from an account into itself — a nonsensical (though ledger-balanceable) posting. Either seed only one side, or add a same-account guard mirroring transfer.

features/transactions/quickEntrySpecs.tsx
```diff
@@ -0,0 +1,329 @@
...
+ * correctly balanced ledger draft — is delegated to the shared `adapter`, so
+ * this file never does accounting math.
+ */
+export type QuickEntrySpec<F extends HeaderFields> = {
+  kind: string;
+  label: string;
+  icon: string;
+  adapter: TransactionTypeAdapter<F>;
+  makeEmpty: (ctx: QuickEntryContext) => F;
+  validate: (fields: F) => string | null;
```
naeemba, line 44: No tests cover the new quick-entry engine. The per-type `validate` functions and the payee-resolution/fallback logic are pure and cheap to test, and the old `QuickExpenseDialog` this replaces had validation inlined too. A small table-driven test over `QUICK_ENTRY_SPECS` (each spec: empty fields fail validation with the expected message; a filled set passes; `resolvePayee` returns the expected leaf) would lock in the behavior the adapters alone do not cover. `vitest run features/transactions/entry` (128 passed) exercises the adapters, not this file.

features/transactions/QuickEntry.tsx
naeemba: Fixed in 9951219. `seed()` now blanks `payee` for every type, so the optional description opens empty across all five and the `resolvePayee`/`label` fallback in `save` drives the payee. Transfer/exchange/fix-balance keep their prior saved-payee text via a `resolvePayee` returning their previous default (`Transfer` / `Currency exchange` / `Balance adjustment`).

features/transactions/quickEntrySpecs.tsx
naeemba: Fixed in 9951219. Added a `gaveFrom !== gotInto` guard to the exchange `validate` mirroring transfer ("Paid-from and received-into accounts must differ."), and `makeEmpty` no longer seeds `gotInto`, so the two pickers no longer start on the same account.

features/transactions/quickEntrySpecs.tsx
naeemba: Added in 14aead7: `features/transactions/quickEntrySpecs.test.tsx`, table-driven over `QUICK_ENTRY_SPECS` — blank-description seeding, empty fields failing with the expected message, filled entries passing, the transfer and exchange same-account guards, and `resolvePayee` returning the expected leaf/label.

PR #93, opened 2026-07-13
feat(transactions): Debt quick-entry with two directions

naeemba:
## What

Adds a **Debt** item to the quick-entry dropdown — a two-direction form for money owed, aimed at a user who doesn't think in double-entry.

- Tabs: **They owe you** / **You owe them**
- Fields: Name, Amount, and the cash account (labelled *Paid from* / *Received into* per direction)

## Account model

The form names a person and builds the right account, so a per-person balance is a single ledger query over the person's leaf (`ledger bal /:<Name>$/`):

| Direction | Account | Effect |
|-----------|---------|--------|
| They owe you | `Assets:Receivable:<Name>` | receivable rises, your cash falls |
| You owe them | `Liabilities:Payable:<Name>` | your cash rises, your payable falls |

Two trees (mirrored leaf) keep net-worth and balance reports correct with no negative-asset caveats.

## How

A debt is mechanically a **transfer**, so it compiles through `transferAdapter` — no new accounting code. `QuickEntrySpec` now carries a `compile` function directly (instead of requiring a full registry adapter), letting a type supply its own compilation without joining the full entry-form registry.

- **Name → single account segment**: colons stripped (no fake sub-hierarchy), doubled spaces collapsed, trimmed.
- **Reuse**: the Name field is a combobox of existing people parsed from `Receivable:*` / `Payable:*` leaves, so the same person maps to the same account.
- Blank description falls back to a direction-aware payee (`Lent to <Name>` / `Borrowed from <Name>`).

## Scope

Entry form only — the accounts show up in existing balance reports. A dedicated per-person "Debts" balances view is a follow-up.

## Tests

`quickEntrySpecs.test.tsx` covers both directions' postings, name sanitization, and validation. `tsc` clean, eslint clean, 27 passing. Not click-tested live yet.

### DIFF
--- a/features/transactions/quickEntrySpecs.tsx
+++ b/features/transactions/quickEntrySpecs.tsx
@@ -2,6 +2,7 @@
 
 import React from 'react';
 import AmountInput from './AmountInput';
+import type { DraftState } from './entry/draftReducer';
 import {
   AccountField,
   CurrencyCombobox,
@@ -21,6 +22,8 @@ import {
 } from './entry/types/fixBalance';
 import { incomeAdapter, type IncomeFields } from './entry/types/income';
 import { transferAdapter, type TransferFields } from './entry/types/transfer';
+import Combobox from '@/components/Combobox';
+import { Button } from '@/components/ui/button';
 
 export type QuickEntryContext = { accounts: string[]; defaultCurrency: string };
 
@@ -30,16 +33,16 @@ type FieldsProps<F> = {
 } & QuickEntryContext;
 
 /**
- * A single quick-entry form: how to seed it, validate it, name it, and render
- * its handful of inputs. The heavy lifting — turning these simple fields into a
- * correctly balanced ledger draft — is delegated to the shared `adapter`, so
- * this file never does accounting math.
+ * A single quick-entry form: how to seed it, validate it, name it, render its
+ * handful of inputs, and compile it to a ledger draft. `compile` delegates to a
+ * shared type adapter, so this file never does accounting math — it only builds
+ * the simple fields and hands them off.
  */
 export type QuickEntrySpec<F extends HeaderFields> = {
   kind: string;
   label: string;
   icon: string;
-  adapter: TransactionTypeAdapter<F>;
+  compile: (fields: F, ctx: TypeContext) => DraftState;
   makeEmpty: (ctx: QuickEntryContext) => F;
   validate: (fields: F) => string | null;
   // Fallback payee when the user leaves it blank (the journal requires one).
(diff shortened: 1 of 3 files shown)


features/transactions/quickEntrySpecs.tsx
```diff
@@ -325,6 +328,143 @@ const fixBalanceSpec: QuickEntrySpec<FixBalanceFields> = {
...
+    cashAccount: firstMoneyAccount(ctx.accounts),
+  }),
+  validate: (f) =>
+    !cleanPerson(f.person)
+      ? 'Enter a name.'
+      : !isPositive(f.amount)
+        ? 'Enter an amount.'
+        : !f.cashAccount.trim()
+          ? 'Pick an account.'
+          : null,
```
naeemba: The transfer and exchange specs both guard against the source and destination resolving to the same account (`from === to` → "Source and destination must differ."), but debt has no equivalent check. Because the cash-account picker offers every asset/liability account — including `Assets:Receivable:*` / `Liabilities:Payable:*` — a user can pick `Assets:Receivable:Alex` as the cash account and name the person "Alex". For direction `owed-to-you` that compiles to `to = from = Assets:Receivable:Alex`, i.e. `[{+50}, {-50}]` on one account. It nets to zero so ledger accepts it, silently saving a no-op transaction. Add a `from === to` check in `validate` for parity with the sibling specs.

naeemba: Fixed in e75c1c8. Added a `from === to` guard to the debt `validate` (message: "The cash account must differ from the person."). Factored the account resolution into a shared `debtAccounts` helper so validate and compile agree on which two accounts a debt touches, and added a test covering the case where the cash account resolves to the person's receivable/payable account.

features/transactions/quickEntrySpecs.tsx
```diff
@@ -325,6 +328,153 @@ const fixBalanceSpec: QuickEntrySpec<FixBalanceFields> = {
...
+const cleanPerson = (raw: string) =>
+  raw.replace(/:/g, ' ').replace(/\s+/g, ' ').trim();
+
+// Existing people, parsed from the receivable/payable account leaves, so the
+// same person reuses the same account instead of spawning a near-duplicate.
+const knownPeople = (accounts: string[]): string[] => {
+  const people = new Set<string>();
+  for (const account of accounts) {
+    const match = account.match(
+      /^(?:Assets:Receivable|Liabilities:Payable):(.+)$/
```
naeemba: nit: `knownPeople` captures `(.+)$` from the account leaf, so a nested receivable like `Assets:Receivable:Family:Alex` is offered to the combobox as `Family:Alex`. Selecting it runs through `cleanPerson`, which flattens the colon to `Family Alex` and builds a *different* account (`Assets:Receivable:Family Alex`) than the one it was parsed from — silently breaking the reuse this combobox exists for. Nested people are unusual and colon-flattening is deliberate, so this is a corner, not a blocker; a `[^:]+$` capture (or flattening the leaf on parse too) would keep parse and build symmetric.

PR #94, opened 2026-07-13
feat(transactions): split expenses and income across categories

naeemba:
Surfaces the entry adapters' existing `extraItems` in the quick-entry form.

## What
A collapsed **"add another line"** affordance under **Expense** and **Income**:
- Expense: split one receipt across several categories (groceries + household).
- Income: carry deductions on a paycheck (salary − tax → net into an account).

The single-category default is unchanged — the extra rows only appear once you add one.

## Accounting stays in ledger
Extra rows compile to postings; the paid-from / received-into line is left
**amount-less** so ledger fills the residual. No JS summation (CLAUDE.md hard rule).
Half-filled rows (account or amount alone) are dropped rather than emitted.

## Tests
`quickEntrySpecs.test.tsx` covers the expense split, income deductions, and the
half-filled-row drop. `tsc --noEmit` and `vitest run features/transactions` green.

### DIFF
--- a/features/transactions/quickEntrySpecs.tsx
+++ b/features/transactions/quickEntrySpecs.tsx
@@ -1,5 +1,6 @@
 'use client';
 
+import { XIcon } from 'lucide-react';
 import React from 'react';
 import AmountInput from './AmountInput';
 import type { DraftState } from './entry/draftReducer';
@@ -16,6 +17,7 @@ import type {
 } from './entry/types/adapter';
 import { exchangeAdapter, type ExchangeFields } from './entry/types/exchange';
 import { expenseAdapter, type ExpenseFields } from './entry/types/expense';
+import type { ExtraItem } from './entry/types/extraItems';
 import {
   fixBalanceAdapter,
   type FixBalanceFields,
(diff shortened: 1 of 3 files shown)


features/transactions/quickEntrySpecs.tsx
```diff
@@ -95,6 +97,74 @@ const AmountRow = ({
...
+  const setItem = (index: number, patch: Partial<ExtraItem>) =>
+    onChange(items.map((it, i) => (i === index ? { ...it, ...patch } : it)));
+  const add = () => onChange([...items, { account: '', amount: '', currency }]);
+  const remove = (index: number) =>
+    onChange(items.filter((_, i) => i !== index));
+
+  return (
+    <>
+      {items.map((item, index) => (
+        <div key={index} className="flex items-end gap-2">
```
naeemba: Rows are keyed by array index, but `AccountField` wraps `Combobox`, which keeps uncontrolled internal state (`open`/`search`) not derived from `value`. Remove a middle row and every row below shifts up one index — React reuses the same DOM/state for a now-different row, so the open-dropdown/search state (and focus) misassociates to the wrong line.

This is exactly the case the existing `ExtraItemsField` (`entry/typeForms/ExtraItemsField.tsx`) guards against with a stable per-row `crypto.randomUUID()` key — see its comment: "keying rows by array index would misassociate that state when a middle row is removed and the rows below shift up." `SplitRows` reintroduces the bug. Key by a stable id.

features/transactions/quickEntrySpecs.tsx
```diff
@@ -95,6 +97,74 @@ const AmountRow = ({
...
+  label: string;
+  addLabel: string;
+  items: ExtraItem[];
+  currency: string;
+  accounts: string[];
+  onChange: (items: ExtraItem[]) => void;
+}) => {
+  const setItem = (index: number, patch: Partial<ExtraItem>) =>
+    onChange(items.map((it, i) => (i === index ? { ...it, ...patch } : it)));
+  const add = () => onChange([...items, { account: '', amount: '', currency }]);
```
naeemba: Each split row snapshots the header `currency` at add-time and `SplitRows` renders no per-row currency control (only `AccountField` + `AmountInput` + remove). So if the user adds a split, then changes the header currency via the `AmountRow` picker, the split row keeps the stale currency and there is no way to correct it. The row then compiles to a posting in the wrong commodity.

The existing `ExtraItemsField` renders a `CurrencyCombobox` per row precisely so each extra line can carry its own commodity. Either add the per-row currency picker, or re-sync each row's currency when the header currency changes.

features/transactions/quickEntrySpecs.tsx
```diff
@@ -95,6 +97,74 @@ const AmountRow = ({
   </Field>
 );
 
+// Optional extra category lines on an expense/income. Collapsed to a single
+// "add line" affordance by default so the common one-category entry stays
+// clean; each added row is a category + amount. The adapters already turn these
+// into balancing postings, so ledger — not this UI — does the summing.
+const SplitRows = ({
```
naeemba: `SplitRows` re-implements the already-existing `ExtraItemsField` (`entry/typeForms/ExtraItemsField.tsx`), which the Types-tab Expense/Income/Transfer/Exchange forms all use for the same `ExtraItem[]` rows. Reinventing it here drops three things that component already handles correctly: (1) stable UUID row keys (see the `key={index}` comment), (2) the `MAX_POSTINGS = 50` cap — `SplitRows` lets a user add unbounded rows, and the compiled draft can exceed ledger's posting handling with no feedback, and (3) the per-row currency picker.

Prefer reusing `ExtraItemsField` (it already takes `items`/`accounts`/`defaultCurrency`/`baseCount`/`onChange`). If the quick-entry visual treatment must differ (collapsed "+ add line" affordance, tighter layout), lift the shared row logic — id-keying, cap, currency — into that one component and parameterize the presentation, rather than maintaining a second divergent copy.

naeemba: Fixed in 51f4a34. Deleted `SplitRows` and reused `ExtraItemsField` in the expense/income specs, parameterizing only its section and add-button labels. That restores the three things the copy dropped — stable UUID row keys, the `MAX_POSTINGS = 50` cap, and the per-row currency picker — and keeps that row logic in one component instead of two divergent copies.

features/transactions/quickEntrySpecs.tsx
naeemba: Fixed in 51f4a34. `SplitRows` is gone; the expense/income splits now render through `ExtraItemsField`, which keys rows by a stable per-row `crypto.randomUUID()` kept in lockstep with the items, so removing a middle row no longer misassociates the open-dropdown/search/focus state to the wrong line.

features/transactions/quickEntrySpecs.tsx
naeemba: Fixed in 51f4a34. By reusing `ExtraItemsField`, each split row now has its own `CurrencyCombobox`, so a row's commodity is editable independently of the header and no longer silently keeps a stale header currency.

PR #96, opened 2026-07-13
feat(transactions): confirm quick-entry saves with an undo toast

naeemba:
Closes the loop on a quick save with feedback + an escape hatch.

## What
On a successful quick-entry save, a success toast names what was saved (`Expense saved` + the description) and offers **Undo**, which removes the just-created transaction.

## Plumbing
`deleteTransactionAction` already existed but needs the transaction `uid` and the current journal fingerprint — neither of which `createTransactionAction` returned. So:
- `addTransaction` now returns the generated `uid`; `createTransactionAction` surfaces it.
- New `undoTransactionAction(uid)` resolves the current journal fingerprint **server-side** and delegates to the existing audited delete path. The toast never carries a fingerprint.

If the journal changed between save and undo, the existing stale-guard makes the undo fail cleanly (an error toast) rather than delete the wrong data. `sonner` was already mounted in the app shell.

## Tests
`tsc --noEmit` + `vitest run features/transactions lib/journal` green (284).

### DIFF
--- a/features/transactions/actions/undoTransaction.ts
+++ b/features/transactions/actions/undoTransaction.ts
@@ -3,19 +3,24 @@
 import { deleteTransactionAction } from './deleteTransaction';
 import type { DeleteTransactionResult } from './deleteTransaction';
 import { requireUser } from '@/lib/auth/require-user';
-import { journalRepository } from '@/lib/journal';
+import { journalService } from '@/lib/journal';
 
 /**
  * Undo a just-created transaction. Unlike a row-level delete, the caller (a
- * toast fired right after save) doesn't hold the journal fingerprint, so we
- * resolve the current one server-side and hand off to the audited delete path.
- * If anything else mutated the journal in between, that guard makes the undo
- * fail cleanly rather than delete stale data.
+ * toast fired right after save) doesn't hold the transaction fingerprint, so we
+ * resolve the created row server-side and hand off to the audited delete path.
+ * The delete guard compares against the per-transaction fingerprint, so we must
+ * pass that (not the whole-journal fingerprint). If anything else mutated the
+ * transaction in between, that guard makes the undo fail cleanly rather than
+ * delete stale data.
  */
 export async function undoTransactionAction(
   uid: string
 ): Promise<DeleteTransactionResult> {
   const user = await requireUser();
-  const fingerprint = await journalRepository.getFingerprint(user.id);
-  return deleteTransactionAction(uid, fingerprint);
+  const tx = await journalService.findTransaction(user.id, uid);
+  if (!tx?.fingerprint) {
+    return { ok: false, message: 'This transaction no longer exists.' };
+  }
+  return deleteTransactionAction(uid, tx.fingerprint);
 }
--- a/features/transactions/actions/undoTransaction.ts
+++ b/features/transactions/actions/undoTransaction.ts
@@ -1,0 +1,21 @@
+'use server';
+
+import { deleteTransactionAction } from './deleteTransaction';
+import type { DeleteTransactionResult } from './deleteTransaction';
+import { requireUser } from '@/lib/auth/require-user';
+import { journalRepository } from '@/lib/journal';
+
+/**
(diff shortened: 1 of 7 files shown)


features/transactions/actions/undoTransaction.ts
```diff
@@ -0,0 +1,21 @@
...
+ * toast fired right after save) doesn't hold the journal fingerprint, so we
+ * resolve the current one server-side and hand off to the audited delete path.
+ * If anything else mutated the journal in between, that guard makes the undo
+ * fail cleanly rather than delete stale data.
+ */
+export async function undoTransactionAction(
+  uid: string
+): Promise<DeleteTransactionResult> {
+  const user = await requireUser();
+  const fingerprint = await journalRepository.getFingerprint(user.id);
```
naeemba: `getFingerprint` returns the wrong fingerprint for the delete stale-guard, so Undo always fails.

`journalRepository.getFingerprint(user.id)` returns the **whole-journal content fingerprint** (`cachedPull` → hash of the entire pulled journal, repository.ts:26/137). But `deleteTransaction`'s stale-guard compares it against `current.fingerprint`, which is the **per-transaction** SHA-256 of the formatted transaction block (`fingerprintDraft`, fingerprint.ts:5-6, stamped per row in parser.ts:173). These are two different fingerprint spaces and will never be equal.

Result: `performDelete` (service.ts:491) hits `current.fingerprint !== input.expectedFingerprint` on every undo and returns `reason: 'stale'`, so `undoTransactionAction` always resolves `{ ok: false, message: 'This transaction was modified somewhere else.' }`. The Undo button is dead on arrival — it never removes the transaction.

Compare the canonical caller `RowActions.tsx:28`: `deleteTransactionAction(t.uid!, t.fingerprint!)` passes the **per-transaction** fingerprint, not the journal one. Undo needs the same value — resolve the created row (e.g. `journalService.findTransaction(user.id, uid)`) and pass its `.fingerprint`, not `getFingerprint`.

No test exercises the undo path end-to-end; the PR's green suite does not cover this. Adding one would have caught it.

naeemba: Fixed in f535da3. Undo now resolves the created row via `journalService.findTransaction` and passes its per-transaction `.fingerprint` to `deleteTransactionAction`, matching the fingerprint space the stale-guard compares against (previously it passed the whole-journal fingerprint from `getFingerprint`, which never matched). Also handles the case where the row no longer exists with a clean failure message.

features/transactions/actions/undoTransaction.ts
```diff
@@ -0,0 +1,26 @@
...
+ * The delete guard compares against the per-transaction fingerprint, so we must
+ * pass that (not the whole-journal fingerprint). If anything else mutated the
+ * transaction in between, that guard makes the undo fail cleanly rather than
+ * delete stale data.
+ */
+export async function undoTransactionAction(
+  uid: string
+): Promise<DeleteTransactionResult> {
+  const user = await requireUser();
+  const tx = await journalService.findTransaction(user.id, uid);
```
naeemba, line 21: No test covers the undo path. The prior round caught a real bug here — `getFingerprint` (whole-journal) vs the per-transaction fingerprint the delete stale-guard compares — and the fix commit resolves it correctly, but adds no regression test. The exact defect that shipped is the kind an end-to-end test would have caught, and undo remains the only mutation action with no coverage (`audit.test.ts` exercises add/edit/delete). Worth a service-level test asserting that `findTransaction(uid).fingerprint` fed to `deleteTransaction` actually removes the row (i.e. the two fingerprint spaces match), so a future `getFingerprint`-style regression fails loudly.

naeemba: Added in 4bc18b2. New service-level test in service.test.ts: adds a transaction, resolves it via findTransaction(uid), feeds that per-transaction .fingerprint straight into deleteTransaction, and asserts the row is gone (findTransaction returns null after). This exercises the exact undo flow, so a future getFingerprint-style regression (whole-journal fingerprint) makes the delete return stale and the test fails loudly.

PR #98, opened 2026-07-13
feat(transactions): repeat a saved template from quick-entry

naeemba:
Gives quick-entry a **Repeat a template** shortcut so a daily recurring entry
isn't retyped.

## What
A menu item (shown only when the user has ≥1 saved template) opens a picker;
clicking a template posts it **immediately** with today's date.

The existing `TemplatePicker` opens the *full* form pre-filled — this adds the
instant one-click repeat that quick-entry is for.

## Reuse
- `Transaction.fromTemplate(draft, currency).withField('date', today)` → the same
  `createTransactionAction` the type forms use, so ledger validates the result.
- `QuickEntrySlot` loads the user's templates alongside accounts/currency (same
  pattern as `NewTransaction`).

## Tests
Added a model test for the hydrate → stamp-date → create-wire composition.
`tsc --noEmit` + `vitest run lib/transactions/model features/transactions` green (214).

### DIFF
--- a/features/transactions/QuickEntry.tsx
+++ b/features/transactions/QuickEntry.tsx
@@ -162,7 +162,10 @@ function RepeatTemplate({
         onDone();
         router.refresh();
       } else {
-        setError(result.formError ?? 'Could not save.');
+        const fieldError = result.fieldErrors
+          ? Object.values(result.fieldErrors).flat()[0]
+          : undefined;
+        setError(result.formError ?? fieldError ?? 'Could not save.');
       }
     });
 
--- a/features/transactions/QuickEntry.tsx
+++ b/features/transactions/QuickEntry.tsx
@@ -6,7 +6,11 @@ import { createTransactionAction } from './actions';
 import { serializeDraftJson } from './entry/draftReducer';
 import { Field } from './entry/typeForms/fields';
 import type { HeaderFields } from './entry/types/adapter';
-import { QUICK_ENTRY_SPECS, type QuickEntrySpec } from './quickEntrySpecs';
+import {
+  QUICK_ENTRY_SPECS,
+  todayLocal,
+  type QuickEntrySpec,
+} from './quickEntrySpecs';
 import { Button } from '@/components/ui/button';
 import {
   Dialog,
@@ -33,8 +37,6 @@ type Props = {
   defaultCurrency: string;
   templates?: Template[];
 };
-
-const todayLocal = () => new Date().toLocaleDateString('en-CA');
 
 /**
  * The dialog body for one entry type. Owns the field state and the save path
--- a/features/transactions/QuickEntry.tsx
+++ b/features/transactions/QuickEntry.tsx
@@ -1,6 +1,6 @@
 'use client';
 
-import { ChevronDownIcon, PlusIcon } from 'lucide-react';
+import { ChevronDownIcon, PlusIcon, RepeatIcon } from 'lucide-react';
 import { useState, useTransition } from 'react';
 import { createTransactionAction } from './actions';
 import { serializeDraftJson } from './entry/draftReducer';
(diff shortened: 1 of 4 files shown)


features/transactions/QuickEntry.tsx
```diff
@@ -120,13 +129,79 @@ function QuickEntryContent({
...
+        defaultCurrency
+      ).withField('date', todayLocal());
+      const formData = new FormData();
+      formData.set('draft', serializeDraftJson(draft, 'create'));
+      const result = await createTransactionAction(null, formData);
+      if (result.ok) {
+        onDone();
+        router.refresh();
+      } else {
+        setError(result.formError ?? 'Could not save.');
```
naeemba: `RepeatTemplate` surfaces only `result.formError`, but `createTransactionAction` returns validation failures as `fieldErrors` with no `formError` (the Zod-invalid path in `journalService.addTransaction` returns `{ ok: false, reason: 'invalid', fieldErrors }` with `formError` undefined — see `lib/journal/service.ts:151`). So if a repeated template produces a draft the schema rejects (e.g. a posting that no longer passes `accountSchema`, a template saved before a schema tightening), the user sees the generic "Could not save." and the real reason is swallowed.

`QuickEntryContent` in this same file already handles both:

```ts
const fieldError = result.fieldErrors
  ? Object.values(result.fieldErrors).flat()[0]
  : undefined;
setError(result.formError ?? fieldError ?? 'Could not save.');
```

Mirror that here so the two save paths report failures consistently.

features/transactions/QuickEntry.tsx
```diff
@@ -19,13 +19,22 @@ import {
...
 import { useRouter } from 'next/navigation';
 
-type Props = { accounts: string[]; defaultCurrency: string };
+type Props = {
+  accounts: string[];
+  defaultCurrency: string;
+  templates?: Template[];
+};
+
+const todayLocal = () => new Date().toLocaleDateString('en-CA');
```
naeemba: nit: this exact helper already exists as a (private) `todayLocal` in `features/transactions/quickEntrySpecs.tsx:53` — same one-liner, `() => new Date().toLocaleDateString('en-CA')`. This file already imports from `quickEntrySpecs`; consider exporting the existing one and reusing it instead of re-defining it, so the two do not drift.

features/transactions/QuickEntry.tsx
naeemba: Fixed in 3187967 — `RepeatTemplate` now flattens `fieldErrors` the same way `QuickEntryContent` does, so a schema-rejected template surfaces the real reason instead of the generic "Could not save."

features/transactions/QuickEntry.tsx
naeemba: Fixed in e8066bd — exported the existing `todayLocal` from `quickEntrySpecs` and imported it here, dropping the duplicate definition.

features/transactions/QuickEntrySlot.tsx
```diff
@@ -1,15 +1,25 @@
...
 
-// Server wrapper: loads the account list + base currency and hands them to the
-// client quick-entry split button. Placed in the app header slot so it's
-// reachable from every page. ponytail: re-runs `ledger accounts` per page
-// render; add a cache wrapper if it shows up in profiling.
+// Server wrapper: loads the account list + base currency + saved templates and
+// hands them to the client quick-entry split button. Placed in the app header
+// slot so it's reachable from every page. ponytail: re-runs `ledger accounts`
+// per page render; add a cache wrapper if it shows up in profiling.
 export default async function QuickEntrySlot() {
```
naeemba, line 11: `requireUser()` here redirects logged-out visitors on every route. `QuickEntrySlot` is passed as `headerSlot` from the root `app/layout.tsx`, so it is eagerly server-rendered on *every* path — including the ones with no session. `requireUser()` → `requireSession()` calls `redirect('/sign-in')` when there is no session, and that redirect fires during the server render regardless of whether the client `AppShell` later chooses to display the slot.

Concrete breakage:
- Logged-out visitor to the marketing landing `/` (a `PUBLIC_PATH`) gets bounced to `/sign-in` — the landing becomes unreachable while signed out.
- `/account/deleted` with a stale session cookie (the exact case commit 9352138 tolerates) redirects to `/sign-in`, nullifying that fix — `getSession()` returns null server-side even though the presence-only proxy cookie check let the request through.
- `/sign-in` and `/sign-up` render the root layout, so this slot's `redirect('/sign-in')` fires on the auth pages themselves.

The sibling slot rendered right next to this one, `BaseCurrencyPickerSlot`, already handles exactly this: it uses `getOptionalUser()` and `if (!user) return null`. Mirror that here:

```ts
const user = await getOptionalUser();
if (!user) return null;
const [accounts, { base }, templates] = await Promise.all([...]);
```

naeemba: Fixed in ab71313 — QuickEntrySlot now uses `getOptionalUser()` and returns null when there is no session, matching the sibling `BaseCurrencyPickerSlot`. Logged-out visitors on public paths (landing, sign-in, sign-up, account/deleted) are no longer bounced to /sign-in by this slot.

PR #99, opened 2026-07-13
feat(debts): per-person net view over the receivable/payable model

naeemba:
Replaces the legacy `Assets:Credited` debts page with a per-person view over the
debt account model the quick-entry **Debt** / **Settle up** features use.

> Per your choice: **replace** the old page (legacy `Assets:Credited` balances no
> longer appear here). If you still have `Assets:Credited` data, a one-time
> migration to `Assets:Receivable:<Name>` / `Liabilities:Payable:<Name>` would be
> needed — happy to script that separately.

## What
Net balance per person across `Assets:Receivable:*` (owed to you) and
`Liabilities:Payable:*` (you owe), one row per person, settled people hidden.

## Ledger does all the math (CLAUDE.md hard rule)
- One `balance` call finds people with an open account.
- One **collapsed `register`** per person nets their two sides into a single
  base-converted figure — its final running-total line. Verified against ledger
  3.4.1 on a synthetic multi-account journal.
- JS only groups people **by name** (never by amount → no ranking) and reads the
  net's **sign** to pick the direction label. No summation, no rate math.

CSV export moves to the same per-person shape.

## Tests
`features/debts/parse.test.ts` (people extraction, net-line parsing, sign→direction,
zero/empty → hidden) and updated `lib/debts/csv.test.ts`. Full suite: 1052 passed.


naeemba (commented):
### Code review

Found 2 issues:

1. `NET_FORMAT` uses `total` instead of `display_total`, so `-X` conversion is not applied and the value expression stays multi-commodity. For any person holding two different original commodities, `%(quantity(scrub(total)))` throws `Cannot convert a balance with multiple commodities to an amount` — `runLedger` rejects and the whole page falls back to "Failed to load debts." Verified on ledger 3.4.1: `total` errors while `display_total` yields the correctly base-converted `$ 94.00`. The repo's own audit already prescribes this form (LEDGER-AUDIT.md:82: `%(quantity(scrub(display_total)))`). Fix: swap `total` → `display_total` in all three format fields.

https://github.com/naeemba/ledger-cli-ui/blob/552415202efe4f67f2b54e44405d27dbe7c4b478/features/debts/getPersonDebts.ts#L14-L17

2. `PERSON_ACCOUNT` uses a greedy `(.+)$` capture, so a nested account like `Assets:Receivable:Bob:Car` and `Assets:Receivable:Bob:Rent` are extracted as two distinct "people" (`Bob:Car`, `Bob:Rent`) rather than one "Bob". The sibling writer that creates these accounts anchors a single segment — `([^:]+)$` in `knownPeople` (features/transactions/quickEntrySpecs.tsx:360-364). Since ledger's register account args are prefix matches, extracting just `Bob` nets both sub-accounts correctly (verified: `Assets:Receivable:Bob` → $150). Fix: match `([^:]+)$` to align with the established pattern and collapse sub-accounts into one person.

https://github.com/naeemba/ledger-cli-ui/blob/552415202efe4f67f2b54e44405d27dbe7c4b478/features/debts/parse.ts#L15-L24

naeemba:
Both findings fixed.

1. \`total\` → \`display_total\` in all three \`NET_FORMAT\` fields (655a384). Verified on ledger 3.4.1: with a person holding \`$50\` and \`EUR 40\`, \`quantity(scrub(total))\` errors with "Cannot convert a balance with multiple commodities to an amount", while \`display_total\` applies the \`-X\` conversion and yields \`94|$|$ 94.00\`.

2. \`PERSON_ACCOUNT\` now captures only the first segment after the root (bbf6e81). Note the reviewed suggestion of anchoring \`([^:]+)$\` would have dropped the person entirely: the \`balance --flat\` listing this parser consumes emits leaf accounts (\`Assets:Receivable:Bob:Car\`, \`:Bob:Rent\`), which an end-anchored single-segment pattern matches on neither. The sibling \`knownPeople\` gets away with the anchor because it reads single-segment accounts. Fix here is an unanchored \`([^:]+)\` first-segment capture, so \`Bob:Car\`/\`Bob:Rent\` collapse to one \`Bob\`; added a regression test.

features/debts/getPersonDebts.ts
```diff
@@ -0,0 +1,66 @@
...
+  const stdout = await runLedger([
+    'register',
+    '-X',
+    base,
+    '--collapse',
+    '--format',
+    NET_FORMAT,
+    '--',
+    `${RECEIVABLE_ROOT}:${person}`,
+    `${PAYABLE_ROOT}:${person}`,
```
naeemba: The person name is interpolated into a ledger **register account argument, which ledger treats as an unanchored regex** — so one person's balance leaks into another's net whenever their name is a prefix (or regex-superstring) of the query.

Verified against ledger 3.4.1 on a synthetic journal:
```
Assets:Receivable:Bob    $100
Assets:Receivable:Bob    $-30   (repay)
Liabilities:Payable:Bob  $-20
Assets:Receivable:Bobby  $40
```
`register -- 'Assets:Receivable:Bob' 'Liabilities:Payable:Bob'` matches `Assets:Receivable:Bob`, `Assets:Receivable:Bob:Car` **and `Assets:Receivable:Bobby`**, netting Bob to **90** instead of the correct **50** — Bobby's $40 is counted under Bob. Any prefix pair (Jo/John, Sam/Samantha) corrupts the netted figure. `peopleFromBalance` lists Bob and Bobby as separate people, so Bob's row is silently wrong while Bobby's is right.

Regex metacharacters in a name compound it: a person `A.C` matches `AXC` (the `.` is a wildcard), and `isSafeLedgerArg` only rejects a leading `-`, not `.`/`(`/`|`/etc.

Fix: anchor + escape the name so the arg matches exactly this person's account and its sub-accounts, not a prefix sibling. `peopleFromBalance` already isolated the exact first segment, so the register query should require a `:` or end-of-account boundary after the name and escape regex metachars in it.

naeemba: Fixed in 21acc9f. The name is now escaped and anchored before it goes into the register query.

Verified against ledger 3.4.1: two subtleties beyond the plain escape+anchor.

1. `(:|$)` over-matches. Ledger's regex leaks siblings through the `$` alternation branch — `Assets:Receivable:Bob(:|$)` matched `Bobby`, `Bob:Car` **and** `Assets:Cash`. Split into two anchored patterns per root instead: `Root:name$` (the exact account) and `Root:name:` (its sub-accounts). Bob then nets to his own postings only, Bobby to his.

2. A single backslash is un-escaped back into a wildcard. Ledger un-escapes the arg once before its regex engine sees it, so `A\.C` still matched `AXC`. `A\\.C` (double backslash) or `A[.]C` matches only `A.C`. Went with the double-backslash escape.

Note on the expected figure: Bob nets to 60, not 50, because `Assets:Receivable:Bob:Car` ($10) is Bob's own sub-account and is intentionally folded in (matches `peopleFromBalance`/`PERSON_ACCOUNT`, which collapse sub-accounts into the person). Bobby's $40 is excluded, which was the actual leak. Added `personAccountPatterns` unit tests for the prefix-sibling and metacharacter cases.

PR #100, opened 2026-07-13
feat(journal): migrate Assets:Credited to the receivable/payable debts model

naeemba:
Companion to #99. Migrates the legacy single-account debts model
(`Assets:Credited:<person>`, one signed account) to the split model the
quick-entry **Debt** / **Settle up** features and the new debts page use.

## What it does
For each `Assets:Credited:<person>` account:
- net **≥ 0** (owed to you) → `Assets:Receivable:<person>`
- net **< 0** (you owe) → `Liabilities:Payable:<person>`

Ledger computes each account's net; the code only reads the **sign** to pick the
root — no summation (CLAUDE.md hard rule). It's a pure account **rename**, so
every transaction stays balanced and the change is reversible.

## Safety
- **Whole-account rewrite only**: boundary-guarded (`(?<![\w:]) … (?![\w:])`) and
  applied **longest-first**, so `Assets:Credited:Bob` can't corrupt
  `Assets:Credited:Bob Smith`, and `…:Al` can't corrupt `…:Alex`.
- **`balance --empty`** pulls zero-net accounts into the rename map, so a hidden
  zero-balance sibling is never left half-matched.
- Mirrors `relocateLegacyDefinitions`: per-user lock, pull fresh, **snapshot every
  touched file** (main + includes), verify the result parses, **roll back all** on
  any failure before pushing. Bails to `skipped` on an encrypted/locked journal so
  plaintext never overwrites ciphertext.
- **Idempotent** — `skipped` once no `Assets:Credited` account remains.

## Trigger
`POST /api/maintenance/migrate-credited-debts` (no UI, like the other
`/api/maintenance/*` repairs). Returns `{ result: 'migrated' | 'skipped' | 'failed' }`;
`migrated` includes the renamed accounts and occurrence count.

## Verification
Unit tests cover `targetAccount`, `planRenames`, and the boundary/longest-first
rewrite cases. Verified end-to-end against **ledger 3.4.1** on a synthetic legacy
journal (including a spaced name `Bob Smith` and a zero-net account): the result
parses, every transaction still balances, and it feeds the #99 debts view correctly
(owes-you / you-owe / hidden). `tsc --noEmit` + `vitest run lib/journal` green (98).

> Run this once after merging #99 if you have legacy `Assets:Credited` data.

### DIFF
--- a/app/api/maintenance/migrate-credited-debts/route.ts
+++ b/app/api/maintenance/migrate-credited-debts/route.ts
@@ -14,6 +14,10 @@ const log = createLogger('migrate-credited-debts');
  * rewritten journal parses, and rolls back every touched file on failure, so a
  * bad migration is fully reversible. We surface a rollback as a structured
  * `failed` result rather than an opaque 500.
+ *
+ * Returns a `manual` result (200) listing any legacy person accounts whose net
+ * spans multiple commodities: those have no single receivable/payable sign and
+ * must be resolved by hand before the migration can run.
  */
 export async function POST(): Promise<NextResponse> {
   const user = await requireUser();
--- a/app/api/maintenance/migrate-credited-debts/route.ts
+++ b/app/api/maintenance/migrate-credited-debts/route.ts
@@ -22,8 +22,10 @@ const log = createLogger('migrate-credited-debts');
 export async function POST(): Promise<NextResponse> {
   const user = await requireUser();
   try {
-    const result = await migrateCreditedToDebts(user.id);
-    return NextResponse.json({ result });
+    const { status, ...details } = await migrateCreditedToDebts(user.id);
+    // Keep `result` a flat string (like the relocate-definitions sibling and
+    // the `failed` error branch); spread any extras alongside it.
+    return NextResponse.json({ result: status, ...details });
   } catch (error) {
     log.error({ err: error }, 'credited debts migration failed');
     return NextResponse.json(
--- a/app/api/maintenance/migrate-credited-debts/route.ts
+++ b/app/api/maintenance/migrate-credited-debts/route.ts
@@ -1,0 +1,30 @@
+import { requireUser } from '@/lib/auth/require-user';
+import { migrateCreditedToDebts } from '@/lib/journal/creditedMigration';
+import { createLogger } from '@/lib/log';
+import { NextResponse } from 'next/server';
+
+const log = createLogger('migrate-credited-debts');
+
+/**
+ * One-shot migration of the legacy `Assets:Credited:<person>` debts model to the
+ * current `Assets:Receivable` / `Liabilities:Payable` split. Operates only on the
+ * caller's own journal. Idempotent — returns `skipped` once nothing remains.
+ *
+ * The service takes the per-user lock and pulls fresh under it, verifies the
+ * rewritten journal parses, and rolls back every touched file on failure, so a
+ * bad migration is fully reversible. We surface a rollback as a structured
+ * `failed` result rather than an opaque 500.
+ */
+export async function POST(): Promise<NextResponse> {
+  const user = await requireUser();
+  try {
+    const result = await migrateCreditedToDebts(user.id);
+    return NextResponse.json({ result });
+  } catch (error) {
+    log.error({ err: error }, 'credited debts migration failed');
+    return NextResponse.json(
+      { result: 'failed', message: 'Could not migrate your journal' },
+      { status: 500 }
+    );
+  }
+}
--- a/lib/journal/creditedMigration.ts
+++ b/lib/journal/creditedMigration.ts
@@ -4,7 +4,6 @@ import 'server-only';
 import { VALID_EXTS, GENERATED_PRICE_DB_NAME, PRICE_DB_NAME } from './layout';
 import { withUserLock } from './mutex';
 import { verifyJournalParseable } from './verify';
-import { parseBalanceRows } from '@/lib/balance/parse';
 import { journalRepository } from '@/lib/journal';
 import { pull, push, StorageConflictError } from '@/lib/storage';
 import { runLedgerForUser } from '@/utils/runLedgerForUser';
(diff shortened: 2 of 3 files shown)


lib/journal/creditedMigration.ts
```diff
@@ -0,0 +1,179 @@
...
+    // account is in the rename map — otherwise a hidden zero sibling like
+    // `Assets:Credited:Bob Smith` could be partially matched by an
+    // `Assets:Credited:Bob` rule.
+    const balance = await runLedgerForUser(userId, [
+      'balance',
+      '--empty',
+      '--flat',
+      '--no-total',
+      '--format',
+      '%(account)|%(quantity(scrub(total)))\n',
```
naeemba: `%(quantity(scrub(total)))` calls `quantity()` on the account's net balance, which **errors** when that balance holds more than one commodity: ledger 3.4.1 aborts with `Cannot convert a balance with multiple commodities to an amount` (exit 1). So if any single person's `Assets:Credited:<person>` net spans two currencies (e.g. lent someone `$50` and `€30`), the whole `runLedgerForUser` call throws, `planRenames` never runs, and the migration returns `failed` (500 with the opaque "Could not migrate your journal").

It fails safe (no journal is touched), so this isn't a corruption risk — but a user in that state can't migrate and gets no signal as to why. Worth either handling explicitly (surface a "multi-commodity account — migrate manually" result, since picking a single receivable/payable sign for a mixed-commodity net is a real valuation decision, not something to guess), or at minimum documenting the limitation.

Verified against ledger 3.4.1:
```
$ ledger -f synth.ledger balance --empty --flat --no-total \
    --format '%(account)|%(quantity(scrub(total)))\n' Assets:Credited
Error: Cannot convert a balance with multiple commodities to an amount
```

naeemba: Fixed in f6dad15.

Switched the format from `%(quantity(scrub(total)))` to raw `%(scrub(total))`, which exits 0 on a multi-commodity net instead of aborting the whole invocation. `planRenames` now reads the sign of each single-commodity net and collects any multi-commodity account (its second commodity arrives on a continuation line with no `|`) into a new `manual` list.

Since a mixed-commodity net has no single receivable/payable sign — picking one is a valuation decision, not something to guess — the migration returns `{ status: 'manual', accounts }`, touches nothing, and hands those accounts back so the user can resolve them and re-run cleanly. Added a test covering the multi-commodity continuation-line case.

lib/journal/creditedMigration.ts
```diff
@@ -0,0 +1,220 @@
...
+    const account = accountRaw.trim();
+    const amount = (amountRaw ?? '').trim();
+    current = null;
+    if (!account || account === 'Total' || !amount) continue;
+    if (!account.startsWith(`${LEGACY_ROOT}:`)) continue;
+    current = account;
+    if (manual.includes(account)) continue;
+    // A single-commodity net: negative iff a minus sign appears before the
+    // number (ledger renders `$ -20.00` / `-20.00 EUR`). Zero renders as `0`.
+    const sign = amount.includes('-') ? -1 : 1;
```
naeemba: `amount.includes(-)` treats *any* hyphen anywhere in the rendered amount as a negative net, but a hyphen can legally appear in a quoted commodity name — `10 "AK-47"`, `5 "T-BILL"`, etc. Ledger renders a positive net of such a commodity as `10 "AK-47"`, which contains a `-`, so the sign flips to `-1` and a balance that is genuinely **owed to you** gets routed to `Liabilities:Payable` instead of `Assets:Receivable`. Silent wrong-account migration, not a crash.

Verified against ledger 3.4.1:
```
$ ledger -f synth.ledger balance --empty --flat --no-total --format %(account)|%(scrub(total))\n Assets:Credited
Assets:Credited:Gun|10 "AK-47"
```
→ `planRenames` yields `Assets:Credited:Gun` → `Liabilities:Payable:Gun` (should be `Assets:Receivable:Gun`).

Detect the sign from the numeric part only, e.g. match a `-` immediately before a digit (`/-\s*\d/` after stripping the commodity) or, more robustly, ask ledger for the sign directly. A hyphenated commodity name is uncommon but fully valid, and this path already special-cases multi-commodity nets, so single-commodity hyphen-name nets deserve the same care.

app/api/maintenance/migrate-credited-debts/route.ts
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * Returns a `manual` result (200) listing any legacy person accounts whose net
+ * spans multiple commodities: those have no single receivable/payable sign and
+ * must be resolved by hand before the migration can run.
+ */
+export async function POST(): Promise<NextResponse> {
+  const user = await requireUser();
+  try {
+    const result = await migrateCreditedToDebts(user.id);
+    return NextResponse.json({ result });
```
naeemba: nit: envelope shape is inconsistent with the `relocate-definitions` reference and with this file's own doc-comment. The sibling service returns a plain string (`relocated | skipped`), so `{ result }` there is a string, matching the `{ result: failed }` error case. Here the service returns an *object* (`{ status: migrated, ... }`), so success is `{ result: { status: ... } }` while the error branch is `{ result: failed }` — a string vs object under the same key. The PR description also advertises `{ result: migrated | skipped | failed }` (flat strings), which the code does not produce. No consumer exists yet, so this is cosmetic, but a future caller has to special-case the two shapes. Consider returning `{ result: result.status, ... }` (spreading the extras) to keep the `result` field a string like the reference.

lib/journal/creditedMigration.ts
naeemba: Fixed in f9e16f7.

Confirmed against ledger 3.4.1: a positive net of a hyphenated commodity renders `10 "AK-47"` (hyphen in the name), while a genuine negative renders `-3 "T-BILL"` / `$ -20.00` (minus before the number). planRenames now strips any quoted commodity name before the sign check (`amount.replace(/"[^"]*"/g, "").includes("-")`), so `Assets:Credited:Gun` → `Assets:Receivable:Gun` as it should. Added a test covering both a positive and a negative hyphenated-commodity net.

app/api/maintenance/migrate-credited-debts/route.ts
naeemba: Fixed in 7c383df.

The route now destructures the status out and spreads any extras: `{ result: status, ...details }`. `result` is a flat string in every case (`migrated | skipped | manual | failed`), matching the relocate-definitions sibling, the `failed` error branch, and the doc-comment, so a future caller never has to special-case a string-vs-object shape.

PR #101, opened 2026-07-14
feat(transactions): unify transaction rows across lists with shared actions

naeemba:
Renders transaction rows with **one shared component** carrying consistent
edit/delete actions across the main Transactions list, the account register,
Dashboard recent, and Reconcile. Every surface maps its source into a common
`TransactionRowView`; the shared `TransactionRow` renders it identically and
mounts actions when the row has a uid.

Built via brainstorm → spec → plan → task-by-task implementation with review
gates. Spec + plan are included under `docs/superpowers/`.

## What changed
- **New:** `uidFromNote` (extract a uid from ledger `%(note)` text),
  `deleteTransactionByUid` (delete from a uid alone), `TransactionRowView` +
  adapters, the shared `TransactionRow` component, and an `AccountRegister`
  client wrapper.
- **Wired:** main list, account register, Dashboard recent, Reconcile now all
  render through `TransactionRow`; the old `TransactionRowItem` is removed.
- Aggregate reports (Payees/Balance/Debts/Monthly) are **untouched** — a row
  there is a rollup, not a transaction.

## How register rows got actions (no journal migration)
The journal stores the uid as a boolean tag (`; :uid: …`), which ledger's
`tag()` can't read. Instead the register queries add `%(note)`, which emits the
raw comment text, and `uidFromNote` pulls the id out — verified against ledger
3.4.1. Delete resolves the **per-transaction** fingerprint server-side via
`findTransaction` (the value `performDelete` actually compares), so rows never
carry a fingerprint. Edit + delete are universal; save-as-template shows only
where the full postings exist (the main list).

## Review notes
Whole-branch review (after fixes): unique per-posting row keys (a `ledger
register` emits one row per posting, so a 2-posting transaction repeats the
uid) and restored account-name links on Dashboard/Reconcile.

**Accepted follow-ups (non-blocking):**
- The account register's desktop **column headers** are gone — consistent with
  the headerless virtualized main list.
- Reconcile's **stale-red age coloring** isn't reproduced in the shared row (the
  raw age shows on mobile + desktop; the page-level "N older than X days" banner
  remains).
- The shared row's desktop grid has two near-identical templates (6-col vs
  7-col with running total) kept in sync by hand — noted in a `ponytail:` comment.

## Tests
Adapters, uid extraction, `deleteTransactionByUid`, and shared-row rendering are
unit-tested. Full suite: **1066 passing**, `tsc --noEmit` clean.


app/accounts/[account]/AccountRegister.tsx
```diff
@@ -0,0 +1,23 @@
+'use client';
```
Comment (naeemba) [inline app/accounts/[account]/AccountRegister.tsx]: This wrapper adds a `'use client'` boundary and a whole extra file for something the other two surfaces in this same PR render inline. Dashboard (`features/dashboard/Dashboard.tsx`) and Reconcile (`features/reconcile/Reconcile.tsx`) both `.map` over their rows into `<TransactionRow>` directly inside the server component — `TransactionRow` only reaches a client boundary at `RowActions`, so a server component can render it fine. `AccountRegister` needs no client state (it just maps static `views` + an empty state), so it could be inlined in `page.tsx` exactly like the sibling surfaces, dropping the file and the needless client boundary. Marking the whole `views` array (including `templateDraft`) as client props is avoidable too.

features/transactions/row/registerRows.ts
```diff
@@ -0,0 +1,22 @@
...
+    .map((chunk) => {
+      const cols = chunk.split('|');
+      const date = (cols[0] ?? '').trim();
+      const payee = (cols[1] ?? '').trim();
+      const amount = (cols[2] ?? '').trim();
+      const runningTotal = (cols[3] ?? '').trim();
+      const note = cols.slice(4).join('|');
+      const uid = uidFromNote(note) ?? undefined;
+      return { date, payee, amount, runningTotal, uid };
+    });
```
naeemba: nit: this is now the third copy of the same `split('NNN') → split('|') → trim fixed fields → `uidFromNote(cols.slice(n).join('|'))` skeleton. `parseRecentPostings` (`features/dashboard/Dashboard.utils.ts`) and `parseReconcileRows` (`features/reconcile/Reconcile.utils.ts`) are the other two, both added/extended in this PR. They also diverge on the malformed-line guard: those two filter `cols.length >= 4 && cols[0].trim()`, while this one only `filter(Boolean)`, so a short/blank register chunk here would render an empty row instead of being dropped. Worth a shared `splitRegisterRows(stdout)` helper that returns trimmed columns + extracted uid, with each surface mapping to its own shape on top.

app/accounts/[account]/AccountRegister.tsx
Comment (naeemba) [reply app/accounts/[account]/AccountRegister.tsx]: Done in 443ea28 — inlined the register rows into the server page component (same pattern as Dashboard/Reconcile) and deleted AccountRegister.tsx, dropping the 'use client' boundary and the client props.

features/transactions/row/registerRows.ts
naeemba: Done in cef4731 — extracted splitRegisterRows(stdout) returning trimmed cols + extracted uid. All three surfaces now map onto it, so they share the same cols.length >= 4 && cols[0].trim() guard (the account register's old filter(Boolean) that let short/blank chunks through is gone) and the note-is-the-remainder rule.

features/transactions/row/TransactionRow.tsx
```diff
@@ -0,0 +1,132 @@
...
+      {view.payee}
+    </Link>
+  ) : (
+    <span>{view.payee}</span>
+  );
+
+const actionsNode = (view: TransactionRowView) =>
+  view.uid ? (
+    <RowActions uid={view.uid} templateDraft={view.templateDraft} />
+  ) : null;
```
naeemba: nit: the shared `TransactionRow` drops the main list's old "no uid" affordance. The removed `TransactionRowItem` rendered a visible `no uid` span with `title="Re-import the journal to enable editing for this transaction"` for un-backfilled rows; here `actionsNode` returns `null` when there is no uid, so on the main list a legacy transaction now shows no actions and no hint about *why* it can't be edited. Silent on the register/dashboard/reconcile surfaces is fine (a bare register row needs no such hint), but on the main list that breadcrumb was useful. Unlike the other UX trims (headers, stale-red), this one isn't called out in the PR notes — worth either restoring the hint on the main list or noting it as an accepted drop.

naeemba: Done in 73b32b5 — restored the hint. `actionsNode` now renders the `no uid` span (with the "Re-import the journal…" tooltip) when there is no uid *and* `templateDraft` is present. Only `transactionRowToView` (the main list) supplies `templateDraft`, so the breadcrumb is back on the main list while the bare register/dashboard/reconcile rows stay silent. Added two tests covering both sides of that discriminator.

PR #102, opened 2026-07-14
feat(transactions): edit transactions through the simplified entry dialogs

naeemba:
## What & why

Editing a transaction now opens the **same simplified dialogs** used to create one from the header, instead of navigating to the full `/transactions/[uid]/edit` page. Clicking Edit (the row's ⋮ menu or the payee text) opens an in-place dialog that:

1. loads the transaction,
2. **detects its type** and opens the matching simplified form, prefilled, or
3. falls back to a **Raw editor** for shapes the simple forms can't faithfully represent.

The full-page edit route stays as the fallback target.

## How it works

- `detectType(draft)` already classifies a saved transaction and returns the exact field shape the simplified forms consume; each spec's `compile` **is** the matching adapter's `compile`, so edit is `load → detect → seed the form → compile → updateTransactionAction`. **No accounting math is added in JS** — everything routes through the existing adapters and ledger validates on save.
- The routing gate (`pickEditSurface`) is intentionally **stricter than "detected"**: anything carrying `extraItems` (splits), a cost/assertion the simple form can't show, or an undetectable multi-posting shape goes to **Raw**, so a saved transaction can never silently desync from the form used to edit it. A clean 2-posting exchange (which legitimately carries a `cost` annotation) correctly opens the **Exchange** form.
- The `expectedFingerprint` from the parser is carried through **unchanged** end-to-end, preserving the edit concurrency guard.

## Pieces

- `editTransactionStore.ts` — `useSyncExternalStore` trigger so any row opens one globally-mounted dialog (no Context threaded through every list).
- `editSurface.ts` — `pickEditSurface` routing gate.
- `actions/loadTransactionForEdit.ts` — server action mirroring the edit page's load.
- `QuickTypeForm.tsx` — extracted from `QuickEntry` so create and edit share one form.
- `TransactionEditDialog.tsx` — the dialog (simple form or Raw, one-way "Edit as raw" escape).
- Wiring: `row/EditPayeeTrigger.tsx`, `row/TransactionRow.tsx`, `RowActions.tsx`, `QuickEntrySlot.tsx`.

## Accepted simplifications

- A **debt** transaction opens as the **Transfer** form (debt isn't a registry adapter; still correct, loses the debt framing).
- The "Edit as raw" escape is **one-way** (reopen to reset).

## Testing

- `pickEditSurface` routing tests for all five simple types (form vs raw), including split→raw, cost-exchange→form, undetectable→raw.
- Round-trip tests (`detect → compile` reproduces the original postings, cost annotation included) for expense/income/transfer/exchange — the guard against silent seed↔compile desync. fixBalance excluded (its compile is not a pure inverse of detect; covered by routing).
- Full suite green; `tsc` and lint clean.

## Not yet done

- Interactive end-to-end app-drive (open `/transactions`, edit, save) — needs an authenticated session; verified by unit/integration tests and type/lint only.

Design spec and plan: `docs/superpowers/specs/2026-07-14-edit-via-quick-dialog-design.md`, `docs/superpowers/plans/2026-07-14-edit-via-quick-dialog.md`.


naeemba (commented):
### Code review

No blocking issues found. Reviewed for correctness and CLAUDE.md compliance.

Traced the flow end-to-end (store trigger → dialog → load action → `pickEditSurface`/`detectType` → `QuickTypeForm`/Raw → `updateTransactionAction`) against the canonical `EditTransaction.tsx` page.

Verified:
- No accounting math added — edit reuses the existing `detect`/`compile` adapters; ledger validates on save (CLAUDE.md hard rule).
- Routing gate is sound: adapter `id`s match spec `kind`s, and `hasSplits` + `detect`'s cost/assertion rejection correctly route splits and undetectable shapes to Raw while a clean cost-exchange opens the Exchange form. `editSurface.test.ts` covers all five simple types.
- `firstFieldError` dropping the old `.flat()` is equivalent since `fieldErrors` is `Record<string, string>`.
- `fingerprint` is carried through unchanged, preserving the concurrency guard.

features/transactions/editSurface.ts
```diff
@@ -0,0 +1,27 @@
...
+import { detectType } from './entry/types/registry';
+import { QUICK_ENTRY_SPECS, type QuickEntrySpec } from './quickEntrySpecs';
+
+export type EditSurface =
+  | { kind: 'type'; spec: QuickEntrySpec<HeaderFields>; fields: HeaderFields }
+  | { kind: 'raw'; seed?: DraftState };
+
+// A simplified spec form renders a single amount/account pair. `detect` returns
+// splits as `extraItems`, which those forms never show — so a detected shape
+// with any split must fall through to Raw instead of round-tripping invisibly.
```
naeemba: This rationale is incorrect, and it makes the gate more conservative than it needs to be.

`expenseSpec.Fields` (quickEntrySpecs.tsx:133) and `incomeSpec.Fields` (:194) both render `<ExtraItemsField>` — the simple forms **do** show and edit splits/deductions. So the premise "those forms never show splits" is false, and a normal income-with-a-deduction (salary + a tax expense posting) or an expense split-across-categories — both fully representable and editable in the simple form — gets forced to Raw here.

Two honest options:
1. Allow the split round-trip: drop `hasSplits`, and add a `detect → compile` round-trip test for the split/deduction case to `roundTrip.test.ts` (the existing tests only cover the no-split path via `balancingPostings`, so this path is currently unverified — which is presumably the *real* reason for the guard).
2. Keep routing splits to Raw, but rewrite the comment to state the actual reason (the split `compile` inverse is untested, so Raw is the safe default) rather than a claim about the forms that isn't true.

Either is fine; the current state ships a misleading comment plus an unnecessary UX limitation.

features/transactions/TransactionEditDialog.tsx
```diff
@@ -0,0 +1,182 @@
...
+ */
+export default function TransactionEditDialog() {
+  const uid = useEditTransactionUid();
+  const router = useRouter();
+  const [loaded, setLoaded] = useState<Loaded | null>(null);
+  const [surface, setSurface] = useState<EditSurface | null>(null);
+  const [notFound, setNotFound] = useState(false);
+
+  useEffect(() => {
+    if (!uid) {
```
naeemba: State is only reset when `uid` becomes null. On a direct `uid` A→B change (without passing through null), `loaded`/`surface` keep A's values while B loads. `QuickTypeForm`'s `key` is `uid`, so it remounts once when B renders (still showing A's data mid-load) — but the key stays `B` when B resolves, so `useState(initialFields)` never re-seeds and the form keeps **A's field values inside B's spec**.

Currently masked because the modal blocks background clicks, so A→B isn't reachable through the UI today — but it's latent and a trivial fix: reset at the top of every run, not just the null branch:
```ts
useEffect(() => {
  setLoaded(null);
  setSurface(null);
  setNotFound(false);
  if (!uid) return;
  let cancelled = false;
  ...
}, [uid]);
```

features/transactions/editSurface.test.ts
```diff
@@ -0,0 +1,99 @@
...
+    expect(surface.kind).toBe('type');
+    if (surface.kind === 'type') expect(surface.spec.kind).toBe('income');
+  });
+
+  it('routes a plain transfer pair to the transfer form', () => {
+    const surface = pickEditSurface(
+      draftOf([
+        { account: 'Assets:Savings', amount: '100', currency: 'USD' },
+        { account: 'Assets:Checking', amount: '-100', currency: 'USD' },
+      ])
```
naeemba, line 79: Routing coverage is solid for the five detectable adapters, but three cases the PR explicitly relies on aren't asserted:

- **debt → transfer**: the accepted simplification (a `Assets:Receivable:X` / `Liabilities:Payable:X` pair opens the Transfer form). Worth a test that it routes to `transfer` *and* that a save round-trips the receivable/payable account back (this is the case most likely to silently corrupt a balance if `transferAdapter` ever changes).
- **refund → raw**: a refund (asset + expense posting, no income posting) is detected by neither expense (negative base) nor income (needs exactly one income posting), so it falls to Raw. Fine, but untested — a regression in `classifyAccount` could flip it to expense and invert a sign.
- **fix-balance after save+parse**: the current fix-balance test uses a synthetic draft with `amount: ''`. `fixBalanceAdapter.detect` requires the adjustment posting to have `amount === ''`, which only holds if the journal writer elides it. No test loads a real, parsed fix-balance transaction to confirm it still detects (vs. quietly routing to Raw).

naeemba (commented):
Reviewed the edit-via-dialog feature. The core wiring is clean and matches existing patterns well: `loadTransactionForEdit.ts` mirrors the sibling `loadTransactionPage.ts` / `EditTransaction.tsx` faithfully, the fingerprint is carried through unchanged, `headerOf`/`compile` preserve `status`+`note`+`uid` so a simple-form edit never silently drops them (good — I checked this specifically), and the module-level client store is the right call over threading Context (no cross-request state, so no `globalThis` pin needed unlike the server-side singletons).

No blocking correctness bugs, but three things worth addressing (inline). Highest-value: the routing gate's stated rationale is factually wrong — the simple expense/income forms **do** render `ExtraItemsField`, so splits/deductions are editable there, yet every split is forced to Raw. Two smaller items: a latent stale-form state bug on a direct uid→uid switch, and some routing test gaps.

Minor (not inline): `loadTransactionForEdit.ts` coalesces `fingerprint ?? ''` and always sends the field, whereas the full-page path omits it entirely when falsy (which would fail with "Missing edit payload"). Harmless because the parser always stamps a fingerprint, but it's a silent behavioral divergence from the reference path.

As the PR notes, the interactive save-and-reload e2e was not run — that's the main residual risk since the detect→compile inverse is only unit-tested.

features/transactions/editSurface.ts
naeemba: Fixed in 129b065 — took option 2 (keep splits→Raw, correct the comment), but with a sharper reason than either the old comment or your option-1 premise.

You're right the forms render `ExtraItemsField`, so the "those forms never show splits" claim was false. But option 1 (drop `hasSplits`, add a `detect → compile` round-trip test) doesn't actually hold: I verified it. A split `compile` emits an **amount-less auto-balance line** (`balancingPostings`) where a parsed journal transaction carries an explicit paying amount, so the exact round trip fails:

```
× split with explicit paying amount round-trips exactly?
-  "account": "Assets:Cash", "amount": "-12"
+  "account": "Assets:Cash", "amount": ""
```

That's the same reason `fixBalance` is excluded from `roundTrip.test.ts` — compile isn't a pure inverse. So splits stay on Raw until that path is verified against a real saved+parsed split, and the comment now states exactly that instead of the false forms claim.

features/transactions/TransactionEditDialog.tsx
naeemba: Fixed in 129b065 — applied your suggestion: reset `loaded`/`surface`/`notFound` at the top of every effect run, guarded only by the `if (!uid) return` after. A direct A→B switch now clears A's state before B loads. Comment records the latent-but-masked nature (modal blocks background clicks today).

features/transactions/editSurface.test.ts
naeemba: Fixed in 129b065 — added all three:

- **debt → transfer**: routing test in `editSurface.test.ts` + a `detect → compile` round-trip in `roundTrip.test.ts` asserting the receivable/payable pair rebuilds back to its accounts (this pair round-trips exactly — explicit amounts, no split).
- **refund → raw**: asset + negative-expense pair, asserts `kind === 'raw'`.
- **fix-balance after parse**: routes the real parser's postings (`parseBlock('… Assets:Checking = 500 USD / Equity:Adjustments')`) — assertion line kept as `amount: ''` per `parser.ts` — through `pickEditSurface`, confirming it detects instead of falling to Raw. This closes the gap where the old synthetic test hard-coded the very `amount: ''` that detect keys on.

features/transactions/editSurface.ts
naeemba: Verified — resolved. The corrected comment is accurate: `balancingPostings` (extraItems.ts:44-51) emits `{ account, amount: '', currency: '' }`, so split `compile` really does drop the explicit paying amount and detect→compile is not a pure inverse. Keeping splits→Raw (option 2) is the defensible conservative call; the false premise is gone. No behavior change, rationale now correct.

features/transactions/TransactionEditDialog.tsx
naeemba: Verified — resolved. Reset now runs at the top of every effect run, guarded by `if (!uid) return` after, so a direct A→B switch clears A's `loaded`/`surface`/`notFound` before B loads. The `cancelled` closure still drops A's in-flight resolve. tsc + lint clean (the single eslint-disable still covers the block).

features/transactions/editSurface.test.ts
naeemba: Verified — resolved. All three added and green: debt→transfer routing (editSurface.test.ts) + an exact detect→compile round trip for the receivable/payable pair (roundTrip.test.ts), refund→raw, and a real `parseBlock` fix-balance routed through `pickEditSurface` (assertion line survives as `amount: ''`). 15/15 pass.

PR #103, opened 2026-07-15
fix(transactions): stop split-category row overflowing the quick dialog

naeemba:
## Problem

Adding a split category in the quick-entry Expense dialog produced a horizontal scrollbar and cut off the split row's Amount/Currency fields.

## Root cause

`ExtraItemsField` keyed its row layout off **viewport** breakpoints (`sm:`). The quick-entry dialog is `max-w-sm` (~352px content), but on any desktop viewport (≥640px) the `sm:` horizontal 4-column layout (`1fr_140px_90px_auto`) activated regardless of the dialog's actual width — and the `1fr` grid track won't shrink below its content's intrinsic size, so the row overflowed its container.

## Fix

Switch the row to **container queries** (`@container` on the section, `@sm:` on the row) so it responds to its own width instead of the viewport: stacked inside the narrow dialog, horizontal in the wide entry form as before. Also `1fr` → `minmax(0,1fr)` so the account column can shrink as insurance.

One file, class-only change. Tailwind v4's native container queries — no new dependency.

## Note

`FormLens.tsx` has the same row markup but only renders in the wide entry page, not a narrow dialog, so it isn't affected and was left unchanged.

### DIFF
--- a/features/transactions/entry/FormLens.tsx
+++ b/features/transactions/entry/FormLens.tsx
@@ -125,7 +125,7 @@ export function FormLens({
           <BalanceIndicator balance={balance} />
         </div>
 
-        <div className="flex flex-col gap-2">
+        <div className="@container flex flex-col gap-2">
           {draft.postings.map((posting, idx) => (
             <PostingRow
               key={idx}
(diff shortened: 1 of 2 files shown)


naeemba (commented):
Verified the container-query switch end to end.

- Dialog: `DialogContent` is `sm:max-w-sm` (384px) minus `p-4` -> ~352px section width, below the `@sm` container breakpoint (24rem/384px), so the row stays stacked. Correct fix.
- Wide entry page: section sits in `AppShell`'s `max-w-7xl` column, well above 384px on desktop -> horizontal layout preserved; stacks on mobile as before.
- Tailwind v4.3.2 has container queries built in, so no plugin needed, and `minmax(0,1fr)` is sensible insurance on the account track.

One non-blocking note: `FormLens.tsx:181` carries the same row markup still keyed off viewport `sm:`. It only renders in the wide form, never a narrow dialog, so it doesn't need this fix, but the two now diverge in responsive strategy. Fine to leave as documented in the PR description.

PR #104, opened 2026-07-15
fix(transactions): open the type form for round-trippable splits

naeemba:
## Problem

Editing a split transaction (e.g. an expense with a second `Expenses:` category) always opened the **raw** editor instead of the Expense form. Reported case:

```
2026-07-14 cigarette
    ; :uid: 01KXJPPMZYQ4VZTVNTEHCV5DA1
    Expenses:Cigarette_Alcohol   Kirt 300
    Expenses:Wage                Kirt 0.9
    Assets:Bank:Blubank
```

`pickEditSurface` detected it correctly as an expense, then a blanket `hasSplits` fallback routed every split to raw.

## Fix

The blanket fallback was a proxy for "compile can't round-trip this." Replace it with the real check: route to the type form only when `compile(detect(draft))` reproduces the draft's postings **exactly**.

- A split written by the entry form leaves its paying line **amount-less** (ledger balances it), and `detect → compile` reproduces that shape byte-for-byte → safe to edit in the Expense/Income form (both render `ExtraItemsField`).
- A split whose paying line carries an **explicit** amount still falls back to raw — split compile would rewrite it amount-less, so the exact round-trip fails.
- Ordering is significant on purpose: a differently-ordered hand-edited transaction also falls to raw instead of being reshuffled on save.
- `fix-balance` stays exempt — its form re-derives the balance from a target rather than reproducing postings (see `fixBalance.ts` / `roundTrip.test.ts`).

No corruption is possible: if the exact round-trip doesn't hold, it goes to raw.

## Tests

- New: amount-less split expense → Expense form (the reported case).
- Kept: explicit-paying-amount split → raw.
- All existing `editSurface` / `roundTrip` routing tests unchanged and green (236 transaction tests pass).

### DIFF
--- a/features/transactions/editSurface.test.ts
+++ b/features/transactions/editSurface.test.ts
@@ -22,7 +22,10 @@ describe('pickEditSurface', () => {
     if (surface.kind === 'type') expect(surface.spec.kind).toBe('expense');
   });
 
-  it('routes a split expense to raw (simple form would hide the split)', () => {
+  it('routes a split expense with an EXPLICIT paying amount to raw', () => {
+    // Split compile leaves the paying line amount-less, so it cannot reproduce
+    // this explicit `-12` — round-trip fails and it falls back to raw rather
+    // than silently rewriting that posting.
     const surface = pickEditSurface(
       draftOf([
         { account: 'Expenses:Food', amount: '10', currency: 'USD' },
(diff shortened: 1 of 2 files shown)


naeemba (commented):
### Code review

No issues found. Verified the round-trip routing logic end to end:

- `roundTrips` compares `compile(detect(draft))` to the draft's postings exactly, and no non-fix-balance adapter's `compile` reads `ctx`, so the routing check uses the same output the save path produces — an unchanged re-save of a round-tripping draft is provably identical to the journal. Corruption (route-to-form-then-rewrite) can't occur.
- The reported amount-less split expense round-trips and routes to the Expense form; explicit-paying-amount splits fail the exact match and correctly fall to raw.
- `balancingPostings` returning `[]` on a zero-residual all-explicit split yields a length mismatch, so that too falls to raw.
- fix-balance is correctly exempted since its `compile` re-derives from the assertion target rather than inverting the postings.

Checked for bugs and CLAUDE.md compliance.

PR #105, opened 2026-07-15
feat(currencies): commodities definitions CRUD tab

naeemba:
## What

Adds a **Commodities** tab to `/currencies` for managing ledger `commodity` directive blocks in `definitions.ledger` — no more download-backup → hand-edit → re-import round trip to change a display precision.

- Table of every commodity observed in the journal, joined with parsed definition blocks: editable rows (dialog), raw-block editor for blocks the form can't model (opaque), read-only rows for definitions living in other files, "Add definition" for undefined symbols.
- Form fields: note, aliases, display decimals (compiled to a `format` line — no raw format strings), `nomarket`, `default` (setting it clears the previous holder).
- Writes splice exact line spans under the per-user lock: pull → snapshot → mutate → `ledger stats` verify → rollback on rejection → push. Non-commodity lines (`account` directives, top-level aliases, comments) between blocks are preserved byte-for-byte (regression-tested).
- One server action per file (Zod-validated, rate-limited, audited); raw editor rejects `include` directives.

Also rides along: `fix(transactions)` — amount-less balancing postings no longer force the edit dialog to the raw editor (currency comparison skipped when no amount), fixing Expenses-created splits opening as raw text.

## Why

A `format KIRT 1,000` block (zero display decimals) made every register render `Kirt 0.9` as `KIRT 1`; there was no in-app way to fix it.

## Testing

- Parser round-trip, service CRUD + default handover + rollback against real ledger 3.4.1, interleaved-content preservation regression, and an end-to-end precision test: decimals 0 → `ledger reg` prints `KIRT 1`; decimals 1 → `KIRT 0.9`.
- Full suite 1139/1139, tsc and lint clean.
- Outstanding: one manual pass over `/currencies` in a browser (tabs, read-only legacy rows, decimals round-trip) — flagged for post-merge smoke.

Spec: `docs/superpowers/specs/2026-07-15-commodities-crud-design.md`

Follow-up filed here: confine ledger `include` resolution to the user's journal directory (pre-existing gap, also reachable via journal upload).

### DIFF
--- a/lib/commodities/blocks.ts
+++ b/lib/commodities/blocks.ts
@@ -23,11 +23,14 @@ const needsQuoting = (symbol: string): boolean => /[\d.,;\s"-]/.test(symbol);
 const renderSymbol = (symbol: string): string =>
   needsQuoting(symbol) ? `"${symbol}"` : symbol;
 
-const NUMBER_TOKEN = /-?[\d,]+(?:\.(\d+))?/;
+// Only a thousands-grouped sample is editable: serialization always emits
+// `1,000...`, so accepting a non-grouped sample (`100000`) would silently
+// switch the commodity to grouped display on the first edit.
+const GROUPED_NUMBER_TOKEN = /-?\d{1,3}(?:,\d{3})+(?:\.(\d+))?/;
 
-/** Decimal places of a format sample, or null when it has no numeric token. */
+/** Decimal places of a format sample, or null when it has no grouped numeric token. */
 const sampleDecimalPlaces = (sample: string): number | null => {
-  const match = NUMBER_TOKEN.exec(sample);
+  const match = GROUPED_NUMBER_TOKEN.exec(sample);
   if (!match) return null;
   return match[1]?.length ?? 0;
 };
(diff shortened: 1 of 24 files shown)


lib/commodities/blocks.ts
```diff
@@ -0,0 +1,128 @@
...
+  needsQuoting(symbol) ? `"${symbol}"` : symbol;
+
+const NUMBER_TOKEN = /-?[\d,]+(?:\.(\d+))?/;
+
+/** Decimal places of a format sample, or null when it has no numeric token. */
+const sampleDecimalPlaces = (sample: string): number | null => {
+  const match = NUMBER_TOKEN.exec(sample);
+  if (!match) return null;
+  return match[1]?.length ?? 0;
+};
```
naeemba, line 36: nit: `sampleDecimalPlaces` extracts only the decimal count and discards whether the sample uses a thousands separator. The design doc lists "no thousands group" as an opaque trigger, but a `format KIRT 100000` (no comma) parses as `decimalPlaces: 0`, editable, and re-serializes via `formatSample` as `1,000` — silently switching that commodity to grouped display on the first form edit. Same for a non-grouped sample with decimals (`1000.5` -> `1,000.0`). Only bites hand-written non-grouped format lines and ledger still parses the result, so low severity — but it is exactly the silent representation change the opaque path exists to prevent. Consider marking a sample without a thousands group opaque.

naeemba: Fixed in c2c8ccb — a sample whose numeric token lacks a thousands group is now parsed as opaque, so non-grouped hand-written format lines round-trip untouched.

PR #108, opened 2026-07-15
feat(transactions): one flow-oriented row per transaction

naeemba:
## Problem

The dashboard's "Recent transactions" list rendered **one row per ledger posting**, so a single transaction showed up as several unrelated rows (a `cigarette` transaction became three rows: two expense legs + the bank leg). And the accounts summary joined the first two postings in journal order with `→`, producing a misleading arrow (`Expenses:Cigarette_Alcohol → Expenses:Wage …`) whose long prefixes truncated away the one leg a human scans for — where the money came from.

## Change

- **Dashboard now shows one row per transaction.** It routes through the same grouped pipeline the main `/transactions` list already uses (`loadJournalTransactions` → `pageTransactions` → `transactionRowToView`) instead of parsing raw `reg` postings. `RECENT_LIMIT` now counts transactions, not postings. Removed the now-dead `getRecentTransactions`/`parseRecentPostings` helpers.
- **Reworked `accountsSummary()`** to split postings into `source → destination` by sign and render **leaf account names**, so the arrow reflects real money flow and the useful legs fit:

  | Before | After |
  |---|---|
  | `Expenses:Cigarette_Alcohol → Expenses:Wage …` | `Blubank → Cigarette_Alcohol, Wage` |
  | `Assets:Crypto:Tabdeal → Assets:Bank:Blubank` | `Tabdeal → Blubank` |

  Overflow beyond two names per side collapses to `+N`; degenerate cases (single posting, unparseable amounts) fall back to a plain leaf list. This flows to every surface using `accountsSummary` (dashboard + main list).

## Ledger safety

Only reads the signs ledger already assigned to parsed postings — no summing across currencies, no conversion in JS.

## Tests

Updated affected assertions; added focused `accountsSummary` cases (multi-destination, transfer, `+N` overflow, fallback). Type-check + suite green.

### DIFF
--- a/lib/transactions/model.ts
+++ b/lib/transactions/model.ts
@@ -289,10 +289,35 @@ export class Transaction {
     return computeBalance(this.postings);
   }
 
-  /** First two posting accounts joined for a compact list summary. */
+  /**
+   * Compact `source → destination` summary using leaf account names. Sides are
+   * split by posting sign (money out vs money in) so the arrow reflects the
+   * real flow — `Blubank → Cigarette_Alcohol, Wage` — rather than journal
+   * order. Leaf names keep the useful legs from being truncated away behind
+   * long `Expenses:`/`Assets:Bank:` prefixes. Not accounting math: it only
+   * reads the signs ledger already assigned, and never sums across currencies.
+   */
   accountsSummary(): string {
-    const accounts = this.postings.slice(0, 2).map((p) => p.account);
-    return `${accounts.join(' → ')}${this.postings.length > 2 ? ' …' : ''}`;
+    const leaf = (account: string) => account.split(':').pop() || account;
+    const side = (accounts: string[]) => {
+      const names = accounts.map(leaf);
+      return names.length > 2
+        ? `${names.slice(0, 2).join(', ')} +${names.length - 2}`
+        : names.join(', ');
+    };
+    const sources: string[] = [];
+    const destinations: string[] = [];
+    for (const p of this.postings) {
+      const value = Number(p.amount);
+      if (value < 0) sources.push(p.account);
+      else if (value > 0) destinations.push(p.account);
+    }
+    if (sources.length && destinations.length) {
+      return `${side(sources)} → ${side(destinations)}`;
+    }
+    // No clear two sides (single posting, or amounts that don't parse):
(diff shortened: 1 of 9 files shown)


lib/transactions/model.ts
```diff
@@ -289,10 +289,35 @@ export class Transaction {
...
+      return names.length > 2
+        ? `${names.slice(0, 2).join(', ')} +${names.length - 2}`
+        : names.join(', ');
+    };
+    const sources: string[] = [];
+    const destinations: string[] = [];
+    for (const p of this.postings) {
+      const value = Number(p.amount);
+      if (value < 0) sources.push(p.account);
+      else if (value > 0) destinations.push(p.account);
```
naeemba, line 314: Bare auto-balanced postings drop out of the source/destination split, so the arrow silently disappears for the most common ledger style.

`parsePostingLine` returns `amount: ''` for a bare balancing posting (see `lib/journal/parser.test.ts` "returns blank amount for bare-account auto-balance"), and this display path reads straight from that parser via `loadJournalTransactions`. Here `Number('')` is `0`, so the elided leg matches neither `value < 0` nor `value > 0` and is excluded from both `sources` and `destinations`.

For a typical transaction:

```
2026/01/01 Coffee
    Expenses:Coffee   $5
    Assets:Checking            ; elided, auto-balanced
```

`sources` ends up empty, so the `sources.length && destinations.length` branch is skipped and the summary falls back to the plain leaf list (`Coffee, Checking`) — no `→`. The new feature only renders the intended `Checking → Coffee` arrow when every posting carries an explicit amount, which idiomatic ledger journals routinely do not. Consider treating a blank/NaN amount on an otherwise two-sided transaction as the balancing (opposite-sign) side, or deriving the missing sign from the per-currency residual.

naeemba: Fixed in 9891ab2. A bare auto-balanced posting parses to `amount: ''` (Number('') = 0), so it was excluded from both flow sides and the arrow vanished. Now the elided leg is collected separately and assigned to whichever side is empty — if every explicit posting is money-out it settles as money-in, and vice versa — so `Checking → Coffee` renders for the common one-elided-leg style. Added a test covering it.

PR #111, opened 2026-07-16
Recurring transactions + research-driven dashboard widgets

naeemba:
## What

Two stacked features driving the dashboard toward "am I okay right now?" (per the July 2026 consumer-dashboard research; prioritization recorded in PLAN.md Phase 6):

### Recurring transactions (`/recurring`)
- Ledger periodic (`~`) directives as first-class app objects: `lib/journal/recurring.ts` (parse/format/Zod schema) with the same `; :uid:` identity + sha256 fingerprint conventions as dated transactions. Kept separate from `parseJournalFile` so no `ParsedTransaction` consumer sees undated rows.
- `JournalService.listRecurring/addRecurring/deleteRecurring` reuse the full write pipeline: per-user lock → Garage pull → append/splice → real-ledger verify with rollback → push. A bad period expression is rejected by `ledger stats` and rolled back.
- New `recurring.add`/`recurring.delete` audit actions; rate-limited server actions; nav entry under Journal.
- Editing is delete + re-add for now; imported `~` blocks without a uid show as read-only "no uid" rows.

### Dashboard widgets (all math in ledger, JS only parses)
- **Safe to Spend hero** — `bal ^Assets and not ^<PORTFOLIO_ACCOUNT_PREFIX> --forecast -e <cutoff>` where the cutoff is the next forecast `^Income` date (fallback +30d). Forecast bill postings debit the assets, so the balance at the cutoff *is* the projected spendable figure — no JS subtraction.
- **Upcoming bills** — `reg ^Expenses --forecast` bounded to the next 30 days.
- **Net-worth sparkline** — 12-month reuse of the /net-worth query (`--display` window, never `-b`).
- **Last month in review** — income/expenses/net from `getCashFlow` (`--invert` for income sign); net-worth delta as a single period-limited `bal ^Assets ^Liabilities`.
- Year-balance card removed; month card relabeled "Spent This Month".

## Verified ledger gotchas (3.4.1, tested on synthetic journals)
- `ledger print` silently drops `~` directives — no current write path uses print round-trips, but noted.
- Under `--forecast`, `Monthly from <date>` fires on the **1st**; only `weekly`/`every N weeks` keep the day-of-month anchor. Amounts are counted once per period so Safe to Spend stays correct; displayed due dates snap to period starts.
- Forecast rows only generate after the journal's last real transaction date, so already-posted bills don't reappear as upcoming.

## Testing
- New unit tests for the recurring parser/format/schema round-trip; full suite 1118 green.
- Every ledger command verified against a synthetic multi-commodity journal before use.
- lint, tsc, and `next build` clean. Live e2e in the dev app not yet run (local Postgres was down) — worth a manual pass on /recurring and the dashboard before merge.


features/dashboard/Dashboard.tsx
```diff
@@ -162,6 +208,166 @@ const Dashboard = async () => {
...
+              No activity recorded last month.
+            </p>
+          ) : (
+            <dl className="grid grid-cols-2 gap-x-4 gap-y-3 text-sm sm:grid-cols-4">
+              <div className="flex flex-col gap-1">
+                <dt className="text-xs font-medium uppercase tracking-wider text-muted-foreground">
+                  Income
+                </dt>
+                <dd className="font-semibold tabular-nums text-positive">
+                  {formatNumber(lastMonthReview.income)}
```
naeemba: nit: income/expenses/net render with `formatNumber` (bare number, no currency), while the adjacent "Net worth Δ" uses `formatAmount` and shows the unit. On a card whose values are ledger-converted to the base currency, a unit-less number next to a unit-ed one reads as inconsistent — consider appending the currency code (or reusing `formatAmount`) for the three figures too.

features/recurring/RecurringView.tsx
```diff
@@ -0,0 +1,285 @@
...
+              onChange={(e) => setNote(e.target.value)}
+              placeholder="Netflix subscription"
+            />
+          </div>
+        </div>
+
+        <div className="space-y-2">
+          {postings.map((p, i) => (
+            <div
+              key={i}
```
naeemba: nit: `key={i}` for the posting rows keys by array index while the list is mutable (add/remove posting). Removing a middle row makes React reconcile inputs to the wrong entries; the state array is the source of truth so values stay correct, but focus/uncontrolled DOM state can jump. A stable per-row id would be cleaner.

PR #112, opened 2026-07-16
Dashboard widget preferences + adjustment-row fix + layout memoization

naeemba:
## What

**Widget preferences (closes the PLAN.md dashboard-revamp task)**
- New `userSetting.dashboardWidgets` column (migration 0014): comma-joined widget order, `-` prefix marks a hidden section (e.g. `stats,-savedViews,trends`).
- `lib/dashboard/widgets.ts`: the six widget ids, labels, zod schema, and parse/serialize with normalization — unknown ids dropped, widgets missing from a saved order appended visible so future widgets show up for users with older preferences.
- Cached `getDashboardWidgets()` getter degrades to the default layout on read failure (mirrors `getEntryTabOrder`); rate-limited `setDashboardWidgetsAction`.
- Settings gains a "Dashboard widgets" card with move up/down and Show/Hide per section.
- `Dashboard.tsx` renders its sections from a `Record<WidgetId, ReactNode>` map in the user's saved order, skipping hidden ones. Data for hidden widgets is still fetched (marked as deliberate — revisit only if ledger spawns get slow).

**JOURNAL-AUDIT A5**
- `getHighestExpense` now skips register rows whose account starts with `<` — the synthetic `<Adjustment>`/`<Revalued>` rows `-X` conversion injects could previously win "Highest Expense This Month".

**REVIEW E4**
- `cachedEnsureLayout` (React `cache()`, same pattern as `cachedPull`) dedupes journal-layout resolution per request; `getFingerprint` and `runLedger` share it, dropping a dashboard render from ~14 identical `userSetting` selects to 1. Write paths keep the uncached methods so layout changes stay visible within a request.

Both audit docs updated to mark the findings fixed.

## Testing

- New tests: widget parse/serialize round-trip + normalization, Settings card render, `<Adjustment>` skip.
- Full suite: 1123 tests pass; type-check and lint clean.


features/settings/DashboardWidgetsForm.tsx
```diff
@@ -0,0 +1,101 @@
...
+): WidgetSetting[] => {
+  if (to < 0 || to >= arr.length) return arr;
+  const next = [...arr];
+  const [item] = next.splice(from, 1);
+  next.splice(to, 0, item);
+  return next;
+};
+
+const serialize = (widgets: WidgetSetting[]) =>
+  widgets.map(({ id, hidden }) => (hidden ? `-${id}` : id)).join(',');
```
naeemba: nit: this local `serialize` re-implements `serializeDashboardWidgets` from `@/lib/dashboard/widgets` (already imported here for its types). Reuse the exported one so the two serialization formats can not drift — the lib version also normalizes, which keeps the dirty check honest if `initial` ever arrives un-normalized.

PR #114, opened 2026-07-17
feat(budgets): allowance-style budgets computed by ledger

naeemba:
## Summary

Allowance-style budgets (the Monarch/Lunch-Money/Actual-Tracking model — research-verified as the right fit for a journal-as-truth design; envelope semantics are out of scope by design since their stateful math can't be delegated to ledger):

- **Budget lines are `; :budget:`-tagged `~` periodic directives** in the journal, created with the same structured-schedule form, uid/fingerprint, and verify/rollback write path as recurring bills. Plain ledger parses everything (tags are comments).
- **Ledger computes the whole report.** The `budget` report with a `get_at(display_total, …)` format emits account | actual | budgeted | difference (difference computed inside ledger via value expressions), `bal --unbudgeted` catches spending with no allowance, and the YTD cumulative column is the `budget` command over a Jan-1 span. JS parses pipe rows and renders; the sole numeric use is `usedRatio` for progress-bar geometry, taken from ledger-emitted `quantity()` fields.
- **Recurring bills count as allowances automatically** (Actual's schedule-template pattern, a deliberate spec decision) — the report labels bill-derived rows "from your bills"; explicit budget lines stack on top.
- **/budgets page**: create form (Expenses accounts enforced server-side), budget-vs-actual table with progress bars and YTD over/under, uid-keyed "Your budget lines" management (every line individually deletable even when accounts collide), unbudgeted section. **Dashboard** gains a Budgets widget (top categories by % used).
- **Guard rails**: budget rules are excluded from the recurring due queue, can never be posted/skipped into transactions, and deletes are kind-scoped (`ruleKind`) so a budgets-page delete can't remove a bill or vice versa. Each guard carries a regression test.

Empirical ledger 3.4.1 findings documented in LEDGER-AUDIT.md: `get_at(display_total, 1)` is the negated allowance; `neg()` fails in budget format context (use `0 - …`); `reg --budget`'s running total is global across accounts, not per-account.

## Test plan

- 1170 tests green; tsc, lint, `pnpm build` clean.
- New coverage: `:budget:` round-trip + fingerprint participation, due-queue/post-skip exclusion, kind-guarded deletes (both directions), report parsers incl. real-ledger integration tests for both the budget format and `--unbudgeted` (grand-total row handling), same-account line-collision component test, Expenses-account validation.
- Real-ledger sanity: journal with a budget line + bill + transactions parses via `ledger stats`; the budget query shows both allowance sources.
- [ ] Manual walk (needs authenticated session): create a budget on /budgets → row appears with correct amounts → dashboard widget shows it → delete line → row reverts to bill-derived/empty.

### DIFF
--- a/LEDGER-AUDIT.md
+++ b/LEDGER-AUDIT.md
@@ -203,3 +203,52 @@ ledger 3.4.1-20251025 on synthetic `~` periodic-transaction journals.
    confirm-to-post state are plain ledger comments — `ledger` ignores them
    entirely (confirmed via `stats`/`bal` on a journal containing both), so
    journals stay portable to any other ledger-reading tool.
+
+---
+
+## Budget report `--format` and running-total gotchas — 2026-07-17
+
+Verified while finishing the budgets feature (Task 8) against a synthetic
+journal with a `:budget:`-tagged periodic transaction (`~ every 1 months from
+2026/07/01`), a recurring bill rule also tagged `:budget:` and carrying a
+`; :handled:` marker, and three real July transactions on
+`Expenses:Groceries` / `Expenses:Rent`.
+
+1. **`budget`'s `--format` needs `get_at(display_total, 0|1)`, and element 1
+   is the *negated* allowance.** `ledger budget ^Expenses --flat --format
+   '%A|%(get_at(display_total, 0))|%(get_at(display_total, 1))\n'` returns
+   the actual in slot 0 and the allowance as a negative amount in slot 1 (a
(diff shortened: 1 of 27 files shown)


naeemba:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance — accounting math stays in ledger (`budget`/`bal --unbudgeted` queries), the only JS numeric use is the progress-bar ratio (display geometry, explicitly allowed), kind-scoped delete/post/skip guards are symmetric, and the `:budget:` tag round-trips through format/parse without polluting the note or the due queue.

PR #115, opened 2026-07-19
fix: robust register separators + reusable add-transaction button

naeemba:
## Invalid Date on account register

Register surfaces (account register at `/accounts/[account]`, reconcile) split `ledger reg --format` output on the literal string **`NNN`**. A ULID uid — emitted in `%(note)` — uses an alphabet that includes the letter **N**, so a uid like `01HZX`**`NNN`**`8E5T…` split mid-record: the trailing chunk shifted into an **"Invalid Date"** row and the uid was silently dropped. Free-text notes/payees containing `NNN` hit the same fault.

Fix: delimit records/fields with ASCII **RS (0x1e)** and **US (0x1f)**, which can never appear in a date, amount, ULID, or a normally-typed note. `parseAccountRegister` and `parseReconcileRows` now compose their format from the shared `registerFormat()` helper so both stay in lockstep.

Regression test added: a row whose uid contains `NNN` parses intact.

> Scope note: other ad-hoc `NNN`-split parsers (payees, net worth, monthly, periodic balance) emit only `%A/%P/%D/%t/%T` — no uid — so they are far less exposed and left untouched here.

## Reusable add-transaction button

Extracted the header quick-entry split button into a self-loading `AddTransactionButton` server component (fetches its own accounts/currency/templates). The header slot and the dashboard "Recent transactions" header both render it, so the dashboard button now matches the header's full entry-variant behavior instead of being a plain link.

### DIFF
--- a/features/reconcile/Reconcile.tsx
+++ b/features/reconcile/Reconcile.tsx
@@ -3,6 +3,7 @@ import ExportButton from '@/components/ExportButton';
 import Help from '@/components/Help';
 import PageContainer from '@/components/PageContainer';
 import TransactionRow from '@/features/transactions/row/TransactionRow';
+import { registerFormat } from '@/features/transactions/row/registerRows';
 import { getBaseCurrency } from '@/lib/settings';
 import runLedger from '@/utils/runLedger';
 
@@ -22,7 +23,7 @@ const Reconcile = async () => {
       '--sort',
       'date',
       '--format',
-      'NNN%D|%P|%A|%t|%(note)\n',
+      registerFormat(['%D', '%P', '%A', '%t']),
     ],
     { sortByDate: false }
   );
--- a/features/transactions/row/registerRows.test.ts
+++ b/features/transactions/row/registerRows.test.ts
@@ -1,29 +1,61 @@
 import { describe, expect, it } from 'vitest';
-import { parseAccountRegister } from './registerRows';
+import { FIELD_SEP, RECORD_SEP, parseAccountRegister } from './registerRows';
+
+// Build a register line the way ledger would, using the real separators.
+const row = (...fields: string[]) => `${RECORD_SEP}${fields.join(FIELD_SEP)}`;
 
 describe('parseAccountRegister', () => {
   it('parses date/payee/amount/total and extracts the uid from the note', () => {
-    const stdout =
-      'NNN2026/01/01|Coffee|$ -5.00|$ -5.00| :uid: 01HZY0Z9QK8G7F6E5D4C3B2A1Z';
-    const [row] = parseAccountRegister(stdout);
-    expect(row.date).toBe('2026/01/01');
-    expect(row.payee).toBe('Coffee');
-    expect(row.amount).toBe('$ -5.00');
-    expect(row.runningTotal).toBe('$ -5.00');
-    expect(row.uid).toBe('01HZY0Z9QK8G7F6E5D4C3B2A1Z');
+    const stdout = row(
+      '2026/01/01',
+      'Coffee',
+      '$ -5.00',
+      '$ -5.00',
+      ' :uid: 01HZY0Z9QK8G7F6E5D4C3B2A1Z'
+    );
+    const [parsed] = parseAccountRegister(stdout);
+    expect(parsed.date).toBe('2026/01/01');
+    expect(parsed.payee).toBe('Coffee');
+    expect(parsed.amount).toBe('$ -5.00');
+    expect(parsed.runningTotal).toBe('$ -5.00');
+    expect(parsed.uid).toBe('01HZY0Z9QK8G7F6E5D4C3B2A1Z');
   });
 
   it('has no uid when the note lacks one (actions disabled downstream)', () => {
-    const stdout = 'NNN2026/01/02|Book|$ -20.00|$ -25.00|';
-    const [row] = parseAccountRegister(stdout);
-    expect(row.uid).toBeUndefined();
+    const stdout = row('2026/01/02', 'Book', '$ -20.00', '$ -25.00', '');
+    const [parsed] = parseAccountRegister(stdout);
+    expect(parsed.uid).toBeUndefined();
   });
 
   it('keeps a multi-commodity running total intact and preserves a note pipe', () => {
+    const stdout = row(
+      '2026/03/01',
+      'Split',
+      'KIRT 100',
+      '$ -5.00\nKIRT 100',
+      'a|b :uid: 01HZY0Z9QK8G7F6E5D4C3B2A1Z'
+    );
+    const [parsed] = parseAccountRegister(stdout);
+    expect(parsed.runningTotal).toBe('$ -5.00\nKIRT 100');
+    expect(parsed.uid).toBe('01HZY0Z9QK8G7F6E5D4C3B2A1Z');
+  });
+
+  it('parses a row whose uid contains the old "NNN" marker without corruption', () => {
+    // Regression: a ULID may contain "NNN" (its alphabet includes N). Splitting
+    // on "NNN" shifted the next chunk into an "Invalid Date" row and dropped
+    // the uid. Control-char separators keep the row and uid intact.
     const stdout =
-      'NNN2026/03/01|Split|KIRT 100|$ -5.00\nKIRT 100|a|b :uid: 01HZY0Z9QK8G7F6E5D4C3B2A1Z';
-    const [row] = parseAccountRegister(stdout);
-    expect(row.runningTotal).toBe('$ -5.00\nKIRT 100');
-    expect(row.uid).toBe('01HZY0Z9QK8G7F6E5D4C3B2A1Z');
+      row(
+        '2026/07/12',
+        'ghaza khoshk',
+        'KIRT -9000',
+        'KIRT -9000',
+        ' :uid: 01HZXNNN8E5T0DJC5G5KJDQRYK'
+      ) + row('2026/07/10', 'earlier', 'KIRT -100', 'KIRT -9100', '');
+    const rows = parseAccountRegister(stdout);
+    expect(rows).toHaveLength(2);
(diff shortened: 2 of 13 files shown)


features/transactions/row/registerRows.ts
```diff
@@ -1,33 +1,48 @@
...
+// like "NNN" appears inside a uid such as `01HZXNNN…`, splitting mid-record and
+// rendering the next chunk as an "Invalid Date" row. Control chars can never
+// occur in a date, amount, ULID, or a normally-typed note.
+export const RECORD_SEP = '\x1e';
+export const FIELD_SEP = '\x1f';
+
+// Row separator `RECORD_SEP`; fields joined by `FIELD_SEP`. `%T` (total) may
+// span multiple commodities (embedded newlines, no field sep); `%(note)` is
+// taken as the rejoined remainder after the four fixed leading fields.
+export const REGISTER_FORMAT = `${RECORD_SEP}%D${FIELD_SEP}%P${FIELD_SEP}%t${FIELD_SEP}%T${FIELD_SEP}%(note)`;
```
naeemba: `REGISTER_FORMAT` hand-builds the exact shape the new `registerFormat()` helper below produces, but inconsistently — the helper appends a trailing `\n`, this constant does not. Define it through the helper so the two can never diverge on separators or the newline:

```suggestion
export const REGISTER_FORMAT = registerFormat(['%D', '%P', '%t', '%T']);
```
(requires moving `registerFormat` above this line)

naeemba: Fixed in 97196d3 — REGISTER_FORMAT is now defined via registerFormat(['%D', '%P', '%t', '%T']), with the helper moved above it.

features/reconcile/Reconcile.tsx
```diff
@@ -22,7 +23,7 @@ const Reconcile = async () => {
       '--sort',
       'date',
       '--format',
-      'NNN%D|%P|%A|%t|%(note)\n',
+      registerFormat(['%D', '%P', '%A', '%t']),
```
naeemba, line 26: This call site was migrated, but its sibling `app/api/reconcile/export/route.ts` was not: it still passes `--format 'NNN%D|%P|%A|%t\n'` into the same `parseReconcileRows`, whose splitter now looks for `RECORD_SEP`/`FIELD_SEP`. The NNN output contains neither, so `splitRegisterRows` produces one 1-column chunk, the `cols.length >= 4` guard drops it, and the reconcile CSV export now returns an empty file for everyone.

Fix: use `registerFormat(['%D', '%P', '%A', '%t'])` there too (the appended `%(note)` column is harmless — `reconcileRowsToCsv` only reads date/payee/account/amount).

While in the reconcile surface: the doc comment on `parseReconcileRows` in `Reconcile.utils.ts` still describes the old `NNN%D|%P|%A|%t|%(note)\n` shape — update it to the new separators so it doesn't mislead.

features/transactions/row/registerRows.ts
```diff
@@ -1,33 +1,48 @@
...
+// rendering the next chunk as an "Invalid Date" row. Control chars can never
+// occur in a date, amount, ULID, or a normally-typed note.
+export const RECORD_SEP = '\x1e';
+export const FIELD_SEP = '\x1f';
+
+// Compose a register format from a field list, e.g. `['%D', '%P', '%A', '%t']`
+// followed by `%(note)`. Keeps every splitRegisterRows caller on the same
+// separators so the split below stays correct.
+export const registerFormat = (leadingFields: string[]): string =>
+  `${RECORD_SEP}${[...leadingFields, '%(note)'].join(FIELD_SEP)}\n`;
```
naeemba: The rationale above (a printable marker collides with free-text content) applies equally to the surfaces this PR did not migrate. In particular, every format that carries `%P` (payee — free text) still uses `NNN`/`|`:

- `features/payees/Payees.tsx` + `lib/payees/parse.ts` (`'NNN%P|%t\n'`)
- `app/api/payees/export/route.ts` (`'NNN%P|%t\n'`)
- `app/api/reconcile/export/route.ts` (`'NNN%D|%P|%A|%t\n'` — also outright broken, see the other comment)

A payee containing `|` (or the marker string) corrupts those rows exactly the way the ULID did here. The date/total-only formats (`'NNN%D|%T\n'` etc.) can't realistically collide, so they can stay. Fine as a follow-up, but the free-text ones should move to `RECORD_SEP`/`FIELD_SEP` (their parsers expect 2–4 columns, so they'd reuse the constants rather than `splitRegisterRows`).

features/transactions/row/registerRows.ts
```diff
@@ -1,33 +1,48 @@
...
-export const REGISTER_FORMAT = 'NNN%D|%P|%t|%T|%(note)';
+// Record/field separators for `ledger reg --format`. ASCII RS (0x1e) and US
+// (0x1f) are used instead of a printable marker because a printable one
+// collides with content: ULID uids (carried in `%(note)`) draw from an
+// alphabet that includes the letter, and a payee/note is free text — a marker
+// like "NNN" appears inside a uid such as `01HZXNNN…`, splitting mid-record and
+// rendering the next chunk as an "Invalid Date" row. Control chars can never
+// occur in a date, amount, ULID, or a normally-typed note.
+export const RECORD_SEP = '\x1e';
+export const FIELD_SEP = '\x1f';
```
naeemba: Naming: `SEP` abbreviates "separator", and the convention here is to spell identifiers out in full (no `btn`/`msg`/`config`-style contractions). Rename to `RECORD_SEPARATOR` / `FIELD_SEPARATOR` — the exports are new in this PR, so the rename is still cheap (this file, the two test files, and `Reconcile.utils.test.ts`).

features/transactions/row/registerRows.test.ts
```diff
@@ -1,29 +1,61 @@
 import { describe, expect, it } from 'vitest';
-import { parseAccountRegister } from './registerRows';
+import { FIELD_SEP, RECORD_SEP, parseAccountRegister } from './registerRows';
+
+// Build a register line the way ledger would, using the real separators.
```
naeemba, line 8: This builder omits the trailing `\n` that `registerFormat` now appends to every record (the reconcile test's `row` helper does append it), so these fixtures drift from real `REGISTER_FORMAT` output — the note column never carries the trailing newline it always has in production. It works today only because `UID_TAG_REGEX` is unanchored. Append `\n` here to keep the fixture faithful:

```suggestion
const row = (...fields: string[]) => `${RECORD_SEP}${fields.join(FIELD_SEP)}\n`;
```

features/transactions/row/registerRows.ts
naeemba: Verified on 97196d3: `REGISTER_FORMAT` is now `registerFormat(['%D', '%P', '%t', '%T'])` with the helper defined above it, so the constant and the helper can no longer diverge on separators or the trailing newline. Confirmed against ledger 3.4.1 output that the control-char format round-trips through `splitRegisterRows` with the `NNN`-containing ULID intact. Resolving.

features/reconcile/Reconcile.tsx
naeemba: Fixed in 07efc15 — the reconcile export route now passes registerFormat(['%D', '%P', '%A', '%t']), and the parseReconcileRows doc comment in Reconcile.utils.ts is updated to the new shape.

features/transactions/row/registerRows.ts
naeemba: Done in 07efc15 rather than as a follow-up — all three free-text surfaces are migrated: lib/payees/parse.ts now splits on RECORD_SEPARATOR/FIELD_SEPARATOR and exports a shared PAYEE_REGISTER_FORMAT used by both features/payees/Payees.tsx and app/api/payees/export/route.ts; app/api/reconcile/export/route.ts now uses registerFormat (see the other thread). The date/total-only formats keep NNN as suggested.

features/transactions/row/registerRows.ts
naeemba: Renamed in 07efc15 — RECORD_SEPARATOR / FIELD_SEPARATOR across registerRows.ts, both test files, and the new lib/payees/parse.ts usage.

features/transactions/row/registerRows.test.ts
naeemba: Fixed in 07efc15 — the row helper now appends the trailing \n, matching registerFormat output.

features/reconcile/Reconcile.tsx
naeemba: Verified on 07efc15: the export route now builds its format via registerFormat(['%D', '%P', '%A', '%t']), so it emits RECORD_SEPARATOR/FIELD_SEPARATOR and parseReconcileRows splits it correctly again. Reconcile.utils tests pass against the new shape.

features/transactions/row/registerRows.ts
naeemba: Verified on 07efc15: all three %P surfaces now use the control-char separators — lib/payees/parse.ts exports PAYEE_REGISTER_FORMAT (used by features/payees/Payees.tsx and app/api/payees/export/route.ts) and app/api/reconcile/export/route.ts uses registerFormat. The only remaining NNN formats carry %D/%A/%t/%T, which cannot realistically collide, as scoped in the original comment.

features/transactions/row/registerRows.ts
naeemba: Verified on 07efc15: RECORD_SEPARATOR / FIELD_SEPARATOR are used consistently across registerRows.ts, both test files, and lib/payees/parse.ts; no SEP-abbreviated identifier remains.

features/transactions/row/registerRows.test.ts
naeemba: Verified on 07efc15: the row helper appends the trailing newline, matching registerFormat output; both register test suites pass.

